PR #2, opened 2019-12-13
rename enterprise_catalog to catalog

adamstankiewicz:


nedbat:
Are we sure "catalog" with no modifier whatsoever will serve us best in the long run? We'll never have another kind of catalog? What;s in this catalog?

PR #10, opened 2019-12-17
add catalog app, adds CatalogQuery model and migrations

adamstankiewicz:
## Description

TODO: Include a detailed description of the changes in the body of the PR

## Manual Test Plan

TODO: Include instructions for any required manual tests, and any manual testing that has
already been performed.

## Testing Checklist
- TODO: Include unit, integration, acceptance tests as appropriate
- TODO: Include accessibility (a11y) tests
- TODO: Include tests that capture the external-query scaling properties
- [ ] Check that Database migrations are backwards-compatible
- [ ] Manually test right-to-left languages and i18n
  of the changes.

## Non-testing Checklist
- TODO: Tag DevOps on your PR (user `edx/devops`), either for review
  (more specific to devops = better) or just a general FYI.
- TODO: Consider any documentation your change might need, and which
  users will be affected by this change.
- TODO: Double-check that your commit messages will make a meaningful release note.

## Post-review
- TODO: Squash commits into discrete sets of changes (see the note about release notes above)

## Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [ ] Code review #1 (TODO: tag a specific user)
- [ ] Code review #2 (TODO: tag a specific user)
- [ ] Docs review (required for UI strings/error messages)
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

### Areas to Consider
- [ ] i18n 
    - Are all user-facing strings tagged?
- [ ] Right-to-left
    - Will the feature work for right-to-left languages?
- [ ] Analytics
    - Are any new events being emitted?
    - Have any events been changed?
    - Are there any new user actions that should be emitted as events?
- [ ] Performance
    - What dimensions does this change scale over? Users? Courses? Course size?
    - How does the feature scale over those dimensions? Sub-linear? Linear? Quadratic? Exponential?
    - How does the scaling affect the number of external calls (database queries,
      api requests, etc) that this code makes?
- [ ] Database migrations
    - Are they backwards compatible?
    - When they run on production, how long will they take? Will they lock the table?


enterprise_catalog/apps/catalog/models.py
```diff
@@ -0,0 +1,44 @@
...
+class CatalogQuery(models.Model):
+    """
+    Stores a re-usable catalog query.
+
+    .. no_pii:
+    """
+
+    title = models.CharField(
+        default='All Content',
+        max_length=255,
```
bbaker6225, line 19: nit: I personally prefer setting a tighter constraint on max length if we can, but that's not really important

enterprise_catalog/apps/catalog/models.py
```diff
@@ -0,0 +1,44 @@
+import collections
+
+from jsonfield.fields import JSONField
+from model_utils.models import TimeStampedModel
+
+from django.db import models
+from django.utils.translation import gettext as _
+
+
+class CatalogQuery(models.Model):
```
bbaker6225, line 10: Not sure if you're planning on waiting until you have more of the models created, but I think factories for the models would be nice to have

bbaker6225 (approved):
Small comments but nice overall ⛵️

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: Good question. At the moment, I'm not entirely sure how to set those up so I'll hold off for now. Maybe once all the models are defined, we can do a pass through to create factories?

PR #11, opened 2019-12-18
Add EnterpriseCatalog model

adamstankiewicz:
## Description

Adds `EnterpriseCatalog` model.


enterprise_catalog/apps/catalog/models.py
```diff
@@ -41,4 +43,48 @@ def __str__(self):
...
+
+    .. no_pii:
+    """
+
+    uuid = models.UUIDField(
+        primary_key=True,
+        default=uuid4,
+        editable=False
+    )
+    enterprise_uuid = models.CharField(
```
bbaker6225: Are we linking with `uuid` or just `id`? If we're doing `uuid` I'd think we'd want to use a `UUIDField`, and if `id`then probably an `IntegerField`

enterprise_catalog/apps/catalog/models.py
```diff
@@ -41,4 +43,48 @@ def __str__(self):
...
+        max_length=255,
+        blank=False,
+    )
+    content_catalog_query = models.ForeignKey(
+        CatalogQuery,
+        blank=False,
+        null=False,
```
bbaker6225, line 65: Could a `null` `CatalogQuery` potentially mean something like there is no restriction on the catalog? Not sure how that's currently represented as a query

enterprise_catalog/apps/catalog/models.py
```diff
@@ -41,4 +43,48 @@ def __str__(self):
...
+        related_name='content_catalog_queries',
```
bbaker6225: I think we want the `related_name` to be `enterprise_catalog`.  I'd also suggest that we make the field name just `catalog_query` now that we changed the model name

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: A catalog with a content filter of an empty object (ie., `{}`) represents "All Content" or no restrictions on the catalog. If an enterprise has no restriction on that, there should be a shared catalog that represents "All Content"  for that case.

bbaker6225 (approved):

PR #12, opened 2019-12-18
Add CatalogContentKey model and migrations

adamstankiewicz:


enterprise_catalog/apps/catalog/models.py
```diff
@@ -63,7 +63,7 @@ class EnterpriseCatalog(TimeStampedModel):
         CatalogQuery,
         blank=False,
         null=False,
-        related_name='enterprise_catalog',
+        related_name='enterprise_catalogs',
```
bbaker6225, line 66: Shouldn't it still be singular since it's a foreign key?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -85,3 +85,46 @@ def __str__(self):
...
+    Associates a stored catalog query with an enterprise customer.
+
+    .. no_pii:
+    """
+
+    catalog_query = models.ForeignKey(
+        CatalogQuery,
+        blank=False,
+        null=False,
+        related_name='catalog_content_keys',
```
bbaker6225, line 101: Same thing with plural/foreign key thing

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: From reading the docs about `related_name`, it's used when referring back to `EnterpriseCatalog` from a `CatalogQuery` object. Based on the docs, it'd be like:
```
q = CatalogQuery
q.enterprise_catalogs.count()
```
in which case, the plural makes more sense. We use plural within `edx-enterprise` as well.

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: See response above.

enterprise_catalog/apps/catalog/models.py
bbaker6225: Whoops yep, got the relationship backwards for a second

bbaker6225 (approved):

PR #16, opened 2019-12-19
Add ContentMetadata model

adamstankiewicz:
## Description

* Adds `ContentMetadata` model.
* Ensures its added to Django admin screen
* Migrations

## Ticket Link

Link to the associated ticket
[ENT-2447](https://openedx.atlassian.net/browse/ENT-2447)

## Post-review


enterprise_catalog/apps/catalog/models.py
```diff
@@ -101,12 +169,14 @@ class CatalogContentKey(TimeStampedModel):
         related_name='catalog_content_keys',
         on_delete=models.deletion.CASCADE
     )
-    content_key = models.CharField(
-        max_length=255,
+    content_key = models.ForeignKey(
```
adamstankiewicz, line 171: The link between `CatalogContentKey` and `ContentMetadata` is on `content_key`, so I'm introducing a foreign key relationship here on the `content_key` column.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -20,7 +20,8 @@ class CatalogQuery(models.Model):
         default='All Content',
         max_length=255,
         blank=False,
-        null=False
+        null=False,
+        unique=True
```
bbaker6225: `unique=True` with a non-unique default value feels a little odd to me, but maybe it's normal?

bbaker6225 (approved):

PR #26, opened 2019-12-27
Miscellaneous model field updates

adamstankiewicz:
## Description

* Add `unique` constraint on `title` on `CatalogQuery` and removes default
* Add `unique` constraint on `title` on `EnterpriseCatalog` and removes default
* Add `unique_together` constraint on `EnterpriseCatalog` for `enterprise_uuid` and `catalog_query`
* Ensure `enterprise_uuid` is a required field on `EnterpriseCatalog`
* Add index to `parent_content_key` on `ContentMetadata`
* Add history table to `ContentMetadata`

## Ticket Link

Link to the associated ticket
[ENT-2447](https://openedx.atlassian.net/browse/ENT-2447)

## Post-review


enterprise_catalog/apps/catalog/models.py
```diff
@@ -63,12 +63,15 @@ class EnterpriseCatalog(TimeStampedModel):
         editable=False
     )
     title = models.CharField(
-        default='All Content',
         max_length=255,
         blank=False,
-        null=False
+        null=False,
+        unique=True
```
brittneyexline: we might not want title to be unique on this model... despite what i may have said earlier. i think it's more relevant on the CatalogQuery model than here.

brittneyexline (approved):
just one small thing to address before merging!

PR #30, opened 2020-01-06
API endpoint for enterprise catalog CRUD operations

adamstankiewicz:
## Description

This PR adds an endpoint for CRUD operations on the `EnterpriseCatalog` model.

Some other minor changes are included:

- `course_run` is now changed to `courserun` to reflect what comes back in the data
-  adds some `REST_FRAMEWORK` settings to generate docs at `/api-docs` and set up pagination
-  Drops `title` from `CatalogQuery`

## Ticket Link

[ENT-2448](https://openedx.atlassian.net/browse/ENT-2448)

## Post-review


enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -0,0 +1,100 @@
...
+class CatalogContentKeySerializer(QuerySerializerMixin, serializers.ModelSerializer):
+    """
+    Serializer for the `CatalogContentKey` model
+    """
+
+    class Meta:
+        model = CatalogContentKey
+        fields = ['id']
+
+class EnterpriseCatalogSerializer(QuerySerializerMixin, serializers.ModelSerializer):
```
brittneyexline: i assume that this serializer should own all the CRUD operations for Enterprise Catalog, so it will need to be updated to support that.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -0,0 +1,100 @@
...
+
+    class Meta:
+        model = EnterpriseCatalog
+        fields = ['uuid', 'title', 'enterprise_customer']
+
+    def get_enterprise_customer(self, obj):
+        return obj.enterprise_uuid
+
+
+class EnterpriseCatalogDetailSerializer(EnterpriseCatalogSerializer):
```
brittneyexline: for this, we may want to rename to something like `EnterpriseCatalogContentMetadataSerializer` and not inherit from `EnterpriseCatalogSerializer`, since this will be read only. This seems to be geared towards supporting the `get_enterprise_catalog` functionality in the ticket, and I suspect it won't own the logic for `is_content_in_catalog` (that will probably have its own serializer).

My idea for what this serializer would do is to basically join from `EnterpriseCatalog` to `CatalogQuery` to `CatalogContentKey` to `ContentMetadata` via foreign key relationships. This should be agnostic of the discovery service at this point.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -1 +1,25 @@
...
+
+
+class EnterpriseCatalogViewSet(viewsets.ViewSet):
+    """ TODO """
+    def list(self, request):
+        queryset = EnterpriseCatalog.objects.all()
+        serializer = EnterpriseCatalogSerializer(queryset, many=True)
+        return Response(serializer.data)
+
+    def retrieve(self, request, pk=None):
```
brittneyexline: let's keep retrieve here as returning the model's data instead of doing the content metadata - i would prefer a separate endpoint for that; it could be a custom method on this viewset though.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -108,6 +108,41 @@ class Meta:
         unique_together = (("enterprise_uuid", "catalog_query"),)
         app_label = 'catalog'
 
+    def get_paginated_content(self, query_parameters=None):
```
brittneyexline: not sure if this should live here... i realize this was started based off edx-enterprise code which is why it is here. it might make sense in the end or not, just food for thought as we refine this implementation.

brittneyexline (commented):
i think we should consider making separate PRs for the EnterpriseCatalog CRUD endpoints vs. returning content_metadata and is_content_in_catalog. That will likely unblock the other tickets faster and make this easier to review.

enterprise_catalog/apps/api/v1/serializers.py
bbaker6225: 👍 removing logic around this stuff from this PR to just get basic CRUD for `EnterpriseCatalog`

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -0,0 +1,71 @@
...
+        model = CatalogContentKey
+        fields = ['id']
+
+
+class EnterpriseCatalogSerializer(QuerySerializerMixin, serializers.ModelSerializer):
+    """
+    Serializer for the `EnterpriseCatalog` model
+    """
+    enterprise_customer = serializers.UUIDField(source='enterprise_uuid')
+    # content_filter = This is in the mgmt command WIP, but not sure if it should be part of this serializer
```
bbaker6225: @brittneyexline Do you mind giving a little more detail on the `content_filter` that we expect to write to this endpoint as part of the management command? 
I know that `content_filter` is related to the catalog by the foreign key on `EnterpriseCatalog` to `CatalogQuery`, but I'm not sure how we plan to populate `CatalogQuery.title` when creating the related query

brittneyexline: good question. my naive thought would be to just pass through the title sent along for the enterprise catalog, but then we have two titles that could start to diverge once another query with the same params is added somewhere else. based on this, i'm tempted to say we should drop title from `CatalogQuery`?

brittneyexline: to be clear (though i think you probably know), my expectation for this serializer is to create/fetch if it exists the `CatalogQuery` object with the `content_filter` value, and for that object to be keyed to from `EnterpriseCatalog`.

bbaker6225: Yeah I'm fine if we drop `CatalogQuery.title` if we're not using it for anything (which sounds like we're not). I wasn't sure if the title provided useful info to anyone

bbaker6225:
For permissions, are we planning on pulling the `EnterpriseCustomer` and `EnterpriseCustomerUser` into this IDA as well? `edx-enterprise` currently uses those for its permissions and I don't think I'm understanding how we plan to do it without them

brittneyexline:
another good question - i would say we should file a follow up story for permissions. we should be able to use rbac to do permissions without pulling in any other enterprise models. i can file this ticket for grooming tomorrow. for now, let's keep it simple and make the permission for staff users only.

enterprise_catalog/apps/api/v1/serializers.py
brittneyexline: it was probably just an oversight that we had title in both places.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -1 +1,22 @@
...
+    EnterpriseCatalogSerializer,
+)
+from enterprise_catalog.apps.catalog.models import EnterpriseCatalog
+
+
+class EnterpriseCatalogViewSet(viewsets.ModelViewSet):
+    """ View for CRUD operations on Enterprise Catalogs """
+    serializer_class = EnterpriseCatalogSerializer
+    queryset = EnterpriseCatalog.objects.all()
+    permission_classes = [permissions.IsAdminUser]
```
brittneyexline: i think this should be `permission_classes = [permissions.IsAuthenticated, permissions.IsAdminUser]` (at least i've seen this elsewhere). we should also add this line so that jwt authentication works: `authentication_classes = (JwtAuthentication, BearerAuthentication, SessionAuthentication,)`

bbaker6225: Good catch on auth classes. For permissions, `IsAdminUser` is a subset of `IsAuthenticated`, but I added it just to be explicit

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,154 @@
...
+    def test_detail(self):
+        """
+        Verify the endpoint returns the details for a single enterprise catalog
+        """
+        url = reverse('api:v1:enterprise-catalog-detail', kwargs={'uuid': self.enterprise_catalog.uuid})
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(response.data, self.serialize_enterprise_catalog(self.enterprise_catalog))
+
+    def test_detail_unauthorized(self):
```
bbaker6225, line 93: Having an unauthorized test for each method feels excessive to me, but I went ahead and added them because I wasn't sure how granular we wanted to be in our restrictions (and the tests on those restrictions). Happy to remove all but one unauthorized test if people feel that we don't need this many

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def create(self, validated_data):
+        content_filter = validated_data.pop('content_filter')
+        catalog_query, _ = CatalogQuery.objects.get_or_create(
+            content_filter=content_filter,
+            content_filter_hash=get_content_filter_hash(content_filter),
+        )
+        return EnterpriseCatalog.objects.create(**validated_data, catalog_query=catalog_query)
+
+    def update(self, instance, validated_data):
```
christopappas, line 38: we don't call super() in this create or update, and while we don't have to if we don't need it, typically i see it when we're overriding some django function

what is to be gained by running super() here? not sure off hand, but it is a thing that caught my eye so i wanted to call it out

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -0,0 +1,163 @@
...
+        """
+        Verify the endpoint returns a list of all enterprise catalogs
+        """
+        url = reverse('api:v1:enterprise-catalog-list')
+        second_enterprise_catalog = EnterpriseCatalogFactory(catalog_query__content_filter='{"content_type":"course"}')
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+        self.assertEqual(
+            response.data['results'],
+            self.serialize_enterprise_catalog([self.enterprise_catalog, second_enterprise_catalog], many=True)
```
christopappas: Maybe personal preference, but I would prefer asserts about the final state of the response. Maybe the length of the list, or the values of the items we would expect in that list. the serialize_enterprise_catalog method seems to be running code similar to what the non test code is doing, which feels more like duplicating the logic, and less about testing the final state of something that we would expect

enterprise_catalog/apps/api/v1/tests/test_views.py
christopappas: Takes 0 time to run I'm sure. I'd say we keep it for now

enterprise_catalog/apps/api/v1/urls.py
```diff
@@ -1,4 +1,20 @@
-""" API v1 URLs. """
+# -*- coding: utf-8 -*-
```
christopappas, line 1: nitpick and personal preference, but there are a lot of newlines in this file ><'

christopappas (approved):
👍 from me. a few nitpicky things, but nothing i think should stop rolling this out

enterprise_catalog/apps/api/v1/serializers.py
bbaker6225: Mm yeah. I'll add the super to `update` since that saves some boilerplate at the end. I'm pretty sure we don't want the `super` for create though (at least based on the examples I've seen)
https://www.django-rest-framework.org/api-guide/serializers/#saving-instances

enterprise_catalog/apps/api/v1/tests/test_views.py
bbaker6225: Yeah you're right. I took this idea from course-discovery... but I don't think it's very helpful in retrospect

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -0,0 +1,57 @@
...
+            'uuid',
+            'title',
+            'enterprise_customer',
+            'enabled_course_modes',
+            'publish_audit_enrollment_urls',
+            'content_filter',
+        ]
+
+    def create(self, validated_data):
+        content_filter = validated_data.pop('content_filter')
```
brittneyexline, line 31: i had to do this locally in order to properly store the json string coming through the api: `content_filter = json.loads(validated_data.pop('content_filter'))`

bbaker6225: Not totally sure, but I believe DRF already handles the back and forth conversion for JSON (https://www.django-rest-framework.org/api-guide/fields/#jsonfield). Also, if you go to http://localhost:18160/api/v1/enterprise-catalog/ and use the rendered API POST/PUT form, `json.loads` will give you a type error but it works normally without it

brittneyexline: hmm. when calling from postman as well as from my script, i had to encode the json fields in order for the request parsing not to fail. when i viewed the objects created, they were stored as json strings, with quotes around the actual json. let me see if there is something wonky on my side.

bbaker6225: I guess I should say that the `JSONField` does validate and make sure that the incoming data is valid `json`, in case that's the issue

brittneyexline (approved):

PR #33, opened 2020-01-10
Added devstack config yml file to app repo

syedimranhassan:
## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review


coryleeio (changes requested):
Since this is devstack specific you might consider moving it into the configuration/docker/build folder for this application and just add a COPY enterprise_catalog.yml /edx/etc/enterprise_catalog.yml to the dockerfile.  

Then you wouldn't have to have the strange config path as the config would be in /edx/etc like normal

PR #38, opened 2020-01-15
Refactoring to use a ManyToMany field for relationship between Conten…

christopappas:
…tMetadata and CatalogQuery

## Description

Removing CatalogContentKey table in favor of using a ManyToMany field

A snippet from the shell to see what working with this would look like:

```
>>> cq1 = CatalogQuery.objects.first()
>>> cq2 = CatalogQuery.objects.last()
>>> cm1 = ContentMetadata.objects.first()
>>> cm2 = ContentMetadata.objects.last() 
>>> 
>>> cm1.catalog_queries.add(cq1)
>>> cq1.contentmetadata_set.add(cm2)
>>> cq2.contentmetadata_set.add(cm2)
>>> 
>>> cq1.contentmetadata_set.all()
<QuerySet [<ContentMetadata: <ContentMetadata for 'testing'>>, <ContentMetadata: <ContentMetadata for 'testing3'>>]>
>>> cq2.contentmetadata_set.all()
<QuerySet [<ContentMetadata: <ContentMetadata for 'testing3'>>]>
>>> 
>>> cm1.catalog_queries.all()
<QuerySet [<CatalogQuery: <CatalogQuery with content filter hash '8393dd4a16f0e1c9723f9d9cfa2c39f1'>>]>
>>> cm2.catalog_queries.all()
<QuerySet [<CatalogQuery: <CatalogQuery with content filter hash '8393dd4a16f0e1c9723f9d9cfa2c39f1'>>, <CatalogQuery: <CatalogQuery with content filter hash '9a293a221790e7dd818b6c83adb7c98f'>>]>
>>> 
```

## Ticket Link

Link to the associated ticket

## Post-review


enterprise_catalog/apps/catalog/models.py
```diff
@@ -159,6 +159,7 @@ class ContentMetadata(TimeStampedModel):
             "endpoint results, specified as a JSON object."
         )
     )
+    catalog_queries = models.ManyToManyField(CatalogQuery)
```
bbaker6225, line 162: Do we want a related name here? Something like `content_metadata_chunks` (although I think we can do better than that)

enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -9,7 +9,6 @@
     json_serialized_course_modes,
 )
 from enterprise_catalog.apps.catalog.models import (
-    CatalogContentKey,
     CatalogQuery,
     ContentMetadata,
```
bbaker6225, line 13: Nit: This factory could be updated now, but I'm fine leaving it until we actually start using it

bbaker6225 (approved):

enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -59,17 +58,6 @@ class Meta:
     json_metadata = "{}"  # Default metadata to empty object
```
adamstankiewicz, line 59: Should we add `catalog_queries` to this factory?

enterprise_catalog/apps/catalog/tests/factories.py
christopappas: So @adamstankiewicz @bbaker6225 , perhaps this is personal preference, but I am of the opinion that factories should not create any related objects when a many-to-many relationship exists, because in theory there could be any number of relationships that could be created by default. Instead, a factory would be used to make any given object, and then the writer of test code would specify the connection between any number of catalog queries and content metadata objects

in other words: I felt like there was no good default to specify with a many-to-many relationship

enterprise_catalog/apps/catalog/models.py
christopappas: by default it is `contentmetadata_set`. i'm comfortable with this because it's a pretty common naming pattern, but if others on the team feel there is a more useful related name, i'm happy to update

enterprise_catalog/apps/catalog/tests/factories.py
bbaker6225: Fine with me. If we find that there is a more convenient way to do this factory down the road we can do it then

PR #40, opened 2020-01-15
Disable changing models through admin

bbaker6225:
https://openedx.atlassian.net/browse/ENT-2544

### DIFF
--- a/enterprise_catalog/apps/catalog/constants.py
+++ b/enterprise_catalog/apps/catalog/constants.py
@@ -1,5 +1,10 @@
 import json
 
+import waffle
+
+
+# Waffle Switches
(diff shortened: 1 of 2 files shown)


adamstankiewicz (approved):
Looks good to me 👍

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -19,3 +24,9 @@ def json_serialized_course_modes():
     :return: serialized course modes.
     """
     return json.dumps(COURSE_MODE_SORT_ORDER)
+
+def admin_model_changes_allowed():
```
bbaker6225, line 28: Added this to make local dev a bit easier, and allow "emergency edits".

TODO: Add this switch to stage and prod once it goes out

PR #42, opened 2020-01-17
ENT-2449 | Adding code to call discovery to get metadata, and to upda…

christopappas:
…te database with that metadata

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review


enterprise_catalog/settings/devstack.py
```diff
@@ -14,8 +14,8 @@
 )
 
 # OAuth2 variables specific to backend service API calls.
-BACKEND_SERVICE_EDX_OAUTH2_KEY = os.environ.get('BACKEND_SERVICE_EDX_OAUTH2_KEY', 'enterprise-catalog-backend-service-key')
-BACKEND_SERVICE_EDX_OAUTH2_SECRET = os.environ.get('BACKEND_SERVICE_EDX_OAUTH2_SECRET', 'enterprise-catalog-backend-service-secret')
+BACKEND_SERVICE_EDX_OAUTH2_KEY = os.environ.get('BACKEND_SERVICE_EDX_OAUTH2_KEY', 'enterprise_catalog-key')
```
christopappas, line 17: I need to verify the value that will need to go here for production still

enterprise_catalog/apps/api_client/tests/test_discovery.py
```diff
@@ -0,0 +1,107 @@
...
+    @mock.patch('enterprise_catalog.apps.api_client.discovery.OAuthAPIClient')
+    def test_get_metadata_by_query_with_traverse(self, mock_oauth_client, mock_traverse):
+        """
+        get_metadata_by_query should call discovery endpoint, and then call
+        traverse_pagination if traverse_pagination is true.
+        """
+        mock_oauth_client.return_value.post.return_value.json.return_value = {
+            'this is': 'just a test'
+        }
+        mock_traverse.return_value = [{'Title': 'My Title', 'Etc': 'More stuff'}]
```
christopappas, line 22: I can make this a little more intuitive as far as what it returns (something that adds to `results`)

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -0,0 +1,91 @@
...
+            response = self.client.post(
+                self.SEARCH_ALL_ENDPOINT,
+                data=content_filter_query,
+                **query_params
+            ).json()
+            results += response.get('results', [])
+
+        return results
+
+    def get_metadata_by_query(self, content_filter_query, query_params=None, traverse_pagination=False):
```
bbaker6225, line 60: Could `query_params` default to `{}`?

enterprise_catalog/apps/api_client/tests/test_discovery.py
```diff
@@ -0,0 +1,107 @@
...
+            current_response_dict,
+            content_filter,
+            query_params
+        )
+        expected_results = [
+            {'item1': 'is from the first page'},
+            {'item2': 'is from the second page'},
+        ]
+
+        assert actual_results == expected_results
```
bbaker6225: I usually prefer an `assertEqual` sort of assert since it will give better debugging output when it fails

enterprise_catalog/apps/catalog/tests/test_models.py
```diff
@@ -0,0 +1,46 @@
...
+                    'title': 'course 3',
+                },
+            ]
+        }
+        catalog = factories.EnterpriseCatalogFactory()
+
+        assert ContentMetadata.objects.count() == 0
+        update_contentmetadata_from_discovery(catalog.uuid)
+        mock_client.assert_called_once()
+        assert ContentMetadata.objects.count() == 3
```
bbaker6225, line 53: Same `assert` vs `assertEqual` comment from before. Also, maybe it's overboard and others disagree, but I think it'd be nice if the test also checked the actual value of one of the new content metadata objects

bbaker6225 (commented):
Few small comments, but looks good overall

enterprise_catalog/apps/api_client/discovery.py
christopappas: Unfortunately no. If you set a default to a dictionary in the function definition, there will only ever be 1 dictionary used throughout the runtime for that variable (the same is true of lists) (better explanation: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html)

bbaker6225: Ugh right I forgot about that

enterprise_catalog/apps/catalog/tests/test_models.py
christopappas: Yeah I can check more things for sure here, will do

bbaker6225 (approved):
LGTM

enterprise_catalog/apps/api_client/discovery.py
adamstankiewicz: Interesting. Thanks for sharing that explanation :)

enterprise_catalog/settings/devstack.py
adamstankiewicz: [Here's] (https://github.com/edx/configuration/blob/18342be5b26384a4f422e6c529dec29984d9d375/playbooks/roles/enterprise_catalog/defaults/main.yml#L128) where the `BACKEND_SERVICE_EDX_OAUTH2_KEY` env variable is set in Ansible.

adamstankiewicz (approved):

chavesj (approved):

enterprise_catalog/apps/catalog/models.py
```diff
@@ -177,3 +178,23 @@ def __str__(self):
...
+    """
+    client = DiscoveryApiClient()
+
+    catalog = EnterpriseCatalog.objects.get(uuid=catalog_uuid)
+    metadata = client.get_metadata_by_query(catalog.catalog_query.content_filter)
+
+    for entry in metadata.get('results', []):
+        defaults = {'content_key': entry['key']}
+        ContentMetadata.objects.update_or_create(
+            json_metadata=entry,
```
brittneyexline, line 198: some things we should ensure we're parsing out here to set on the model are content type and parent content key (should only apply to course runs).

PR #43, opened 2020-01-21
Add contains_content_items endpoint

bbaker6225:
## Ticket Link

Link to the associated ticket
[ENT-2563](https://openedx.atlassian.net/browse/ENT-2563)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/decorators.py
+++ b/enterprise_catalog/apps/api/v1/decorators.py
@@ -1,0 +1,42 @@
+from functools import wraps
+
+from rest_framework.exceptions import ValidationError
+
+
+def require_at_least_one_query_parameter(*query_parameter_names):
+    """
+    Ensure at least one of the specified query parameters are included in the request.
+
+    This decorator checks for the existence of at least one of the specified query
+    parameters and passes the values as function parameters to the decorated view.
+    If none of the specified query parameters are included in the request, a
+    ValidationError is raised.
+
+    Usage::
+        @require_at_least_one_query_parameter('program_uuids', 'course_run_ids')
+        def my_view(request, program_uuids, course_run_ids):
+            # Some functionality ...
+    """
+    def outer_wrapper(view):
+        """ Allow the passing of parameters to require_at_least_one_query_parameter. """
+        @wraps(view)
+        def wrapper(request, *args, **kwargs):
+            """
+            Checks for the existence of the specified query parameters, raises a
+            ValidationError if none of them were included in the request.
+            """
+            requirement_satisfied = False
+            for query_parameter_name in query_parameter_names:
+                query_parameter_values = request.query_params.getlist(query_parameter_name)
+                kwargs[query_parameter_name] = query_parameter_values
+                if query_parameter_values:
+                    requirement_satisfied = True
+            if not requirement_satisfied:
+                raise ValidationError(
+                    detail='You must provide at least one of the following query parameters: {params}.'.format(
+                        params=', '.join(query_parameter_names)
+                    )
+                )
+            return view(request, *args, **kwargs)
+        return wrapper
+    return outer_wrapper
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -173,7 +173,7 @@ class EnterpriseCatalogViewSetTests(APITestCase):
 
     def _assert_correct_response(self, url, expected_value):
         """
-        Helper to asssert that the contains_content_items endpoint specified by the url returns the correct value
+        Helper to assert that the contains_content_items endpoint specified by the url returns the correct value
         """
         response = self.client.get(url)
         self.assertEqual(response.json()['contains_content_items'], expected_value)
@@ -211,7 +211,7 @@ class EnterpriseCatalogViewSetTests(APITestCase):
         """
         child_key = 'child-key'
         parent_metadata = ContentMetadataFactory(content_key='parent-key')
-        ContentMetadataFactory(content_key='child-key', parent_content_key=parent_metadata.content_key)
+        ContentMetadataFactory(content_key=child_key, parent_content_key=parent_metadata.content_key)
         # Link the catalog to the parent metadata
         self.enterprise_catalog.catalog_query.contentmetadata_set.add(parent_metadata)
 
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -8,6 +8,7 @@ from rest_framework.test import APITestCase
 from enterprise_catalog.apps.catalog.models import EnterpriseCatalog
 from enterprise_catalog.apps.catalog.tests.factories import (
     USER_PASSWORD,
+    ContentMetadataFactory,
     EnterpriseCatalogFactory,
     UserFactory,
 )
@@ -58,10 +59,10 @@ class EnterpriseCatalogViewSetTests(APITestCase):
 
     def test_list(self):
         """
-        Verify the endpoint returns a list of all enterprise catalogs
+        Verify the viewset returns a list of all enterprise catalogs
         """
         url = reverse('api:v1:enterprise-catalog-list')
-        second_enterprise_catalog = EnterpriseCatalogFactory(catalog_query__content_filter='{"content_type":"course"}')
+        second_enterprise_catalog = EnterpriseCatalogFactory()
         response = self.client.get(url)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.data['count'], 2)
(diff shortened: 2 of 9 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -116,6 +117,50 @@ def __str__(self):
...
+        Arguments:
+            content_id_values (set): Return only the content IDs that are members of this set
+                                      and members of the catalog.
+
+        Returns:
+            set: The set of content IDs that are members of the catalog.
+        """
+        all_metadata = self.catalog_query.contentmetadata_set.all()
+        return {metadata_chunk.content_key for metadata_chunk in all_metadata
+                if metadata_chunk.content_key in content_id_values}
```
bbaker6225: If we were to add `or metadata_chunk.parent_content_key in content_id_values` to this set comprehension, then we would be filtering to the content keys and their children that are in the catalog

enterprise_catalog/apps/api/v1/decorators.py
```diff
@@ -0,0 +1,42 @@
+from functools import wraps
```
bbaker6225, line 1: This file is just copied over from `edx-enterprise`

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -16,15 +23,30 @@
...
-        if action == 'create':
+        request_action = getattr(self, 'action', None)
+        if request_action == 'create':
             return EnterpriseCatalogCreateSerializer
 
         return EnterpriseCatalogSerializer
+
+    @method_decorator(require_at_least_one_query_parameter('course_run_ids', 'program_uuids'))
+    @action(detail=True)
+    def contains_content_items(self, request, uuid, course_run_ids, program_uuids):
```
bbaker6225, line 40: It's annoying to still support both query params even though they both have the same effect now for backwards compatibility, but oh well

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -157,9 +158,66 @@ def test_post(self):
...
+        """
+        response = self.client.get(self._get_contains_content_base_url(self.enterprise_catalog))
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+
+    def test_contains_content_items_no_catalog_query(self):
+        """
+        Verify the contains_content_items endpoint returns false if there is no associated catalog query
+        """
+        no_catalog_query_catalog = EnterpriseCatalogFactory(catalog_query=None)
+        url = self._get_contains_content_base_url(no_catalog_query_catalog) + '?program_uuids=test-uuid'
```
christopappas, line 193: For building your URLs, you can hand reverse args or kwargs as well to avoid string concatenation. Something like:

`reverse('api:v1:enterprise-catalog-contains-content-items', kwargs={'program_uuids': 'test-uuid'})`

(https://docs.djangoproject.com/en/3.0/ref/urlresolvers/#reverse)

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -157,9 +158,66 @@ def test_post(self):
...
+        response = self.client.get(url)
```
christopappas: What type of object is response in this case?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -16,15 +23,30 @@
...
-    serializer_class = EnterpriseCatalogSerializer
     queryset = EnterpriseCatalog.objects.all().order_by('created')
     permission_classes = [permissions.IsAuthenticated, permissions.IsAdminUser]
     authentication_classes = [JwtAuthentication, BearerAuthentication, SessionAuthentication]
     lookup_field = 'uuid'
 
     def get_serializer_class(self):
-        action = getattr(self, 'action', None)
-        if action == 'create':
+        request_action = getattr(self, 'action', None)
```
christopappas, line 32: Curious about why the change from `action` to `request_action`, as `action` tends to have meaning within DRF itself (https://www.django-rest-framework.org/api-guide/viewsets/#viewset-actions)

enterprise_catalog/apps/api/v1/tests/test_views.py
christopappas: I read a little further. Probably personal preference but to me scanning for a substring feels less intuitive than using a response's json() method `assert response.json()['contains_content_items'] is False`

enterprise_catalog/apps/catalog/models.py
```diff
@@ -116,6 +116,32 @@ def __str__(self):
...
+        for content_key in content_keys:
+            try:
+                parent_content_key = ContentMetadata.objects.get(content_key=content_key).parent_content_key
+            except ContentMetadata.DoesNotExist:
+                parent_content_key = None
+            # The content key is contained in the catalog if its key is explictly part of the associated metadata, or
+            # its parent's key is.
+            contained_in_catalog = contained_in_catalog and (content_key in associated_metadata_content_keys
+                or parent_content_key in associated_metadata_content_keys)
+
```
christopappas, line 147: How long do we think content_keys can get? Would it make sense to break/return early if `contained_in_catalog` is False?

enterprise_catalog/apps/api/v1/views.py
christopappas: Do we need to worry about quoting/unquoting program uuids like we do course run ids? The answer is probably no, but it jumped out at me seeing course_run_ids being processed to some extent.

enterprise_catalog/apps/catalog/models.py
bbaker6225: Sure, I can make it break early

enterprise_catalog/apps/api/v1/views.py
bbaker6225: quality was complaining about redefining `action` (imported from drf)

christopappas: okay 🤦‍♂

enterprise_catalog/apps/api/v1/tests/test_views.py
bbaker6225: Unfortunately I'm pretty sure `kwargs` can't be used for query params (or at least I have not figured out how it can)

enterprise_catalog/apps/api/v1/views.py
bbaker6225: Nah the dashes/hyphens in program uuids work fine for me in testing, as opposed to the pluses in course run ids.
Mostly copied from `edx-enterprise` also: https://github.com/edx/edx-enterprise/blob/30a3a930358e059fa469fdfadd1ddbb084cb29a1/enterprise/api/v1/views.py#L332-L333

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -157,9 +158,72 @@ def test_post(self):
...
+
+    def _get_contains_content_base_url(self, enterprise_catalog):
+        """
+        Helper to construct the base url for the contains_content_items endpoint
+        """
+        return reverse('api:v1:enterprise-catalog-contains-content-items', kwargs={'uuid': enterprise_catalog.uuid})
+
+    def _assert_correct_response(self, url, expected_value):
+        """
+        Helper to asssert that the contains_content_items endpoint specified by the url returns the correct value
```
adamstankiewicz: nit: little typo in the docstring -- "assert"

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -157,9 +158,72 @@ def test_post(self):
...
+        url = self._get_contains_content_base_url(self.enterprise_catalog) + '?course_run_ids=' + content_key
+        self._assert_correct_response(url, True)
+
+    def test_contains_content_items_parent_keys_in_catalog(self):
+        """
+        Verify the contains_content_items endpoint returns True if the parent's key is in the catalog
+        """
+        child_key = 'child-key'
+        parent_metadata = ContentMetadataFactory(content_key='parent-key')
+        ContentMetadataFactory(content_key='child-key', parent_content_key=parent_metadata.content_key)
```
adamstankiewicz: nit: would it make sense to use `child_key` variable instead of hardcoded string here?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -16,15 +23,30 @@
...
+        if request_action == 'create':
             return EnterpriseCatalogCreateSerializer
 
         return EnterpriseCatalogSerializer
+
+    @method_decorator(require_at_least_one_query_parameter('course_run_ids', 'program_uuids'))
+    @action(detail=True)
```
adamstankiewicz, line 39: Is there any difference between the `@action` decorator and the `@detail_route` decorator? Looks like we use the [`@detail_route` decorator in edx-enterprise](https://github.com/edx/edx-enterprise/blob/23af86a57a944da2839ad50d348adc403158fa13/enterprise/api/v1/views.py#L139).

bbaker6225: `@action` is the newer version
https://www.django-rest-framework.org/community/3.9-announcement/#action-decorator-replaces-list_route-and-detail_route

enterprise_catalog/apps/api/v1/tests/test_views.py
bbaker6225: yes it would

christopappas (approved):

PR #47, opened 2020-01-23
Chavesj/ent 2536

chavesj:
## Description

Exposed API endpoint to trigger the update of discovery metadata for a given uuid

## Ticket Link

Link to the associated ticket
https://openedx.atlassian.net/browse/ENT-2536

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -1,0 +1,11 @@
+from celery import shared_task
+
+from enterprise_catalog.apps.catalog.models import (
+    update_contentmetadata_from_discovery,
+)
+
+
+@shared_task(bind=True)
+# pylint: disable=unused-argument
+def update_catalog_metadata_task(self, catalog_uuid):
+    update_contentmetadata_from_discovery(catalog_uuid)
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -271,7 +271,7 @@ class EnterpriseCatalogViewSetTests(APITestCase):
         """
         Verify the refresh_metadata endpoint successfully updates the catalog metadata with a post request
         """
-        url = reverse('api:v1:update-ent-catalog', kwargs={'uuid': self.enterprise_catalog.uuid})
+        url = reverse('api:v1:update-enterprise-catalog', kwargs={'uuid': self.enterprise_catalog.uuid})
         response = self.client.post(url)
         self.assertEqual(response.status_code, status.HTTP_200_OK)
 
@@ -279,7 +279,7 @@ class EnterpriseCatalogViewSetTests(APITestCase):
         """
         Verify the refresh_metadata endpoint does not update the catalog metadata with a get request
         """
-        url = reverse('api:v1:update-ent-catalog', kwargs={'uuid': self.enterprise_catalog.uuid})
+        url = reverse('api:v1:update-enterprise-catalog', kwargs={'uuid': self.enterprise_catalog.uuid})
         response = self.client.get(url)
         self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
 
@@ -289,6 +289,6 @@ class EnterpriseCatalogViewSetTests(APITestCase):
         """
         catalog_uuid = self.enterprise_catalog.uuid
         EnterpriseCatalog.objects.all().delete()
-        url = reverse('api:v1:update-ent-catalog', kwargs={'uuid': catalog_uuid})
+        url = reverse('api:v1:update-enterprise-catalog', kwargs={'uuid': catalog_uuid})
         response = self.client.post(url)
-        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
+        self.assertEqual(response.status_code, status.HTTP_404_NOT_FOUND)
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -1,6 +1,7 @@
 import uuid
 from collections import OrderedDict
 
+from django.test import override_settings
 from rest_framework import status
 from rest_framework.reverse import reverse
 from rest_framework.test import APITestCase
@@ -227,3 +228,30 @@ class EnterpriseCatalogViewSetTests(APITestCase):
 
         url = self._get_contains_content_base_url(self.enterprise_catalog) + '?course_run_ids=' + 'test-key'
         self._assert_correct_response(url, False)
+
+    @override_settings(task_always_eager=True)
+    def test_refresh_catalog_on_post_returns_200_ok(self):
+        """
+        Verify the refresh_metadata endpoint successfully updates the catalog metadata with a post request
+        """
+        url = reverse('api:v1:update-ent-catalog', kwargs={'uuid': self.enterprise_catalog.uuid})
+        response = self.client.post(url)
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
+
+    def test_refresh_catalog_on_get_returns_405_not_allowed(self):
+        """
+        Verify the refresh_metadata endpoint does not update the catalog metadata with a get request
+        """
+        url = reverse('api:v1:update-ent-catalog', kwargs={'uuid': self.enterprise_catalog.uuid})
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, status.HTTP_405_METHOD_NOT_ALLOWED)
+
+    def test_refresh_catalog_on_invalid_uuid_returns_400_bad_request(self):
+        """
+        Verify the refresh_metadata endpoint returns an HTTP_400_BAD_REQUEST status when passed an invalid ID
+        """
+        catalog_uuid = self.enterprise_catalog.uuid
+        EnterpriseCatalog.objects.all().delete()
+        url = reverse('api:v1:update-ent-catalog', kwargs={'uuid': catalog_uuid})
+        response = self.client.post(url)
+        self.assertEqual(response.status_code, status.HTTP_400_BAD_REQUEST)
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -265,7 +265,7 @@ class EnterpriseCatalogViewSetTests(APITestCase):
         self.assertEqual(response.json()['title'], self.enterprise_catalog.title)
         self.assertEqual(uuid.UUID(response.json()['enterprise_customer']), self.enterprise_catalog.enterprise_uuid)
         self.assertEqual(response.json()['results'], [json_metadata_1, json_metadata_2])
-    
+
     @override_settings(task_always_eager=True)
     def test_refresh_catalog_on_post_returns_200_ok(self):
         """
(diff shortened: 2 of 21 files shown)


enterprise_catalog/apps/api/v1/urls.py
```diff
@@ -14,5 +15,11 @@
 router = DefaultRouter()  # pylint: disable=invalid-name
 router.register(r'enterprise-catalog', views.EnterpriseCatalogViewSet, basename='enterprise-catalog')
 
-urlpatterns = []
+urlpatterns = [
+    url(
+        r'^update-ent-catalog',
```
bbaker6225: I imagine we'd want this to just be a new detail action on the enterprise catalog viewset (you can look at the `contains_content_items` endpoint as an example)

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -50,3 +56,17 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids):
...
+    View to update metadata in Catalog with most recent data from Discovery service
+    """
+    @task(bind=True)
+    def post(self, request):
+        uuid = request.data.get("uuid")
+        # call update function
+        update_contentmetadata_from_discovery(uuid)
+        response_string = "Successfully updated metadata for catalog UUID "
+        response_string += str(uuid)
+        return Response(response_string, status=HTTP_200_OK)
```
bbaker6225: Personally I think we should just return a status code (and nothing else) based on the update task, but others might disagree

bbaker6225 (commented):
Few comments but the gist makes sense to me

enterprise_catalog/apps/api/v1/urls.py
brittneyexline: agreed^^

enterprise_catalog/apps/api/v1/views.py
brittneyexline: agreed^^

enterprise_catalog/celery.py
```diff
@@ -0,0 +1,25 @@
...
+# - namespace='CELERY' means all celery-related configuration keys
+#   should have a `CELERY_` prefix.
+app.config_from_object('django.conf:settings', namespace='CELERY')
+
+# Load task modules from all registered Django app configs.
+app.autodiscover_tasks()
+
+
+@app.task(bind=True)
+def debug_task(self):
```
brittneyexline: i'm not really familiar with celery, what is the purpose of defining this function?

brittneyexline (commented):
wouldn't hurt to write a test even though the functionality is fairly simple.

another thought on the mechanics of this api that are out of scope and i'm not sure how to approach: maybe down the road it would be good to be able to query the status of the task somehow. i've worked with an api that had a background task that ran, the initial call to start it would return an id, and then you could query against that id to see the status of the task. again, this comment is more food for thought than anything - i don't think we need that right now.

enterprise_catalog/celery.py
chavesj: good catch, it's something they include as an example in the celery starter code, forgot to remove it

enterprise_catalog/apps/api/v1/urls.py
chavesj: done

enterprise_catalog/apps/api/v1/views.py
chavesj: ✅

bbaker6225 (approved):
Not familiar with the celery config but seems fine to me.

I do agree with @brittneyexline that a test would be nice, even though it's simple

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -50,3 +55,12 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids):
...
         return Response({'contains_content_items': contains_content_items})
+
+    @action(detail=True)
+    @task(bind=True)
+    def update_catalog_metadata(self, request, uuid):
+        """
+        Update metadata in Catalog with most recent data from Discovery service
+        """
+        update_contentmetadata_from_discovery(uuid)
+        return Response(status=HTTP_200_OK)
```
adamstankiewicz: Is there a chance that `update_contentmetadata_from_discovery` will fail / throw an exception? If so, should this be wrapped in a try/catch and return the appropriate HTTP status code, as opposed to always returning 200 OK?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -50,3 +55,12 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids):
         enterprise_catalog = self.get_object()
         contains_content_items = enterprise_catalog.contains_content_keys(course_run_ids + program_uuids)
         return Response({'contains_content_items': contains_content_items})
+
+    @action(detail=True)
+    @task(bind=True)
+    def update_catalog_metadata(self, request, uuid):
```
brittneyexline: just want to ensure this is a POST call still - it's not obvious if it is or not.

brittneyexline (commented):
once the test is in, feel free to merge.

enterprise_catalog/apps/api/v1/views.py
brittneyexline: good catch here... i suppose it would be good to format the error message explicitly if an exception is thrown, as is this would return a 500 if there were an exception.

enterprise_catalog/apps/api/v1/views.py
bbaker6225: ah good catch. @chavesj You'll need another argument that says which http methods are allowed I believe

enterprise_catalog/apps/api/v1/views.py
chavesj: yes it will fail if the UUID is not found. catch exception by returning 404_NOT_FOUND?

bbaker6225: DRF will automatically handle the 404 for a not found uuid. You should just need something to indicate that the update job failed (other than a 500) if there is an exception there

enterprise_catalog/apps/api/v1/urls.py
bbaker6225: I think I made this more difficult than it needed to be for Jeff. All the cambridge folk had a lengthy discussion in person and I think we're now agreed that we want to do a `POST` on a different view since it's weirdly difficult to use a different serializer on a custom action

brittneyexline: hmm ok. my request would be to make the endpoint name structure as RESTful as possible, meaning following the conventions of structuring the endpoint.

christopappas:
First time actually i'm looking at this PR, but I think based on the things we've encountered, an APIview is probably the way to go here (and not a model view set in this case). As Jeff found, the modelviewset requires a serializer for the serialization of model fields when returning a response (and complains if it's not present. We've tried overriding and setting it to None). Makes me feel like we're fighting the framework and should use a different approach that's more suited to what we're trying to do.

I guess the one note would be that if we do go the APIView route (pun optional), would be to configure the url in the urls.py file to appear to be restful (with the resource name first etc)

enterprise_catalog/apps/api/tasks.py
```diff
@@ -0,0 +1,11 @@
+from celery import shared_task
+
+from enterprise_catalog.apps.catalog.models import (
+    update_contentmetadata_from_discovery,
+)
+
+
+@shared_task(bind=True)
+# pylint: disable=unused-argument
+def update_catalog_metadata_task(self, catalog_uuid):
```
bbaker6225, line 10: Does it need self?

enterprise_catalog/apps/api/v1/urls.py
```diff
@@ -14,5 +15,11 @@
 router = DefaultRouter()  # pylint: disable=invalid-name
 router.register(r'enterprise-catalog', views.EnterpriseCatalogViewSet, basename='enterprise-catalog')
 
-urlpatterns = []
+urlpatterns = [
+    url(
+        r'^enterprise-catalog/(?P<uuid>[\S]+)/refresh_metadata',
+        views.EnterpriseCatalogRefreshDataFromDiscovery.as_view(),
+        name='update-ent-catalog'
```
bbaker6225: Nit: I think it's worth going for the more verbose `update-enterprise-catalog`

enterprise_catalog/settings/base.py
```diff
@@ -235,15 +235,64 @@
...
+    CELERY_BROKER_PASSWORD,
+    CELERY_BROKER_HOSTNAME,
+    CELERY_BROKER_VHOST
+)
+
+# Celery task time limits.
+# Tasks will be asked to quit after four minutes, and un-gracefully killed
+# after five.
+# This should prevent UserTasks from getting stuck indefinitely in an
+# In-Progress/Pending state, which in the case of enrollment-writing tasks,
```
bbaker6225: I think the comments here could be updated since we don't do enrollment writing tasks

bbaker6225 (commented):
Not familiar with the celery configuration so I'll let someone else look at that. Just a few small comments on other things

enterprise_catalog/apps/api/tasks.py
chavesj: yes, this is something that they needed to use in registrar as well

enterprise_catalog/settings/base.py
chavesj: removed

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -83,3 +86,18 @@ def get_content_metadata(self, request, uuid, **kwargs):
...
+
+class EnterpriseCatalogRefreshDataFromDiscovery(APIView):
+    """
+    View to update metadata in Catalog with most recent data from Discovery service
+    """
+    def post(self, request, uuid):
+        # ensure catalog exists before starting celery task
+        if not EnterpriseCatalog.objects.filter(uuid=uuid):
+            # respond with 400 status if catalog doesn't exist
+            return Response(status=HTTP_400_BAD_REQUEST)
```
bbaker6225: This feels like a `404` situation to me

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -264,3 +265,30 @@ def test_get_content_metadata(self):
         self.assertEqual(response.json()['title'], self.enterprise_catalog.title)
         self.assertEqual(uuid.UUID(response.json()['enterprise_customer']), self.enterprise_catalog.enterprise_uuid)
         self.assertEqual(response.json()['results'], [json_metadata_1, json_metadata_2])
+    
+    @override_settings(task_always_eager=True)
+    def test_refresh_catalog_on_post_returns_200_ok(self):
```
bbaker6225, line 270: I'd suggest moving these tests to a new test class so that the `EnterpriseCatalogViewSet` test class doesn't balloon up *too* much

enterprise_catalog/apps/api/v1/urls.py
brittneyexline: agreed, regarding this name variable.

enterprise_catalog/apps/api/v1/views.py
brittneyexline: agreed^^

brittneyexline (approved):
looks good overall, with a few small changes brandon suggested that i second. once those are addressed, feel free to merge

PR #49, opened 2020-01-24
Add get_content_metadata endpoint on catalogs

bbaker6225:
ENT-2563

## Ticket Link

[ENT-2563](https://openedx.atlassian.net/browse/ENT-2563)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views.py
+++ b/enterprise_catalog/apps/api/v1/views.py
@@ -68,7 +68,7 @@ class EnterpriseCatalogViewSet(viewsets.ModelViewSet):
             'title': enterprise_catalog.title,
             'enterprise_customer': enterprise_catalog.enterprise_uuid,
             'count': 0,
-            'previous': None, 'next': None,  # Currently unsure how to handle these pieces of metadata
+            'previous': None, 'next': None,  # Kept for parity with edx-enterprise
             'results': [],
         }
 
--- a/enterprise_catalog/apps/api/v1/views.py
+++ b/enterprise_catalog/apps/api/v1/views.py
@@ -8,7 +8,9 @@ from edx_rest_framework_extensions.auth.jwt.authentication import (
 from rest_framework import permissions, viewsets
 from rest_framework.authentication import SessionAuthentication
 from rest_framework.decorators import action
+from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
+from rest_framework_xml.renderers import XMLRenderer
 from six.moves.urllib.parse import quote_plus, unquote
 
 from enterprise_catalog.apps.api.v1.decorators import (
@@ -18,7 +20,10 @@ from enterprise_catalog.apps.api.v1.serializers import (
     EnterpriseCatalogCreateSerializer,
     EnterpriseCatalogSerializer,
 )
-from enterprise_catalog.apps.catalog.models import EnterpriseCatalog
+from enterprise_catalog.apps.catalog.models import (
+    ContentMetadata,
+    EnterpriseCatalog,
+)
 
 
 class EnterpriseCatalogViewSet(viewsets.ModelViewSet):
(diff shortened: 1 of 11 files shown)


enterprise_catalog/apps/api/v1/views.py
```diff
@@ -18,14 +20,18 @@
...
+    EnterpriseCatalog,
+)
 
 
 class EnterpriseCatalogViewSet(viewsets.ModelViewSet):
     """ View for CRUD operations on Enterprise Catalogs """
     queryset = EnterpriseCatalog.objects.all().order_by('created')
     permission_classes = [permissions.IsAuthenticated, permissions.IsAdminUser]
     authentication_classes = [JwtAuthentication, BearerAuthentication, SessionAuthentication]
+    renderer_classes = [JSONRenderer, XMLRenderer]
```
bbaker6225, line 34: Added for parity with `edx-enterprise`, although it's kind of annoying to have to specify `.json` locally

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -50,3 +56,30 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids):
...
+        """
+        Returns all the content linked to the specified catalog, ordered by content key.
+        """
+        enterprise_catalog = self.get_object()
+        metadata = {
+            'uuid': enterprise_catalog.uuid,
+            'title': enterprise_catalog.title,
+            'enterprise_customer': enterprise_catalog.enterprise_uuid,
+            'count': 0,
+            'previous': None, 'next': None,  # Currently unsure how to handle these pieces of metadata
```
bbaker6225: We can talk more about handling this and the child/parent content key link in grooming today

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -50,3 +56,30 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids):
         enterprise_catalog = self.get_object()
         contains_content_items = enterprise_catalog.contains_content_keys(course_run_ids + program_uuids)
         return Response({'contains_content_items': contains_content_items})
+
+    @action(detail=True)
+    def get_content_metadata(self, request, uuid, **kwargs):
```
christopappas, line 61: is `uuid` actually used in this method?

enterprise_catalog/apps/api/v1/views.py
christopappas: As far as next and previous go, I think it's fine to leave them as none. We're effectively "traversing pagination" when we hand them back all the content metadata for a given catalog, so it would all be one page conceptually (or at least how I think of it

christopappas (approved):
I think this looks good. I think the question about parent content key linking is an issue we should definitely think about/discuss, but I think we can do that after merging this

enterprise_catalog/apps/api/v1/views.py
bbaker6225: uuid is only used in the sense that it gets uuid from the detail route
i.e. `api/v1/enterprise-catalog/fake-uuid/get_content_metadata/` passes `fake-uuid` as uuid)

enterprise_catalog/apps/api/v1/views.py
bbaker6225: I guess that makes sense to me. Just wasn't sure whether or not we should be concerned about metadata size & keeping the backwards compatibility with `previous/next` from `edx-enterprise`. Although anything that can handle multiple pages of results should be able to handle one so that last point probably doesn't matter

enterprise_catalog/apps/api/v1/views.py
christopappas: i guess i'm just surprised pylint didn't yell at you ><'

enterprise_catalog/apps/api/v1/views.py
christopappas: `Although anything that can handle multiple pages of results should be able to handle one so that last point probably doesn't matter` <- this convinces me that we're probably fine

Yeh currently I think we do hand a mega json blob rn 😬 

If we did want paging, we could tweak the endpoint later too, by creating a ContentMetadata serializer, and then using the catalog query to create a queryset of ContentMetadata objects to serialize

PR #50, opened 2020-01-28
Add contains_content_items on enterprise customer

bbaker6225:
## Description

Add `contains_content_items` endpoint for "enterprise customer". (Really just looks at all the catalogs that share an enterprise uuid since we don't have an `EnterpriseCustomer` model in this IDA) 

## Ticket Link

Link to the associated ticket
[ENT-2576](https://openedx.atlassian.net/browse/ENT-2576)

## Post-review


enterprise_catalog/apps/api/v1/views.py
```diff
@@ -83,3 +82,35 @@ def get_content_metadata(self, request, uuid, **kwargs):
...
+
+
+class EnterpriseCustomerViewSet(viewsets.ViewSet):
+    """
+    Viewset for operations on enterprise customers.
+
+    Although we don't have a specific EnterpriseCustomer model, this viewset handles operations that use an enterprise
+    identifier to perform operations on their associated catalogs, etc.
+    """
+    permission_classes = [permissions.IsAuthenticated, permissions.IsAdminUser]
```
brittneyexline, line 111: just a note here on the intersection with @adamstankiewicz's work - doesn't have to be addressed in this pr but want to make sure we add permissions as a quick follow up.

brittneyexline (approved):
looks good! i leave it to you and adam to figure out the best way to follow up on adding permissions to this endpoint.

enterprise_catalog/apps/api/v1/views.py
adamstankiewicz: @brittneyexline @bbaker6225 Yeah, I will be updating my permissions branch to add similar permissions check for this `EnterpriseCustomerViewSet`. Also, the addition of the `APITestMixin` has some implications for the tests I'm wrapping up. I will rebase my branch on `master` now that this PR is merged and make updates accordingly.

PR #51, opened 2020-01-30
Set content_type and parent_content_key

bbaker6225:
## Description

Set `content_type` and `parent_content_key` on `ContentMetadata` objects when updating metadata from discovery

## Ticket Link
[ENT-2585](https://openedx.atlassian.net/browse/ENT-2585)

## Post-review


christopappas (approved):
looks good.

i like the test refactor as well 👍

PR #56, opened 2020-02-04
Introduce edx-rbac permissions via implicit JWT access

adamstankiewicz:
## Description

Introduces edx-rbac permissions.

## Ticket Link

[ENT-2534](https://openedx.atlassian.net/browse/ENT-2534)

## Post-review


docker-compose.yml
```diff
@@ -27,6 +27,7 @@ services:
     container_name: enterprise.catalog.app
     volumes:
       - .:/edx/app/enterprise_catalog/enterprise_catalog
+      - ../src:/edx/src:cached
```
adamstankiewicz, line 30: Adds the devstack `src` directory as a volume in the Docker container to be able to install packages from `src` directory into enterprise-catalog.

enterprise_catalog/apps/api/v1/tests/mixins.py
```diff
@@ -0,0 +1,112 @@
...
+        payload = generate_unversioned_payload(self.user)
+        payload.update({
+            'roles': [role_data]
+        })
+        jwt_token = generate_jwt_token(payload)
+
+        self.client.cookies[jwt_cookie_name()] = jwt_token
+
+
+class APITestMixin(JwtMixin, APITestCase):
```
adamstankiewicz, line 68: Moved from `utils.py` to `mixins.py`

enterprise_catalog/apps/catalog/admin.py
```diff
@@ -44,19 +42,6 @@ class ContentMetadataAdmin(UnchangeableMixin):
     list_display = ('id', 'content_key', 'content_type',)
 
 
-class CatalogQueryForm(forms.ModelForm):
```
adamstankiewicz, line 47: moved to forms.py

christopappas (approved):
looks like some commits are still coming through but overall i think it looks good 👍

adamstankiewicz:
> looks like some commits are still coming through but overall i think it looks good 👍

Yeah, just noticed a couple clean up things like docstrings. No additional commits coming through at this time :)

enterprise_catalog/apps/api/v1/tests/mixins.py
```diff
@@ -0,0 +1,112 @@
+# -*- coding: utf-8 -*-
+"""Broadly-useful mixins for use in automated tests."""
+from __future__ import absolute_import
```
bbaker6225: nit: I don't think we need this for python 3

enterprise_catalog/apps/api/v1/tests/mixins.py
```diff
@@ -0,0 +1,112 @@
...
+        self.client.logout()
+        non_staff_user = UserFactory()
+        self.client.login(username=non_staff_user.username, password=USER_PASSWORD)
+
+    def set_up_non_catalog_admin(self):
+        """
+        Helpr for logging in as a user that does not have the appropriate role(s) in the JWT
+        """
+        self.client.logout()
+        non_staff_user = UserFactory()
```
bbaker6225: I think it might make more sense for the user to be a staff user and not have the appropriate role

enterprise_catalog/apps/api/v1/urls.py
```diff
@@ -19,8 +19,9 @@
 urlpatterns = [
     url(
         r'^enterprise-catalog/(?P<uuid>[\S]+)/refresh_metadata',
-        views.EnterpriseCatalogRefreshDataFromDiscovery.as_view(),
+        views.EnterpriseCatalogRefreshDataFromDiscovery.as_view({'post': 'post'}),
```
bbaker6225, line 20: Did doing it in the `router.register` way we talked about not work? Not a big deal, but consistency would be nice

enterprise_catalog/settings/base.py
```diff
@@ -234,7 +253,7 @@
 LOGGING = get_logger_config(debug=DEBUG, dev_env=True, local_loglevel='DEBUG')
 
 EXTRA_APPS = []
-CERTIFICATE_LANGUAGES= {
+CERTIFICATE_LANGUAGES = {
```
bbaker6225, line 255: 💯

bbaker6225 (approved):
Few small comments but otherwise makes sense to me

enterprise_catalog/apps/api/v1/tests/mixins.py
adamstankiewicz: @bbaker6225 Yeah, I was deciding between the staff user without appropriate role or non-staff user. Happy to switch to using `self.user` here instead.

enterprise_catalog/apps/api/v1/urls.py
adamstankiewicz: > Did doing it in the `router.register` way we talked about not work? Not a big deal, but consistency would be nice

@bbaker6225 I was unable to get `router.register` to work properly in terms of using `reverse('api:v1:update-enterprise-catalog')`. From what I could gather from the docs, it seems that the router only works with viewsets, not views. 

`EnterpriseCatalogRefreshDataFromDiscovery` inherits from `APIView` which is a standard view, not a viewset. So, unless we change our use of `APIView` for `EnterpriseCatalogRefreshDataFromDiscovery`, we'll need to use the urlpatterns like we have here.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -29,10 +30,18 @@
 )
 
 
-class EnterpriseCatalogViewSet(viewsets.ModelViewSet):
+class BaseViewSet(PermissionRequiredMixin, viewsets.ViewSet):
+    """
+    Base class for all enterprise catalog view sets.
+    """
+    authentication_classes = [JwtAuthentication, BearerAuthentication, SessionAuthentication]
+    permission_classes = [permissions.IsAuthenticated, permissions.IsAdminUser]
```
brittneyexline: i think we can remove isAdminUser now that we've added the other permission stuff

brittneyexline (approved):
just one thing i noticed, looks good!

PR #59, opened 2020-02-10
ENT-2608 | Adding logic to metadata update function to associate Cont…

christopappas:
…entMetadata objects with their catalog query

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -234,11 +234,12 @@ def associate_content_metadata_with_query(metadata, catalog_query):
         content_key = get_content_key(entry)
         defaults = {
             'content_key': content_key,
+            'json_metadata': entry,
             'parent_content_key': get_parent_content_key(entry),
             'content_type': get_content_type(entry),
         }
         cm, __ = ContentMetadata.objects.update_or_create(
-            json_metadata=entry,
+            content_key=content_key,
             defaults=defaults,
         )
         LOGGER.info(
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -219,6 +219,57 @@ class ContentMetadata(TimeStampedModel):
             )
         )
 
+def associate_content_metadata_with_query(metadata, catalog_query):
+    """
+    get_or_create a content metadata object for entry in metadata
+    and then associate that object with the catalog_query provided.
+
+    metadata: Dictionary containing metadata
+    catalog_query: CatalogQuery object
+
+    Returns set of content_keys
+    """
+    content_keys = set()
+    for entry in metadata.get('results', []):
+        content_key = get_content_key(entry)
+        defaults = {
+            'content_key': content_key,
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -219,6 +219,57 @@ def __str__(self):
...
+    content_keys = set()
+    for entry in metadata.get('results', []):
+        content_key = get_content_key(entry)
+        defaults = {
+            'content_key': content_key,
+            'parent_content_key': get_parent_content_key(entry),
+            'content_type': get_content_type(entry),
+        }
+        cm, __ = ContentMetadata.objects.update_or_create(
+            json_metadata=entry,
```
bbaker6225: Just noticing this now: Shouldn't the `update_or_create` use `content_key` as the lookup field (as well as being in defaults) while json_metadata should be getting updated?

chavesj (approved):
:shipit:

enterprise_catalog/apps/catalog/tests/test_models.py
```diff
@@ -41,31 +41,54 @@ def test_contentmetadata_update_from_discovery(self, mock_client):
...
         self.assertEqual(ContentMetadata.objects.count(), 3)
 
+        associated_metadata = catalog.catalog_query.contentmetadata_set.all()
+
         # Assert stored content metadata is correct for each type
         course_cm = ContentMetadata.objects.get(content_key=course_metadata['key'])
         self.assertEqual(course_cm.content_type, COURSE)
         self.assertEqual(course_cm.parent_content_key, None)
         self.assertEqual(course_cm.json_metadata, course_metadata)
+        assert course_cm in associated_metadata
```
bbaker6225, line 71: 😞 y tho

christopappas: I'm not doing this to antagonize you I swear!

`assert course_cm in associated_metadata` reads like an english sentence. Do you really prefer `self.assertIn(course_cm, associated_metadata)`? At a glance, the latter is not clear me to me what should be in what

enterprise_catalog/apps/catalog/models.py
christopappas: I think you're right. Looking at the docs, the non "default" fields are the fields used for searching, and then the stuff in defaults are the values to be used when updating or creating.  Change incoming

enterprise_catalog/apps/catalog/tests/test_models.py
bbaker6225: haha no no I know. I agree plain `assert` reads a little better, but _usually_ I think the class methods often a bit better output on failures: 

> In addition, if first and second are the exact same type and one of list, tuple, dict, set, frozenset or str or any type that a subclass registers with addTypeEqualityFunc() the type-specific equality function will be called in order to generate a more useful default error message (see also the list of type-specific methods).

It's fine though, just a nit

bbaker6225 (approved):

PR #65, opened 2020-02-19
docker app name fix, settings updated

chavesj:
## Description

The webapp container name was updated in the docker-compose file to simplify the container information (previous name was redundant).

Django logging settings needed in devstack seem to suppress logging in production. The setting was taken out of base.py and replaced in devstack.py and production.py with the appropriate setting for each environment. 

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
     # ports:
     #  - "3307:3306"
 
-  enterprise_catalog:
+  app:
     # Uncomment this line to use the official catalog base image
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -15,7 +15,7 @@ services:
     # ports:
     #  - "3307:3306"
 
-  enterprise_catalog:
+  app:
```
adamstankiewicz, line 18: 💯

adamstankiewicz (approved):

PR #71, opened 2020-02-28
BOM-1347: Create Django 2.x Travis workers 

iamsobanjaved:
## Description

- Added django20,21,22 workers to travis
- Removed `SessionAuthenticationMiddleware` to fix djnago 2.x error failures

## Ticket Link

Link to the associated ticket
[BOM-1347](https://openedx.atlassian.net/browse/BOM-1347)

## Post-review

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,27 @@
 language: python
-python:
-    - 3.5
+
+matrix:
+  include:
+    - python: 3.5
+      env: TOXENV=django111
+      script: make test
+    - python: 3.5
+      env: TOXENV=django20
+      script: make test
+    - python: 3.5
+      env: TOXENV=django21
+      script: make test
+    - python: 3.5
+      env: TOXENV=django22
+      script: make test
+    - python: 3.5
+      env: TESTNAME=quality
+      script:
+        - make quality
+        - make pii_check
+    - python: 3.5
+      env: TESTNAME=translations
+      script: make validate_translations
 sudo: false
 branches:
   only:
@@ -10,11 +31,5 @@ cache: pip
 install:
     - pip install -U pip wheel codecov
     - make requirements
-script:
-    - make validate_translations
-    - make validate
-    - make pii_check
-    - make quality
-    - make test
 after_success:
     - codecov
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 .DEFAULT_GOAL := test
-
+TOX = ''
 .PHONY: help clean piptools requirements dev_requirements \
         doc_requirementsprod_requirements static shell test coverage \
         isort_check isort style lint quality pii_check validate \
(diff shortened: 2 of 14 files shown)


mraarif:
looks like tests for views are only failing because `SessionAuthenticationMiddleware` was removed in Django 2.0, we can fix these tests by just removing it from `MIDDLEWARE` attribute in settings.

mraarif (approved):

requirements/pip-tools.txt
```diff
@@ -5,5 +5,5 @@
 #    make upgrade
 #
 click==7.0                # via pip-tools
-pip-tools==4.5.0          # via -r requirements/pip-tools.in (line 3)
+pip-tools==4.5.1          # via -r requirements/pip-tools.in (line 3)
```
jmbowman: If you re-run `make upgrade` with this new version of pip-tools, it should get rid of the line numbers in the comments.  The diff for this PR will still be large because they're removed, but future diffs should be smaller and less prone to merge conflicts.

tox.ini
```diff
@@ -4,13 +4,14 @@ skipsdist=true
 envlist = {py35,py38}-django{111,20,21,22}
 
 [testenv]
+passenv = *
```
jmbowman: Is it a lot of extra work to figure out which environment variables need to be passed through?  I'm a little uncomfortable completely eliminating the separation between the calling environment and the tox execution environment.

tox.ini
```diff
@@ -4,13 +4,14 @@ skipsdist=true
...
 deps =
     django111: -r {toxinidir}/requirements/django.txt
     django20: Django>=2.0,<2.1
     django21: Django>=2.1,<2.2
     django22: Django>=2.2,<2.3
     -r{toxinidir}/requirements/test.txt
 whitelist_externals =
-    make
+    pycodestyle
```
jmbowman: Shouldn't pycodestyle be installed in the tox virtualenv?  If pylint and isort are, I'd expect pycodestyle to be there also.

iamsobanjaved: @jmbowman Its because isort, pylint are in test.txt as sub-dependencies while pycodestyle only exists in quality.txt. Shouldn't we exclude quality checks from being executed under all django versions, instead make another travis worker for it?

jmbowman:
Ok, it looks like edx-lint is in both `test.in` and `quality.in`, but pycodestyle is only in the latter.  I'm starting to wonder how much value there is in keeping them as separate files, but for now pycodestyle should probably be added to the testing dependencies.

tox.ini
jmbowman: Sure, that works also if you can take the time to refactor it.

Makefile
```diff
@@ -79,6 +83,10 @@ pii_check: ## check for PII annotations on all Django models
 
 validate: test quality pii_check ## run tests, quality, and PII annotation checks
 
+validate_python: test ## run tests
```
jmbowman: Could we just use `make test` directly instead of creating an alias for it?

.travis.yml
```diff
@@ -1,6 +1,25 @@
...
+      script: make validate_python
+    - python: 3.5
+      env: TOXENV=django21
+      script: make validate_python
+    - python: 3.5
+      env: TOXENV=django22
+      script: make validate_python
+    - python: 3.5
+      env: TESTNAME=quality
+      script: make validate_quality
```
jmbowman: Maybe it would be clearer to just call `make quality` and `make pii_check` directly?  Having both `quality` and `validate_quality` make targets seems confusing.

jmbowman (approved):

PR #75, opened 2020-03-12
OPS-4596 | Standard/Productionizing Dockerfile

christopappas:
## Description

Standardizing/Productionizing Dockerfile. We are moving towards deployment processes that will require each service to have a dockerfile that can be used in Production (but also devstack). This is PR related to that work, based off of work done on the notes api.

Big things:
- Be able to create and publish Dockerfiles for this service to Dockerhub
- Update dev env (in docker-compose and some code changes) to run `gunicorn` instead of `runserver` to have parity with how we run things in production

## Ticket Link

Link to the associated ticket

## Post-review


Dockerfile
```diff
@@ -1,10 +1,51 @@
...
+RUN make requirements
+
+# Code is owned by root so it cannot be modified by the application user.
+# So we copy it before changing users.
+USER app
+
+# This line is after the requirements so that changes to the code will not
+# bust the image cache
+COPY . /edx/app/enterprise_catalog/enterprise_catalog
+
```
coryleeio, line 57: Be sure to add the Gunicorn config and Gunicorn default command for a production centric file

Gunicorn handles large amounts of traffic better than a run server as the run server is single threaded to make debugging easier.

Notes should have an example of both of these things

Dockerfile
```diff
@@ -1,10 +1,51 @@
...
+pip3 install --upgrade pip setuptools && \
+rm -rf /var/lib/apt/lists/*
+
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+ENV LC_ALL en_US.UTF-8
+
 EXPOSE 18160
 EXPOSE 18161
```
coryleeio: I would expose the production centric ports instead of the dev centric ports

Keep in mind that the expose directive by itself doesn't actually do anything, it's just documentation.

I would document the production ready ports instead of the runserver ones, as these are typically only used in the development environment.  You can still use these development ports in the docker compose file.

In general think of the docker-compose file as the development part, and make the docker file for the production application when possible.

Dockerfile
christopappas: Alright 👍 -- yeah I wasn't sure if that was notes specific. I will look to adapt that

Dockerfile
christopappas: Awesome, will do

coryleeio: The reasoning for this is that if you misconfigure an application it's usually better to launch a battle hardened production config by default then to launch a dev server, which may have security / encryption turned off, maybe it will have debugger ports open, stuff like that

coryleeio: Exposing both development and production would also be reasonable, especially with a comment saying what is for what.

Typically we use like
18160
18161 for dev
8160
8161 for prod...

I would prefer them be the same, but that's not what we've done historically.

You could also just make the run server use the same ports as production and that would make a lot of this simpler...

Dockerfile
```diff
@@ -1,10 +1,59 @@
...
+#     ubuntu doesnt ship with python, so this is the python we will use to run the
+#     application
+# python3-pip
+#     install pip to install application requirements.txt files
+# libssl-dev
+#     mysqlclient wont install without this.
+# libmysqlclient-dev
+#     to install header files needed to use native C implementation for
+#     MySQL-python for performance gains.
+
```
christopappas, line 21: I think the above comments are more readable formatted like this. Hopefully I'm not breaking a linter rule or something..

enterprise_catalog/docker_gunicorn_configuration.py
```diff
@@ -0,0 +1,45 @@
+"""
+gunicorn configuration file: http://docs.gunicorn.org/en/develop/configure.html
+"""
+import multiprocessing
+
+preload_app = True
+timeout = 300
+bind = "0.0.0.0:8160"
```
christopappas, line 9: so we do have a celery worker container specified in docker-compose.yml, which builds using the same Dockerfile. Not sure how running the app with gunicorn affects this, but wanted to say words before I leave this afternoon so I remember to come back to it.

.travis.yml
```diff
@@ -1,35 +1,43 @@
...
-      env: TESTNAME=quality
-      script:
-        - make quality
-        - make pii_check
-    - python: 3.5
-      env: TESTNAME=translations
-      script: make validate_translations
+  - python: 3.5
+    env: TOXENV=django111
+    script: make test
```
jdmulloy:
```suggestion
```
This is an override, since most jobs are ```make test```we could make it the default script and only override the script on the few that are different

Maybe something more like this
```
python: 3.5

script:
  - make test

matrix:
  include:
    - env: TOXENV=django111
    - env: TOXENV=django20
    - env: TOXENV=django21
    - env: TOXENV=django22
    - env: TESTNAME=quality
      script:
        - make quality
        - make pii_check
    - env: TESTNAME=translations
      script:
        - make validate_translations
```

diff
```
diff --git a/.travis.yml b/.travis.yml
index c927441..575e1f5 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,23 @@
 language: python
+
+python: 3.5
+
+script:
+  - make test
+
 matrix:
   include:
-  - python: 3.5
-    env: TOXENV=django111
-    script: make test
-  - python: 3.5
-    env: TOXENV=django20
-    script: make test
-  - python: 3.5
-    env: TOXENV=django21
-    script: make test
-  - python: 3.5
-    env: TOXENV=django22
-    script: make test
-  - python: 3.5
-    env: TESTNAME=quality
-    script:
-    - make quality
-    - make pii_check
-  - python: 3.5
-    env: TESTNAME=translations
-    script: make validate_translations
+    - env: TOXENV=django111
+    - env: TOXENV=django20
+    - env: TOXENV=django21
+    - env: TOXENV=django22
+    - env: TESTNAME=quality
+      script:
+        - make quality
+        - make pii_check
+    - env: TESTNAME=translations
+      script:
+        - make validate_translations
 sudo: false
 branches:
   only:
```

Makefile
```diff
@@ -169,3 +169,20 @@ dev.stop: # Stops containers so they can be restarted
 
 attach:
 	docker attach enterprise.catalog.app
+
+travis_docker_auth:
```
jdmulloy: NIT: I would move this block below docker_build

jdmulloy (changes requested):

enterprise_catalog/docker_gunicorn_configuration.py
christopappas: i tested locally, and was able to confirm that the app container could talk to the worker container without anhy issues

enterprise_catalog/docker_gunicorn_configuration.py
```diff
@@ -0,0 +1,47 @@
+"""
```
christopappas, line 1: This is copy-pasted from https://github.com/edx/edx-notes-api/blob/master/notesserver/docker_gunicorn_configuration.py (with exception of changing the bind port)

Dockerfile
```diff
@@ -1,10 +1,67 @@
...
+# If you add a package here please include a comment above describing what it is used for
+RUN apt-get update && apt-get upgrade -qy && apt-get install language-pack-en locales git python3.5 python3-pip \
+python3-pip libmysqlclient-dev libssl-dev python3-dev -qy && \
+pip3 install --upgrade pip setuptools && \
+rm -rf /var/lib/apt/lists/*
+
```
coryleeio, line 27: Can you symlink python3 so you can just call 'python' instead of 'python3'
https://github.com/edx/ecommerce/blob/coryleeio/dockerfile/Dockerfile#L9-L10

Dockerfile
coryleeio: You're good, they're just comments, change as you like

Dockerfile
coryleeio: and pip, same thing

Dockerfile
```diff
@@ -1,10 +1,67 @@
...
+FROM app as devapp
+# Dev ports
 EXPOSE 18160
 EXPOSE 18161
-EXPOSE 18162
+USER root
+RUN pip3 install -r /edx/app/enterprise_catalog/enterprise_catalog/requirements/dev.txt
+USER app
+CMD gunicorn --reload --workers=2 --name enterprise_catalog -b :18160 -c /edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/docker_gunicorn_configuration.py --log-file - --max-requests=1000 enterprise_catalog.wsgi:application
```
coryleeio, line 69: You can remove the binding and port form here if you like -b.

It's in the config also, and so it's unnecessary to have it here and there.

coryleeio: I've been meaning to clean that up in notes

christopappas: Binding is necessary to override the prod port that is set in the config

christopappas: it's 8160 in the config file, but we want 18160 in dev env

Dockerfile
christopappas: cool, can do

coryleeio (approved):

Dockerfile
coryleeio: right, but this is the command used in the prod environment bc it's in the Dockerfile

coryleeio: 18160 is in docker-compose

coryleeio: This is gunicorn, which isn't used in dev

coryleeio: Ah yeah, so that port is wrong actually too

coryleeio: Ooooh this is devapp

coryleeio: So if thats the case, I would change the command to runserver here and keep the port

.travis.yml
```diff
@@ -1,35 +1,37 @@
...
   only:
-    - master
-
+  - master
 cache: pip
 install:
-    - pip install -U pip wheel codecov
-    - make requirements
+- pip install -U pip wheel codecov
+- make requirements
```
coryleeio, line 25: Since you have a dev image you could use the dev image for running the tests, it would use the correct version of python and whatnot then without you having to remember to update it in .travis

coryleeio: You don't have to do this, but it's worth considering

enterprise_catalog/settings/devstack.py
```diff
@@ -51,3 +51,6 @@
 CELERY_ALWAYS_EAGER = (
     os.environ.get("CELERY_ALWAYS_EAGER", "false").lower() == "true"
 )
+
+if 'console' not in LOGGING['loggers']['django']['handlers']:
```
coryleeio: You'll likely need to fix logging in production.py too, (I didn't look), I did it with settings in notes

coryleeio: but both should work for console logging, I had to change the get_logger function in production.py in notes to use settings instead of function parameters

https://github.com/edx/edx-notes-api/blob/master/notesserver/settings/yaml_config.py#L48

that's the function I had to change


It's possible you've already taken care of this in production.py, but if it is copied you'll run into it later when you deploy to prod so I am mentioning it.

Dockerfile
christopappas: so the motivation here is that I also changed docker-compose (albeit commented out right now) to use gunicorn in order to keep the dev and prod environments closer in how they are run. any reason we wouldn't want dev to use gunicorn?

coryleeio: Ah okay I see,

I'll let you decide, but the difference is roughly this:
the django runserver will restart when code changes are made, and is single threaded.
Gunicorn will not restart on changes, and when you attach to a process to debug you'll probably have multiple worker threads going.

christopappas: Cool. Yeah totally get that having multiple worker threads could be a point of trouble, although I did confirm the --reload flag worked on change with both worker processes ( pids 18 and 20, change observed and workers reloaded with pids 22 and 24):

Screen Shot 2020-03-18 at 1 22 17 PM

The reason I am also leaning towards using gunicorn locally is that it exposed a bunch of issues (related to admin css and logging), which, while in production it might not being an issue, I imagine that having that parity between dev and prod would be really helpful.

I'm gonna say, let's try going forward with gunicorn for now on dev and if we get surprised by a bunch of pain points, we can always switch back to runserver

(Provided of course enterprise team in question is cool with this when they do their review)

enterprise_catalog/settings/devstack.py
christopappas: yeah there is an analogous logging function called in base.py in the settings in this project, but DEBUG is set to False at the point which it is called, and therefore 'console' is not added to handlers when running the container locally.

I assumed that production logging happened via log files (which would require no changes in this PR) so instead of refactoring how/when the LOGGER is created, I added that one line to devstack.py to allow for logging to happen in context of the dev container running a gunicorn command

jdmulloy (approved):

Dockerfile
coryleeio: That sounds fine to me, so long as it doesn't cause any issues in dev I would say it is more ideal to have the parity also :+1:

enterprise_catalog/settings/devstack.py
coryleeio: Production logging is via console in K8s, and in docker in general typically

coryleeio: If you just have everything log to console you can decide what you want to do with the logs externally to the application

coryleeio: https://12factor.net
is an oft cited resource for this sort of thing.  It plays really nice with k8s.

#11 is about logging

#10 is about dev/prod parity which is also relevant. xD

Dockerfile
adamstankiewicz: @christopappas Yeah, testing this PR locally, I noticed that changes made in the code no longer triggers the app to restart. The downside of this is that we would have to manually restart the server for each code change made during development which could be a potential bottleneck. This was mentioned above by @coryleeio.

I do see that code changes are captured in the logs, where the worker reloads due to a modified file but it doesn't seem to trigger a restart:

image

Seems we'll now have to run `docker-compose restart app` to see code changes reflected in the running app if we proceed with using gunicorn locally as opposed to Django's runserver.

christopappas: hmm... mine does pick up code changes and I did not have to run a docker-compose command 🤔

adamstankiewicz: Just following up here based on the Slack discussion:

* `docker-compose restart app` is necessary to reload the app for settings changes
* hot reloads are working for non-settings code changes.

I think this will be fine moving forward, given that settings changes are much less frequent than non-settings changes 👍

adamstankiewicz (approved):

Dockerfile
```diff
@@ -1,10 +1,70 @@
...
+
 WORKDIR /edx/app/enterprise_catalog/enterprise_catalog
-ADD requirements.txt /edx/app/enterprise_catalog/enterprise_catalog/
-ADD Makefile /edx/app/enterprise_catalog/enterprise_catalog/
-ADD requirements/ /edx/app/enterprise_catalog/enterprise_catalog/requirements/
-RUN make requirements
-ADD . /edx/app/enterprise_catalog/enterprise_catalog
+
+COPY requirements/ /edx/app/enterprise_catalog/enterprise_catalog/requirements/
+RUN pip3 install -r /edx/app/enterprise_catalog/enterprise_catalog/requirements/base.txt
```
chavesj: do we need to install requirements from base here? I think everything from base.txt is replicated in production.txt

christopappas: had a look, looks like production.in requires base.txt, so yeah you are right, production.txt would ultimately have everything base.txt does

I can drop this line 👍

PR #76, opened 2020-03-24
Add learner access to content items/metadata apis

bbaker6225:
For new pages in the learner portal, we're allowing enterprise learner
access to the contains_content_items and get_content_metadata
endpoints.
Also changes the following:
* Add catalog learner role and permission rules around new role
* Refactor tests
* Add missing migrations
* Move contains_content_items and get_content_metadata to viewset
  separate from the enterprise catalog crud viewset

ENT-2687


enterprise_catalog/apps/api/v1/views.py
```diff
@@ -69,6 +69,27 @@ def get_permission_object(self):
             return str(enterprise_catalog.enterprise_uuid)
         return None
 
+
+class EnterpriseCatalogActionViewSet(PermissionRequiredMixin, BaseViewSet, viewsets.ModelViewSet):
```
brittneyexline: i like this separation 👍

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -29,19 +29,19 @@
 )
 
 
-class BaseViewSet(PermissionRequiredMixin, viewsets.ViewSet):
```
brittneyexline: not a strong opinion, but we could presumably keep this here and only define `permission_required` in the subclasses... not sure if there is a good way to enforce that it will be defined though.

enterprise_catalog/apps/catalog/admin.py
```diff
@@ -78,7 +78,7 @@ def get_username(self, obj):
 
     class Meta:
         """
-        Meta class for EnterpriseDataRoleAssignmentAdmin.
+        Meta class for EnterpriseCatalogRoleAssignmentAdmin.
```
brittneyexline, line 81: nice catch 👍

brittneyexline (approved):
looks good overall!

enterprise_catalog/apps/api/v1/views.py
bbaker6225: Good catch, I moved this on an earlier version and forgot about it

PR #79, opened 2020-03-27
Upgrade edx-rbac for Django 2.2 support

jmbowman:
## Description

Upgraded edx-rbac to a version that properly checks for Django 2.2 support.  Also pulled in assorted other dependency updates from the past 2 weeks.

## Ticket Link

[Support Django 2.2 in edx-rbac](https://openedx.atlassian.net/browse/BOM-1042)

## Post-review

### DIFF
--- a/requirements/constraints.txt
+++ b/requirements/constraints.txt
@@ -13,3 +13,6 @@ drf-jwt==1.14.0
 
 # jsonfield2 > 3.0.3 dropped support for python 3.5
 jsonfield2==3.0.3
+
+# 3.3.0 updates models with data from get_user_details(), which overrides is_superuser
+social-auth-core<3.3.0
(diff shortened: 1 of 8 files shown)


requirements/constraints.txt
```diff
@@ -13,3 +13,6 @@ drf-jwt==1.14.0
 
 # jsonfield2 > 3.0.3 dropped support for python 3.5
 jsonfield2==3.0.3
+
+# 3.3.0 updates models with data from get_user_details(), which overrides is_superuser
```
robrap, line 17: 1. Maybe add the related JIRA ticket in the comment so it is simpler to follow-up?
2. Is this constraint needed elsewhere as well?

robrap (approved):
One comment.

jmbowman:
Added a link to the ticket.  It probably should be pinned elsewhere also, this is being done for course-discovery and there are workarounds in place for the other IDAs that already upgraded it.  Followup work for that falls under the linked ticket.

PR #82, opened 2020-04-08
Restart containers as the last step of the provision script.

iloveagent57:
## Description

The provision script tries to restart the catalog app docker container by name, but docker-compose really wants the "service" name defined in the `docker-compose.yml` file for doing operations.  Hence, `docker-compose restart app`.  Also restarts the worker container, to make sure it's "in-sync" with the app container.

## Post-review


chavesj (approved):
:shipit:

bbaker6225 (approved):

adamstankiewicz (approved):

PR #85, opened 2020-04-13
ENT-2728: Fixing enterprise catalog 500 errors for anonymous users

chavesj:
## Description
Adding check for anonymous users where it currently fails causing a 500 status to be returned

## Ticket Link
[Link to the associated ticket](https://openedx.atlassian.net/browse/ENT-2728)

## Post-review


brittneyexline (commented):
looks simple enough! could you add a test?

enterprise_catalog/apps/catalog/rules.py
```diff
@@ -81,7 +81,7 @@ def has_explicit_access_to_catalog_learner(user, context):
     Returns:
         boolean: whether the request user has access or not
     """
-    if not context:
+    if not context or user.is_anonymous:
```
adamstankiewicz: I'd recommend having a similar check for `user.is_anonymous` within edx-rbac's `user_has_access_via_database` method such that a 500 doesn't get thrown for other consumers of edx-rbac, where if `user.is_anonymous` is truthy, `user_has_access_via_database` would return false.

That would be in place of having the check here in enterprise-catalog. That way, we can prevent 500 errors if/when other people use `user_has_access_via_database` and pass an `AnonymousUser` as well.

brittneyexline: agreed ^^ didn't realize this was in rules.py

chavesj: https://github.com/edx/edx-rbac/pull/42

PR #88, opened 2020-04-16
Docker shim fixes INFRA-2

jdmulloy:
## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review


coryleeio (approved):

bderusha (approved):

christopappas:
Have you spun up devstack and confirmed logging is still working? Specifically, do you know if DEBUG level logs are actually coming through to the console with devstack? (I can, just wanted to double check given that I don't love that logging util logic)

jdmulloy:
@christopappas good point. I have not tried it in devstack. I think that would still be ok since devstack already should have been skipping the local logger. I think I only removed stuff related to the local logger.

PR #89, opened 2020-04-22
Add api/api-compact api yaml files for API gateway

bbaker6225:
This will be used by api-manager to configure enterprise-catalog for
use in AWS API gateway.

ENT-2578

### DIFF
--- a/api-compact.yaml
+++ b/api-compact.yaml
@@ -1,0 +1,278 @@
+# Enterprise Catalog IDA: upstream API definitions, including vendor extensions.
+# Note, while this document is not strictly to Swagger spec, each HTTP method's
+# definition _must_ be to spec or the downstream ref will fail.
+#
+# Clients must include an Authorization header with OAuth2 access token in order to communicate:
+#   Authorization: JWT {access_token}
+#
+#
+# Available service endpoints -- note that alternate endpoints may be presented at the API Gateway tier
+#   /enterprise-catalog/v1/enterprise-catalogs/
+#   /enterprise-catalog/v1/enterprise-catalogs/{uuid}/
+
+apigateway_responses: &apigateway_responses
+  default:
+    statusCode: "400"
+  200:
+    statusCode: "200"
+  401:
+    statusCode: "401"
+  403:
+    statusCode: "403"
+  404:
+    statusCode: "404"
+  429:
+    statusCode: "429"
+  500:
+    statusCode: "500"
+
+apigateway_responses_with_mapping_template: &apigateway_responses_with_mapping_template
+  default:
+    statusCode: "400"
+  200:
+    statusCode: "200"
+    responseTemplates:
+      # Response body mapping template, this template is used for updating
+      # "next" and "previous" urls (both host and url path) while preserving
+      # the protocol (http | https) and querystring parameters.
+      application/json: >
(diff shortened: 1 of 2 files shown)


api-compact.yaml
```diff
@@ -0,0 +1,278 @@
...
+  httpMethod: "GET"
+  type: "http"
+  requestParameters:
+    integration.request.header.Authorization: "method.request.header.Authorization"
+    integration.request.path.uuid: "method.request.path.uuid"
+    integration.request.querystring.page: "method.request.querystring.page"
+    integration.request.querystring.page_size: "method.request.querystring.page_size"
+
+endpoints:
+  v1:
```
brittneyexline: should this be v2?

api-compact.yaml
```diff
@@ -0,0 +1,278 @@
...
+          produces: *produces
+          parameters:
+            - *auth_header
+            - *page_qs_parameter
+            - *page_size_qs_parameter
+          operationId: "get_enterprise_catalogs"
+          responses: *responses
+          x-amazon-apigateway-integration:
+            <<: *apigateway_integration_enterprise_catalog_list
+            uri: "https://${stageVariables.enterprise_host}/api/v1/enterprise-catalogs/"
```
brittneyexline, line 261: are we including this list endpoint because it exists for the edx-enterprise mapped endpoints? i'm not sure that we have tested if this behaves in the same way... which means we may have to revisit this endpoint and tweak it. my specific concern is around returning a filtered list of the enterprise catalogs linked to the user's enterprise.

brittneyexline (approved):
also wondering if contains_content_items is mapped at all for the current endpoints? looks like it may not be which is fine, just want to check.

api-compact.yaml
bbaker6225: Yes, I was including it to match the old one. You're correct though in that the enterprise catalog endpoint does not have that same filtering.
I'm a bit concerned that might be a larger fix though as we don't have any concept of `EnterpriseCustomer` and the users attached to it beyond a `uuid`.

api-compact.yaml
bbaker6225: I believe that this anchor refers to the API versioning internal to `enterprise-catalog`, i.e. all the endpoints below are at `enterprise-catalog.edx.org/api/v1/...`

bbaker6225: I'm happy to mark it as `v2` though for simplicity

bbaker6225:
@brittneyexline correct, `contains_content_items` is not mapped to API gateway in API manager

api-compact.yaml
bbaker6225: Updated

api-compact.yaml
brittneyexline: would you be able to file a ticket for adding that filtering so that we can get to it at some point? i think we can defer on that work in the immediate term since we don't need to transition external consumers of the endpoint immediately and the current version is still there.

brittneyexline (approved):
:shipit:

adamstankiewicz (approved):

api-compact.yaml
bbaker6225: Yep! https://openedx.atlassian.net/browse/ENT-2809

PR #92, opened 2020-04-28
Fix query parameters in discovery client

bbaker6225:
We're currently not seeing any query parameters showing up in splunk
for calls to the discovery api, which we aim to fix.
Based off of:
https://requests.readthedocs.io/en/master/user/quickstart/#passing-parameters-in-urls
we should be passing query parameters as a dictionary under the keyword
arg `params`.


adamstankiewicz (approved):
LGTM :shipit:

PR #95, opened 2020-05-04
Adds and updates content metadata fields to match edx-enterprise

adamstankiewicz:
## Description

Updates the json metadata contents for `ContentMetadata` objects to:
* include `enrollment_url` field
* append UTM query params to `marketing_url` field
* include `active` field

## Ticket Link

Link to the associated ticket
[ENT-2770](https://openedx.atlassian.net/browse/ENT-2770)

## Post-review


enterprise_catalog/apps/catalog/migrations/0019_add_enterprise_customer_name.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('catalog', '0018_update_jsonfield_formatting'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='enterprisecatalog',
+            name='enterprise_name',
+            field=models.CharField(blank=True, max_length=255, null=True),
```
adamstankiewicz: For the initial migration, `enterprise_name` is set to `blank=True` and `null=True` since none of the `EnterpriseCatalog` objects will have a known value for this field yet. Ideally, once all catalogs have an `enterprise_name` value, we can change this to `blank=False` and `null=False`.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -92,3 +116,56 @@ class EnterpriseCatalogCreateSerializer(EnterpriseCatalogSerializer):
...
+
+    def to_representation(self, instance):
+        """
+        Return the updated course data dictionary.
+
+        Arguments:
+            instance (dict): ContentMetadata object.
+
+        Returns:
+            dict: The updated course data.
```
iloveagent57: Nit: s/course/content/

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -92,3 +116,56 @@ class EnterpriseCatalogCreateSerializer(EnterpriseCatalogSerializer):
     UUID is writable to allow importing existing Enterprise Catalogs and keeping the same UUID
     """
     uuid = serializers.UUIDField(read_only=False, required=False)
+
+
+class ContentMetadataSerializer(ImmutableStateSerializer):
+    """
+    Serializer for the `ContentMetadata` model to add/modify additional
+    fields (e.g., enrollment_url, marketing_url)
```
iloveagent57: Nit: add something to the effect of "...of the serialized model instance" to the end of this sentence.  Otherwise, it's a little confusing to see this inherit from `ImmutableStateSerializer` but have a docstring that talks about modifying fields.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -92,3 +116,56 @@ class EnterpriseCatalogCreateSerializer(EnterpriseCatalogSerializer):
...
+        """
+        Return the updated course data dictionary.
+
+        Arguments:
+            instance (dict): ContentMetadata object.
+
+        Returns:
+            dict: The updated course data.
+        """
+        updated_metadata = copy.deepcopy(instance)
```
iloveagent57: Is there a missing `super()` call in here?  This line is making a copy of a `ContentMetadata` instance, right?  Should it actually be `updated_metadata = super().to_representation(instance)`?

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -1,8 +1,116 @@
...
+        'utm_medium': 'enterprise',
+    }
+
+    if enterprise_name:
+        utm_context['utm_source'] = slugify(enterprise_name)
+
+    return utm_context
+
+
+def parse_datetime_handle_invalid(datetime_value):
```
iloveagent57: totally-ignorable-nit: A naming pattern I like for things like this is `safe_do_the_thing`, so `safe_parse_datetime` in this example.

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -1,8 +1,116 @@
...
+    3. enrollment_end is greater than now, or undefined
+    """
+    now = datetime.datetime.now(pytz.UTC)
+    reasonable_enrollment_window = now + datetime.timedelta(days=1)
+    end = parse_datetime_handle_invalid(course_run.get('end'))
+    enrollment_start = parse_datetime_handle_invalid(course_run.get('enrollment_start'))
+    enrollment_end = parse_datetime_handle_invalid(course_run.get('enrollment_end'))
+    return (not end or end > reasonable_enrollment_window) and \
+           (not enrollment_start or enrollment_start < now) and \
+           (not enrollment_end or enrollment_end > now)
```
iloveagent57: Nit: Because my brain is really slow at parsing boolean expressions, it might be good to do something like this to avoid all the null-checking:
```python
course_end = parse_datetime_handle_invalid(course_run.get('end')) or datetime.datetime.max
enrollment_start = parse_datetime_handle_invalid(course_run.get('enrollment_start')) or datetime.datetime.min
enrollment_end = parse_datetime_handle_invalid(course_run.get('enrollment_end')) or datetime.datetime.max
return (reasonable_enrollment_window < course_end) and \
           (enrollment_start < now) and \
           (now < enrollment_end)
```

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -1,8 +1,116 @@
...
+
+
+def is_course_run_available_for_enrollment(course_run):
+    """
+    Check if a course run is available for enrollment.
+    """
+    if course_run['availability'] not in ['Current', 'Starting Soon', 'Upcoming']:
```
iloveagent57: Nit of the "weird python conventions" variety:  it's customary to put sets of allowed values in a tuple (or a set) instead of a list.

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -1,8 +1,116 @@
...
+        # course run is archived so not available for enrollment
+        return False
+
+    # now check if the course run is enrollable on the basis of enrollment
+    # start and end date
+    return is_course_run_enrollable(course_run)
+
+
+def has_course_run_available_for_enrollment(course_runs):
```
iloveagent57: naming suggestion: `is_any_course_run_enrollable(course_runs)`.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -164,6 +174,34 @@ def contains_content_keys(self, content_keys):
...
+            content_resource,
+            content_key,
+        )
+
+        url = update_query_parameters(url, get_enterprise_utm_context(self.enterprise_name))
+
+        if self.publish_audit_enrollment_urls:
+            url = update_query_parameters(url, {'audit': 'true'})
+
+        return update_query_parameters(url, {'catalog': self.uuid})
```
iloveagent57: Instead of calling `update_query_parameters()` 2 or 3 times, what if you just build up the additional params that have to be added to the url?  Like
```python
params = get_enterprise_utm_context(self.enterprise_name)
if self.publish_audit_enrollment_urls:
   params['audit'] = 'true'
params['catalog'] = self.uuid
return update_query_parameters(url, params)
```

iloveagent57 (commented):
On the DRF serializer comment - it's completely possible that I don't fully understand DRF serializers :)

enterprise_catalog/apps/api/v1/utils.py
adamstankiewicz: Yeah, the new methods in this file were largely copy/pasted from edx-enterprise. I didn't spend much time trying to figure out if they should be refactored. I like this refactoring, though, as it simplifies the logic. Will bring it in :)

enterprise_catalog/apps/api/v1/serializers.py
adamstankiewicz: From what I can tell, `super().to_representation(instance)` does not return a copy of the instance. Instead, it returns an empty OrderedDict with no key/value pairs. Since this serializer inherits from `ImmutableStateSerializer`, seems that calling `super().to_representation(instance)`, it wouldn't be aware of which model and/or fields to return.

`copy.deepcopy(instance)`, on the other hand, does a legit copy of the `ContentMetadata` instance.

iloveagent57: Oh, right, it's not a model serializer, so it will do pretty much nothing by default.  But still, `to_representation()` should generally convert an object (like a model instance) into a dict, or some other, primitive data structure: https://www.django-rest-framework.org/api-guide/serializers/#to_representationself-instance

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -92,3 +116,56 @@ class EnterpriseCatalogCreateSerializer(EnterpriseCatalogSerializer):
...
+
+        Arguments:
+            instance (dict): ContentMetadata instance.
+
+        Returns:
+            dict: The updated ContentMetadata object.
+        """
+        updated_metadata = copy.deepcopy(instance)
+        enterprise_catalog = self.context['enterprise_catalog']
+        content_type = updated_metadata.content_type
```
bbaker6225: I don't see any harm in having this in the response, but just wanted to point out that I don't think this is in the old endpoint's response

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -92,3 +116,56 @@ class EnterpriseCatalogCreateSerializer(EnterpriseCatalogSerializer):
     UUID is writable to allow importing existing Enterprise Catalogs and keeping the same UUID
     """
     uuid = serializers.UUIDField(read_only=False, required=False)
+
+
+class ContentMetadataSerializer(ImmutableStateSerializer):
```
bbaker6225: Mostly just a note to myself, but I think that with this serializer and with the changes on my branch, the pagination for `get_content_metadata` is almost ready to go with the exception of returning a dump of results from `traverse_pagination` or whatever the `edx-enterprise` mechanism for that is

enterprise_catalog/apps/catalog/migrations/0019_add_enterprise_customer_name.py
bbaker6225: I believe you should only need `blank=True` since it's a `CharField`. That way we'll also only have empty strings for the initially missing values of `enterprise_name` as opposed to empty string and `NULL`

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -125,11 +123,15 @@ def get_content_metadata(self, request, uuid, **kwargs):
         if not catalog_query:
             return Response(metadata)
 
-        associated_metadata = catalog_query.contentmetadata_set.all()
-        sorted_content_keys = sorted([metadata_chunk.content_key for metadata_chunk in associated_metadata])
-        metadata['results'] = [ContentMetadata.objects.get(content_key=content_key).json_metadata for content_key
-                               in sorted_content_keys]
-        metadata['count'] = len(sorted_content_keys)
+        context = self.get_serializer_context()
```
bbaker6225: Another note that almost all of this is changing on my branch, but should be fine for me to pull in the context and serializer and get rid of everything else

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -1,8 +1,115 @@
...
+    # now check if the course run is enrollable on the basis of enrollment
+    # start and end date
+    return is_course_run_enrollable(course_run)
+
+
+def is_any_course_run_enrollable(course_runs):
+    """
+    Iterates over all course runs to check if there any course run that is available for enrollment.
+
+    :param course_runs: list of course runs
```
bbaker6225: Nit: Different style of docstring here than other places

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -1,8 +1,115 @@
...
+    """
+    try:
+        if not isinstance(datetime_value, datetime.datetime):
+            datetime_value = parse_datetime(datetime_value)
+        return datetime_value.replace(tzinfo=pytz.UTC)
+    except TypeError:
+        return None
+
+
+def is_course_run_enrollable(course_run):
```
bbaker6225: One thing I think could be worthwhile is updating `course-discovery` to show the `is_enrollable` field for course runs on the `/search/all` endpoint like it is on the course runs endpoint: https://discovery.stage.edx.org/api/v1/course_runs/?fields=is_enrollable
That way we could drop a lot of this logic that is duplicated from discovery

enterprise_catalog/apps/api/v1/serializers.py
adamstankiewicz: The `content_type` here doesn't get included in the response. It's just used to figure out which version of the `enrollment_url` to use in the conditional logic below.

enterprise_catalog/apps/api/v1/utils.py
adamstankiewicz: I agree this makes sense to do. Here's a PR to add `is_enrollable` field for the `/search/all` endpoint: https://github.com/edx/course-discovery/pull/2590

As a result, I removed a lot of this logic to rely on the `is_enrollable` field instead.

enterprise_catalog/apps/api/v1/serializers.py
iloveagent57: ...so you could make it a dictionary like
```python
updated_metadata = {
    'content_type': instance.content_type,
    'json_metadata': copy.deepcopy(instance.json_metadata),
}
content_type = updated_metadata['content_type']
json_metadata = updated_metadata['json_metadata']
```
and I think everything else could stay the same.

iloveagent57 (commented):
This all looks good, just that one request to change the object returned by `to_representation` into a dict, instead of a model instance.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -125,11 +123,15 @@ def get_content_metadata(self, request, uuid, **kwargs):
...
+        context['enterprise_catalog'] = enterprise_catalog
+
+        associated_metadata = catalog_query.contentmetadata_set.order_by('content_key')
+        serializer = ContentMetadataSerializer(associated_metadata, context=context, many=True)
+        updated_associated_metadata = serializer.data
+
+        metadata['results'] = [metadata_obj.json_metadata for metadata_obj in updated_associated_metadata]
```
iloveagent57: whoops, one last change: `metadata_obj` is a dict now, so you want `[metadata_dict['json_metadata'] for metadata_dict in updated_associated_metadata]`

iloveagent57 (approved):
I _think_ that one last change will do it :)

enterprise_catalog/apps/api/v1/views.py
bbaker6225: Just a note that the method around this is changing a bunch in #97. I thought it'd be easier to implement this on top of that, but if you'd rather wrap this up, I can readjust

PR #100, opened 2020-05-12
Switch to GET request for `/search/all/` instead of POST

adamstankiewicz:
## Description

When fetching the metadata from discovery service, we are currently using a POST request and passing the `content_filter_query` as the data for the request.

Replicating the POST request in Postman using the parameters from an [existing `content_filter_query`](https://enterprise-catalog.edx.org/admin/catalog/catalogquery/20/change/), I get back 5 results. All 5 of these `ContentMetadata` objects are properly linked to the `CatalogQuery` object in question. This corresponds with the same 5 content items that get associated with the `CatalogQuery` when triggering an update to the same catalog in the LMS django admin, as seen by Splunk logs.

However, using a GET request in Postman with the same parameters from the `content_filter_query` as query parameters, I get back 103 results which matches the results given by the "Preview" link from the corresponding catalog in the LMS. However, all 103 results (minus the 5 mentioned above) **are not** associated with the `CatalogQuery` object in question.

Given this, I am hypothesizing that we should not be using a `POST` request to fetch the content metadata from discovery. Rather, I believe it should be a `GET` request. Conceptually, I'm not sure why this was initially a POST given that we're retrieving data, not creating new data.

This PR swaps over the POST request to a GET request.

**Testing Plan**

To test if this change makes a difference, I plan to trigger the `post_save` method on a catalog within the LMS django admin to update the corresponding catalog in the enterprise-catalog service. I will then monitor Splunk logs to determine which `ContentMetadata` items were associated with the `CatalogQuery`.

## Ticket Link

TBD

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api_client/discovery.py
+++ b/enterprise_catalog/apps/api_client/discovery.py
@@ -31,7 +31,7 @@ class DiscoveryApiClient:
     def oauth2_client_secret(self):
         return settings.BACKEND_SERVICE_EDX_OAUTH2_SECRET
 
-    def get_metadata_by_query(self, content_filter_query, query_params=None):
+    def get_metadata_by_query(self, content_filter_query):
         """
         Return results from the discovery service's search/all endpoint.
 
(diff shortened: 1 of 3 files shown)


adamstankiewicz:
Closing this PR as we confirmed that this request needs to stay as a POST due to some content filters being too long for a GET request. That said, we will need to figure out why the POST request is not returning all the results as described in the PR description above.

PR #110, opened 2020-05-18
Only update when json_metadata changes

bbaker6225:
Also removes old logging

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -277,6 +277,19 @@ class ContentMetadata(TimeStampedModel):
         )
 
 
+def get_sorted_string_from_json(json_metadata):
+    """
+    Get the string representing a json piece of metadata in alphabetical order for comparisons.
+
+    Arguments:
+        json_metadata (json): The json metadata of a particular piece of content metadata.
+
+    Returns:
+        string: The json metadata as a sorted string
+    """
+    return sorted(json.dumps(json_metadata))
+
+
 def associate_content_metadata_with_query(metadata, catalog_query):
     """
     get_or_create a content metadata object for entry in metadata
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -287,10 +300,17 @@ def associate_content_metadata_with_query(metadata, catalog_query):
 
     Returns set of content_keys
     """
-    number_of_items_that_have_not_changed = 0
     content_keys = set()
     for entry in metadata:
         content_key = get_content_key(entry)
+        if content_key in content_keys:
```
bbaker6225, line 306: Moved this to the top so we can just `continue` for metadata that hasn't changed. Every entry should have its content key added so doesn't change any logic

enterprise_catalog/apps/catalog/models.py
```diff
@@ -303,8 +323,9 @@ def associate_content_metadata_with_query(metadata, catalog_query):
             old_metadata = None
 
         if old_metadata:
-            if sorted(json.dumps(entry)) == sorted(json.dumps(old_metadata.json_metadata)):
-                number_of_items_that_have_not_changed += 1
+            if get_sorted_string_from_json(entry) == get_sorted_string_from_json(old_metadata.json_metadata):
+                # Only update the existing ContentMetadata object if its json has not changed
```
adamstankiewicz: nit: should this comment say "if its json **has** changed"?

adamstankiewicz (approved):
Just one nit about a comment, but otherwise LGTM.

enterprise_catalog/apps/catalog/models.py
bbaker6225: Oops, yes good catch

PR #111, opened 2020-05-18
Swap sets for lists in association/unassociation

bbaker6225:
Trying this out to see if it fixes the issues we're seeing in
inconsistent counts from edx-enterprise vs. this service

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review


adamstankiewicz (approved):
LGTM. The docstrings don't match the `collections.Counter` data type (i.e., still says `list`), but that's definitely a nit and can be cleaned up once we're done debugging the issue.

PR #112, opened 2020-05-18
simplify ContentMetadata to CatalogQuery association logic

adamstankiewicz:
## Description

This PR simplifies the association/dissociation logic into a single function to avoid needing to pass content_keys around between 2 functions when they were separate. It makes use of `.set(clear=True)` to remove all CatalogQuery -> ContentMetadata associations before adding the updated list.

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -1,5 +1,4 @@
 import collections
-import json
 from logging import getLogger
 from uuid import uuid4
 
@@ -27,6 +26,7 @@ from enterprise_catalog.apps.catalog.utils import (
     get_content_key,
     get_content_type,
     get_parent_content_key,
+    get_sorted_string_from_json,
 )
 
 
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -277,22 +278,9 @@ def __str__(self):
         )
 
 
-def get_sorted_string_from_json(json_metadata):
```
adamstankiewicz, line 280: moves to `enterprise_catalog/apps/catalog/utils.py`

enterprise_catalog/apps/catalog/models.py
```diff
@@ -327,52 +308,18 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
-            'Associating content_metadata %s with catalog_query %s.',
-            cm,
-            catalog_query
-        )
-        catalog_query.contentmetadata_set.add(cm)
-
-    return content_keys
+        metadata_list.append(cm)
 
+    catalog_query.contentmetadata_set.set(metadata_list, clear=True)
```
adamstankiewicz, line 328: Per the Django docs, `.set(clear=True)` will remove all prior relationships between the CatalogQuery's associated ContentMetadata objects before setting the new list from the result of the `for` loop.

adamstankiewicz: https://docs.djangoproject.com/en/2.2/ref/models/relations/#django.db.models.fields.related.RelatedManager.set

bbaker6225 (approved):
LGTM from a quick scan, great find on `set`

enterprise_catalog/apps/catalog/models.py
bbaker6225: Might be helpful to add a note about why we're using `clear` (i.e. that we want to make sure updated metadata are re-associated with metadata.) Just as I was initially a little confused why we would want to use `clear=False` which is similar, but doesn't clear the old ones first

enterprise_catalog/apps/catalog/models.py
```diff
@@ -327,52 +307,18 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
-                    catalog_query
-                )
-                catalog_query.contentmetadata_set.add(old_metadata)
+                metadata_list.append(old_metadata)
                 continue
 
         cm, __ = ContentMetadata.objects.update_or_create(
             content_key=content_key,
             defaults=defaults,
         )
```
iloveagent57: We already have the existing instance at this point (if it exists) in `old_metadata`.  We could eliminate an extra read here by doing:
```python
if old_metadata:
    for key, value in defaults.items():
        setattr(old_metadata, key, value)
    old_metadata.save()
else:
    ContentMetadata.objects.create(**defaults)
```

iloveagent57 (approved):
👍 Just the one suggestion.

PR #115, opened 2020-05-19
Add page_size query param support

bbaker6225:
Implemented through custom pagination class. This is done for parity
with edx-enterprise.

Example test call result:
Screenshot_2020-05-19 Screenshot

### DIFF
--- a/enterprise_catalog/apps/api/v1/pagination.py
+++ b/enterprise_catalog/apps/api/v1/pagination.py
@@ -1,0 +1,10 @@
+from rest_framework.pagination import PageNumberPagination
+
+
+class PageNumberWithSizePagination(PageNumberPagination):
+    """
+    Custom pagination that allows clients to pass a `page_size` query param.
+
+    Example usage: /api/v1/enterprise-catalogs/{uuid}/get_content_metadata/?page_size=1000
+    """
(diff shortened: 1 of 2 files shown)


chavesj (commented):
Might want to add a test for the new page_size param, besides that LGTM

bbaker6225:
I can add a test to the `get_content_metadata` tests to verify that it's using the correct pagination class, however I didn't feel testing the `page_size` param itself was particularly valuable since it's 3rd party code that we don't control
https://www.django-rest-framework.org/api-guide/pagination/#modifying-the-pagination-style

adamstankiewicz (approved):
LGTM. Do the other fields for `PageNumberPagination` also match up with edx-enterprise (i.e., default `page_size` and `max_page_size`)?

adamstankiewicz (commented):
Also, do we have any need to support the `page_size` in any of the other endpoints in enterprise-catalog, beyond `get_content_metadata`?

bbaker6225:
@adamstankiewicz yep, `max_page_size` isn't set in `edx-enterprise` and default `page_size` appears to be 10 from the tests I've run.
I'll add the `page_size` support to the crud views as well, since we might as well have it there too (and it doesn't apply to refresh metadata or contains content items)

PR #116, opened 2020-05-19
use correct content_key for course runs

adamstankiewicz:
## Description

The enrollment_url for course_runs that are part of a course is currently using the content_key of the course instead of the content_key for each of the course_runs.

## Post-review


enterprise_catalog/apps/catalog/models.py
```diff
@@ -312,7 +312,6 @@ def associate_content_metadata_with_query(metadata, catalog_query):
                 metadata_list.append(existing_metadata)
                 continue
 
-        if existing_metadata:
```
adamstankiewicz, line 315: unnecessary conditional

bbaker6225 (approved):

PR #117, opened 2020-05-26
increase page_size query parameter on /search/all request

adamstankiewicz:
## Description

Increases the `page_size` query parameter to 200 (from the default 20) due to non-deterministic pagination behavior within the /search/all endpoint, and likely the reason why we're seeing duplicate results from discovery.

## Ticket Link

N/A

## Post-review


chavesj (approved):
🚢

PR #119, opened 2020-05-28
an attempt to get consistent paginated results from search/all

adamstankiewicz:
## Description

Another attempt to address the non-deterministic pagination behavior of the /search/all endpoint. The endpoint enables ordering for the `start` field, but doesn't seem to apply that ordering by default. This query param may help to get consistent results when traversing pagination.

## Post-review


bbaker6225 (approved):
Worth a shot

PR #123, opened 2020-05-29
improve error handling and add logging around requests to /search/all/

adamstankiewicz:
## Description

Adding a try/catch block around requests to discovery to catch when it throws a `JsonDecodeError`, when the /search/all/ requests returns nothing (i.e., fails):

https://rpm.newrelic.com/accounts/88178/applications/512093214/filterable_errors#/table?top_facet=transactionUiName&primary_facet=error.class&barchart=barchart&filters=%5B%7B%22key%22%3A%22error.class%22%2C%22value%22%3A%22simplejson.errors%3AJSONDecodeError%22%2C%22like%22%3Afalse%7D%5D

Also, logs the error including the `CatalogQuery` to be able to figure out which catalog queries the issue is affecting for debugging purposes.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api_client/discovery.py
+++ b/enterprise_catalog/apps/api_client/discovery.py
@@ -3,10 +3,15 @@
 Discovery service api client code.
 """
 
+import logging
 from urllib.parse import urljoin
 
 from django.conf import settings
 from edx_rest_api_client.client import OAuthAPIClient
+from simplejson import JSONDecodeError
+
+
+LOGGER = logging.getLogger(__name__)
 
 
 class DiscoveryApiClient:
@@ -31,7 +36,7 @@ class DiscoveryApiClient:
     def oauth2_client_secret(self):
         return settings.BACKEND_SERVICE_EDX_OAUTH2_SECRET
 
-    def get_metadata_by_query(self, content_filter_query, query_params=None):
+    def get_metadata_by_query(self, catalog_query, query_params=None):
         """
         Return results from the discovery service's search/all endpoint.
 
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -45,22 +50,31 @@ def get_metadata_by_query(self, content_filter_query, query_params=None):
...
+            while response.get('next'):
+                page += 1
+                query_params.update({'page': page})
+                response = self.client.post(
+                    self.SEARCH_ALL_ENDPOINT,
+                    json=catalog_query.content_filter,
+                    params=query_params
+                ).json()
+                results += response.get('results', [])
+        except JSONDecodeError:
```
brittneyexline, line 73: should we consider broadening this exception, or add exceptions as we see them occur?

brittneyexline (approved):
one small question but not blocking

bbaker6225 (approved):
Seems reasonable to dig more

enterprise_catalog/apps/api_client/discovery.py
adamstankiewicz: I'm not sure I'm following the question. Do you mind clarifying? :)

brittneyexline: we're only catching this particular error, but others could pop up... it's possible the client is already handling some of these cases but it was more a question of, do we know of any other likely exceptions that we would want to catch here? or should we just wait until they happen?

adamstankiewicz: AFAIK, that is likely the only exception that might occur from this `try` block. Given that, I think I'd be in favor of waiting until other errors occur, if any, and adjust accordingly at that point.

adamstankiewicz:
[Build](https://travis-ci.org/github/edx/enterprise-catalog/builds/692597648) is green on travis, but the status is not reported back to Github. Merging regardless.

PR #124, opened 2020-05-29
run task only for queries used by at least one catalog

adamstankiewicz:
## Description

Right now, we have 118 unused `CatalogQuery` records in our DB (i.e., no associated `EnterpriseCatalog` records). Despite this, we still try to update those `CatalogQuery` records even though no catalog will be using those results.

To further reduce the number of requests/load to the `/search/all/` endpoint during the `update_content_metadata` job that runs daily, we can iterate through just the `CatalogQuery` records that are used by at least 1 `EnterpriseCatalog`.

Using today's numbers, we will reduce the number of tasks generated and requests to the `/search/all/` endpoint from 820 to 702. This will speed up the `update_content_metadata` job further and reduce load for course-discovery.

## Post-review


iloveagent57 (approved):
Good idea 👍

PR #126, opened 2020-05-31
refactor contains_content_keys to handle when content key exists in a course's course_runs

adamstankiewicz:
## Description

> May 31 18:28:45 ip-10-2-12-181 [service_variant=lms][enterprise.api.v1.serializers][env:prod-edx-edxapp] ERROR [ip-10-2-12-181  31166] [user 32845815] [serializers.py:703] - [Enterprise API] The course run id is not in the catalog for the Enterprise Customer. EnterpriseCustomer: University of Iceland, CourseRun: course-v1:IMFx+MRCx+1T2020

The contains_content_items call for this course run returns `True` on edx-enterprise but enterprise-catalog returns `False`.

After investigating, this error is valid as the course run is technically part of the catalog (the course run with the specified content key comes back as part of the “course_runs” list for the associated course in the get_content_metadata response.

In the contains_content_keys method, we should also check to see if the specified content key matches the key of any course runs for a course.

### Django Silk

This PR also installs [django-silk](https://github.com/jazzband/django-silk) locally to capture requests for SQL query and performance optimization debugging.

#### Instructions

1. `make dev.up.build`
1. `make app-shell` and then `./manage.py migrate silk` to run django-silk's migrations.
1. Browse to http://localhost:18160/silk/ for the Django Silk UI.
1. Make any request (e.g., http://localhost:18160/api/v1/enterprise-catalogs/3a18971e-dede-41d3-a7da-6db2dba9c80e/get_content_metadata) and check back on the Silk UI. You should see your request appear. You can now click on it to see the details about SQL queries and performance.

Example:

image

image

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -176,24 +176,30 @@ class EnterpriseCatalog(TimeStampedModel):
 
         content_keys = set(content_keys)
 
+        # construct a query on the associated catalog's content metadata to return metadata
+        # where content_key and parent_content_key matches the specified content_keys to
+        # handle the following cases where the catalog:
+        #   - contains courses and the specified content_keys are course ids
+        #   - contains course runs and the specified content_keys are course ids
+        #   - contains course runs and the specified content_keys are course run ids
+        #   - contains programs and the specified content_keys are program ids
+        query = Q(content_key__in=content_keys) | Q(parent_content_key__in=content_keys)
+        
         # retrieve content metadata objects for the specified content keys to get a set of
-        # parent content keys to use in the below query. this handles the case for when a
-        # catalog contains only courses but course run(s) are specified in content_keys.
+        # parent content keys, i.e. course ids associated with the specified content_keys
+        # (if any) to handle the following case:
+        #   - catalog contains courses and the specified content_keys are course run ids.
         searched_metadata = ContentMetadata.objects.filter(content_key__in=content_keys)
         parent_content_keys = {
             metadata.parent_content_key
             for metadata in searched_metadata
             if metadata.parent_content_key
         }
-
-        # find all content metadata associated with the catalog that match specified content_keys
-        query = Q(content_key__in=content_keys)
-        query |= Q(parent_content_key__in=content_keys)
         query |= Q(content_key__in=parent_content_keys)
 
-        content_metadata = self.catalog_query.contentmetadata_set.filter(query)
+        content_metadata = self.content_metadata.filter(query)
 
-        # if content metadata was returned, the specified content keys exist in the catalog
+        # if content metadata was returned, the specified content_keys exist in the catalog
         return bool(content_metadata)
 
     def get_content_enrollment_url(self, content_resource, content_key):
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -4,6 +4,7 @@ from uuid import uuid4
 
 from django.conf import settings
 from django.db import models
+from django.db.models import Q
 from django.utils.translation import gettext as _
 from edx_rbac.models import UserRole, UserRoleAssignment
 from jsonfield.encoder import JSONEncoder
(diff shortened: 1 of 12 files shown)


enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -60,10 +60,12 @@ class Meta:
 
     @factory.lazy_attribute
     def json_metadata(self):
-        return {
-            'key': self.content_key,
-            'marketing_url': 'http://marketing.yay/{}'.format(self.content_key),
-        }
+        json_metadata = {'key': self.content_key}
+        if self.content_key == COURSE:
+            json_metadata.update({
```
adamstankiewicz, line 65: `marketing_url` is only on `COURSE` content metadata objects.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -156,34 +156,50 @@ def content_metadata(self):
...
+        # iterate through content_metadata to find all associated content keys
+        for metadata_chunk in content_metadata:
+            # content key of metadata_chunk is always considered part of catalog
+            associated_content_keys.add(metadata_chunk.content_key)
+
+            if metadata_chunk.parent_content_key:
+                # parent content key of metadata_chunk is considered part of catalog
+                associated_content_keys.add(metadata_chunk.parent_content_key)
+
+            json_metadata = metadata_chunk.json_metadata
```
brittneyexline: i wonder how this will compare performance wise... perhaps we could get which course runs might apply by looking at which content metadata has the course key as their parent key?

adamstankiewicz: I updated this method so we no longer need to iterate through all of `self.content_metadata` to find all the content keys associated with the catalog. Rather, we can get all content metadata associated with the catalog that match the provided `content_keys` in a single query.

If that query returns at least one content metadata object, the `content_keys` are contained by the catalog.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -156,34 +157,44 @@ def content_metadata(self):
...
+        parent_content_keys = {
+            metadata.parent_content_key
+            for metadata in searched_metadata
+            if metadata.parent_content_key
+        }
+
+        # find all content metadata associated with the catalog that match specified content_keys
+        query = Q(content_key__in=content_keys)
+        query |= Q(parent_content_key__in=content_keys)
+        query |= Q(content_key__in=parent_content_keys)
```
bbaker6225: Could you add a comment about the difference between `parent_content_key__in=content_keys` vs. `content_key__in=parent_content_keys`? I think it might be helpful to have that distinction a bit more clear

enterprise_catalog/apps/catalog/models.py
```diff
@@ -156,34 +157,44 @@ def content_metadata(self):
...
+
+        content_metadata = self.catalog_query.contentmetadata_set.filter(query)
```
bbaker6225: Nit: You should be able to just do `self.content_metadata.filter(query)`

bbaker6225 (commented):
Looks good overall, although I'd definitely like to see some correctness testing on stage to verify

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: Cleaned up the code comments to be a bit more descriptive.

adamstankiewicz:
> Looks good overall, although I'd definitely like to see some correctness testing on stage to verify

@bbaker6225 Agreed. I plan on testing all cases outlined in the code comments on stage to verify these changes.

bbaker6225 (approved):
Nice

enterprise_catalog/apps/catalog/models.py
```diff
@@ -156,34 +157,50 @@ def content_metadata(self):
...
+            metadata.parent_content_key
+            for metadata in searched_metadata
+            if metadata.parent_content_key
+        }
+        query |= Q(content_key__in=parent_content_keys)
+
+        content_metadata = self.content_metadata.filter(query)
+
+        # if content metadata was returned, the specified content_keys exist in the catalog
+        return bool(content_metadata)
```
iloveagent57: This is just a side note and not a suggestion, since I don't think it makes any difference here (unless the `content_metadata` query set is very large): https://docs.djangoproject.com/en/2.2/ref/models/querysets/#django.db.models.query.QuerySet.exists

enterprise_catalog/settings/local.py
```diff
@@ -46,6 +46,18 @@
 INTERNAL_IPS = ('127.0.0.1',)
 # END TOOLBAR CONFIGURATION
 
+# DJANGO-SILK CONFIGURATION
+# See: https://github.com/jazzband/django-silk
+if os.environ.get('ENABLE_DJANGO_SILK', False):
```
iloveagent57, line 51: Nice

iloveagent57 (approved):

enterprise_catalog/apps/catalog/models.py
```diff
@@ -156,34 +157,50 @@ def content_metadata(self):
...
+        # parent content keys, i.e. course ids associated with the specified content_keys
+        # (if any) to handle the following case:
+        #   - catalog contains courses and the specified content_keys are course run ids.
+        searched_metadata = ContentMetadata.objects.filter(content_key__in=content_keys)
+        parent_content_keys = {
+            metadata.parent_content_key
+            for metadata in searched_metadata
+            if metadata.parent_content_key
+        }
+        query |= Q(content_key__in=parent_content_keys)
```
iloveagent57, line 198: One other thought: I wonder if it's worth enforcing some maximum length of `content_keys`, since we're now increasing the size of the query string by 1/3 (there's some limit on the length of the string you can send to MySQL, I think it's 64KB).

adamstankiewicz: Given that I haven't really seen any cases in our logging or code paths that send more than 1 `content_key` at a time, I believe we're safe for now but agreed we should definitely keep an eye out if it becomes an issue later.

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: Yeah, `content_metadata` here shouldn't be too large since we usually se at most  `content_key` at a time, and we'd only be returning `ContentMetadata` that matches that `content_key`.

But, I'm going to incorporate `.exists` since it feels more Django-y 😄

PR #127, opened 2020-06-04
Don't associate metadata when discovery call(s) fails during `update_content_metadata` job

adamstankiewicz:
## Description

Several Enterprise catalogs had 0 results on the `get_content_metadata` response but edx-enterprise was returning 31 results. I re-synced one of these catalogs and the update worked; `get_content_metadata` for that catalog now returns the expected 31 results.

Digging into splunk, this was caused when calls to Discovery fails during the `update_content_metadata` job, we were returning 0 results and continuing to associate 0 metadata results with the catalog query, overwriting what was already in there.

This PR proposes a solution to only associate content metadata when we get all results for a catalog query. When a call(s) to discovery fails for a catalog, it would need to be re-synced or wait until the next run of the the `update_content_metadata` job.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api_client/discovery.py
+++ b/enterprise_catalog/apps/api_client/discovery.py
@@ -45,7 +45,8 @@ class DiscoveryApiClient:
         query_params (dict): additional query params for the rest api endpoint
              we're hitting. e.g. - {'page': 3}
 
-        Returns a list of the results.
+        Returns a list of the results, or `None` if there was an error calling the
+        discovery service.
         """
         if query_params is None:
             query_params = {}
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -70,11 +71,15 @@ def get_metadata_by_query(self, catalog_query, query_params=None):
...
                 ).json()
                 results += response.get('results', [])
-        except JSONDecodeError:
+        except JSONDecodeError as exc:
             LOGGER.error(
-                'Could not retrieve content items from course-discovery (page %s) for catalog query %s',
+                'Could not retrieve content items from course-discovery (page %s) for catalog query %s: %s',
                 page,
                 catalog_query,
+                exc,
```
adamstankiewicz, line 79: Adding the exception to know what it was caused by (so we don't have to go digging in the discovery logs).

enterprise_catalog/apps/api_client/tests/test_discovery.py
```diff
@@ -29,3 +30,20 @@ def test_get_metadata_by_query(self, mock_oauth_client):
 
         expected_response = [{'key': 'fakeX'}]
         self.assertEqual(actual_response, expected_response)
+
+    @mock.patch('enterprise_catalog.apps.api_client.discovery.OAuthAPIClient')
+    def test_get_metadata_by_query_with_error(self, mock_oauth_client):
+        """
+        get_metadata_by_query should return None when a call(s) to discovery endpoint fail.
+        """
+        mock_oauth_client.return_value.post.side_effect = JSONDecodeError('error', '{}', 0)
```
adamstankiewicz, line 39: these arguments to `JSONDecodeError` are [required](https://github.com/simplejson/simplejson/blob/00ed20da4c0e5f0396661f73482418651ff4d8c7/simplejson/errors.py#L26).

chavesj (approved):
noice

PR #129, opened 2020-06-06
Index course data in enterprise_catalog Algolia index

adamstankiewicz:
## Description

This PR indexes course data in our own `enterprise_catalog` index within Algolia, a hosted a search provider. By indexing data using this approach, we don't need to rely on the implementation within course-discovery. One tradeoff with this approach, however, is that we need to calculate some Algolia fields ourselves, e.g. 'availability', 'language'. However, this means that we have full control over what gets included in our Algolia index.

A new management command `reindex_algolia` was created to spin off Celery tasks to traverse course-discovery's `/api/v1/courses/` endpoint for all content keys that are used by at least one EnterpriseCatalog. We add a few additionals fields (e.g., `enterprise_catalog_uuids`, `enterprise_customer_uuids`) to each course and then extract out all the necessary course fields for Algolia to search against, filter, and/or display in the UI.

---

When performing searches from the search page within the Learner Portal, we can use one of two approaches to filter content for a given Enterprise:
1. Use a "hidden" facet for `enterprise_catalog_uuids` to return only content that is contained by one or more specific EnterpriseCatalog(s);
2. Use a "hidden" facet for `enterprise_customer_uuids` to return only content that is contained by a specific EnterpriseCustomer.

---

How this looks in the Algolia UI:

image

## More information
https://docs.google.com/document/d/1MAsXS3QMpLaX49CrD3RNWvZ3-bSXGgJ6IgTHJou9Jc0/edit?usp=sharing

## Ticket Link

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -1,8 +1,90 @@
+import logging
+
 from celery import shared_task
 
+from enterprise_catalog.apps.api.v1.utils import (
+    add_uuids_to_courses,
+    create_algolia_objects_from_courses,
+    initialize_algolia_index,
+)
+from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.catalog.models import (
+    get_related_enterprise_catalogs_for_content_keys,
     update_contentmetadata_from_discovery,
 )
+
+
+logger = logging.getLogger(__name__)
+
+
+@shared_task(bind=True)
+# pylint: disable=unused-argument
+def index_enterprise_catalog_courses_in_algolia(self, content_keys, index_name, app_id, api_key, algolia_fields):
+    """
+    Index a batch of course content into Algolia.
+
+    Arguments:
+        content_keys (list): a list of content_keys to index in Algolia
+        index_name (str): name of an Algolia index
+        app_id (str): an APPLICATION_ID for Algolia
+        api_key (str): an API_KEY for Algolia
+    """
+    algolia_index = initialize_algolia_index(index_name, app_id, api_key)
+    if not algolia_index:
+        # without an Algolia index, we can't really do much of anything so exit early
+        return
+
+    discovery_client = DiscoveryApiClient()
+    query_params = {
+        'keys': ','.join(content_keys),
+        'limit': 100,
+        'ordering': 'key',
+    }
+    courses = discovery_client.get_courses(query_params=query_params)
+    logger.info(
+        'Retrieved %d courses from course-discovery for %d content keys: %s',
+        len(courses),
+        len(content_keys),
+        content_keys,
+    )
+
+    if courses is not None:
+        # Get related enterprise_catalog_uuids and enterprise_customer_uuids for the provided content_keys
+        related_enterprise_catalogs = get_related_enterprise_catalogs_for_content_keys(content_keys)
+        for content_key, uuids in related_enterprise_catalogs.items():
+            # add those enterprise-specific uuids to the appropriate course object within the list of courses
+            courses = add_uuids_to_courses(content_key, uuids, courses)
+
+        # create objects to send to Algolia by extracting only the fields we care about
+        algolia_objects = create_algolia_objects_from_courses(courses, algolia_fields)
+
+        try:
+            # Add algolia_objects to the Algolia index
+            response = algolia_index.partial_update_objects(algolia_objects, {
+                'createIfNotExists': True,
+            })
+            object_ids = []
+            for response in response.raw_responses:
+                object_ids += response.get('objectIDs', [])
+            logger.info(
+                'Successfully indexed %d (%d unique) courses in Algolia\'s %s index: %s',
+                len(object_ids),
+                len(set(object_ids)),
+                index_name,
+                object_ids,
+            )
+        except Exception as exc:  # pylint: disable=broad-except
+            logger.error(
+                'Could not index %d course(s) in Algolia\'s %s index: %s',
+                len(algolia_objects),
+                index_name,
+                exc,
+            )
+    else:
+        logger.error(
+            'No courses retrieved from course-discovery for content keys: %s',
+            content_keys,
+        )
 
 
 @shared_task(bind=True)
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -2,6 +2,7 @@ import copy
 import logging
 
 from celery import shared_task
+from celery_utils.logged_task import LoggedTask
 
 from enterprise_catalog.apps.api.v1.utils import (
     create_algolia_objects_from_courses,
@@ -19,7 +20,7 @@ from enterprise_catalog.apps.catalog.models import (
 logger = logging.getLogger(__name__)
 
 
-@shared_task
+@shared_task(base=LoggedTask)
 def index_enterprise_catalog_courses_in_algolia(content_keys, algolia_fields):
     """
     Index a batch of course content into Algolia.
@@ -83,6 +84,6 @@ def index_enterprise_catalog_courses_in_algolia(content_keys, algolia_fields):
     algolia_client.partially_update_index(algolia_objects)
 
 
-@shared_task
+@shared_task(base=LoggedTask)
 def update_catalog_metadata_task(catalog_query_id):
     update_contentmetadata_from_discovery(catalog_query_id)
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -1,12 +1,14 @@
+import copy
 import logging
 
 from celery import shared_task
 
 from enterprise_catalog.apps.api.v1.utils import (
-    add_uuids_to_courses,
     create_algolia_objects_from_courses,
-    initialize_algolia_index,
+    find_index_in_courses_for_content_key,
+    get_algolia_object_id,
 )
+from enterprise_catalog.apps.api_client.algolia import AlgoliaSearchClient
 from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.catalog.models import (
     get_related_enterprise_catalogs_for_content_keys,
@@ -17,30 +19,30 @@ from enterprise_catalog.apps.catalog.models import (
 logger = logging.getLogger(__name__)
 
 
-@shared_task(bind=True)
-# pylint: disable=unused-argument
-def index_enterprise_catalog_courses_in_algolia(self, content_keys, index_name, app_id, api_key, algolia_fields):
+@shared_task
+def index_enterprise_catalog_courses_in_algolia(content_keys, algolia_fields):
     """
     Index a batch of course content into Algolia.
 
     Arguments:
         content_keys (list): a list of content_keys to index in Algolia
-        index_name (str): name of an Algolia index
-        app_id (str): an APPLICATION_ID for Algolia
-        api_key (str): an API_KEY for Algolia
+        algolia_fields (list): list of course fields we want to index in Algolia
     """
-    algolia_index = initialize_algolia_index(index_name, app_id, api_key)
-    if not algolia_index:
-        # without an Algolia index, we can't really do much of anything so exit early
-        return
+    # initialize Algolia index
+    algolia_client = AlgoliaSearchClient()
+    algolia_client.init_index()
 
-    discovery_client = DiscoveryApiClient()
     query_params = {
         'keys': ','.join(content_keys),
         'limit': 100,
         'ordering': 'key',
     }
-    courses = discovery_client.get_courses(query_params=query_params)
+    try:
+        discovery_client = DiscoveryApiClient()
+        courses = discovery_client.get_courses(query_params=query_params)
+    except Exception:  # pylint: disable=broad-except
+        courses = None
+
     logger.info(
         'Retrieved %d courses from course-discovery for %d content keys: %s',
         len(courses),
@@ -48,46 +50,39 @@ def index_enterprise_catalog_courses_in_algolia(self, content_keys, index_name, 
         content_keys,
     )
 
-    if courses is not None:
-        # Get related enterprise_catalog_uuids and enterprise_customer_uuids for the provided content_keys
-        related_enterprise_catalogs = get_related_enterprise_catalogs_for_content_keys(content_keys)
-        for content_key, uuids in related_enterprise_catalogs.items():
-            # add those enterprise-specific uuids to the appropriate course object within the list of courses
-            courses = add_uuids_to_courses(content_key, uuids, courses)
-
-        # create objects to send to Algolia by extracting only the fields we care about
-        algolia_objects = create_algolia_objects_from_courses(courses, algolia_fields)
-
-        try:
-            # Add algolia_objects to the Algolia index
-            response = algolia_index.partial_update_objects(algolia_objects, {
-                'createIfNotExists': True,
-            })
-            object_ids = []
-            for response in response.raw_responses:
-                object_ids += response.get('objectIDs', [])
-            logger.info(
-                'Successfully indexed %d (%d unique) courses in Algolia\'s %s index: %s',
-                len(object_ids),
-                len(set(object_ids)),
-                index_name,
-                object_ids,
-            )
-        except Exception as exc:  # pylint: disable=broad-except
-            logger.error(
-                'Could not index %d course(s) in Algolia\'s %s index: %s',
-                len(algolia_objects),
-                index_name,
-                exc,
-            )
-    else:
+    if not courses:
         logger.error(
             'No courses retrieved from course-discovery for content keys: %s',
             content_keys,
         )
+        return
+
+    # Get related enterprise_catalog_uuids and enterprise_customer_uuids for the provided content_keys
+    related_enterprise_catalogs = get_related_enterprise_catalogs_for_content_keys(content_keys)
+
+    for content_key, uuids in related_enterprise_catalogs.items():
+        # add those enterprise-specific uuids to the appropriate course object within the list of courses
+        course_index = find_index_in_courses_for_content_key(content_key, courses)
+        if course_index is not None:
+            course = copy.deepcopy(courses[course_index])
+            course.update({
+                'objectID': get_algolia_object_id(course['uuid']),
+                'enterprise_catalog_uuids': uuids.get('enterprise_catalog_uuids', []),
+                'enterprise_customer_uuids': uuids.get('enterprise_customer_uuids', []),
+            })
+            courses[course_index] = course
+        else:
+            logger.error(
+                'Could not find content_key "%s" in list of %d courses.',
+                content_key,
+                len(content_keys),
+            )
+
+    # extract out only the fields we care about and send to Algolia index
+    algolia_objects = create_algolia_objects_from_courses(courses, algolia_fields)
+    algolia_client.partial_update_algolia_index(algolia_objects)
 
 
-@shared_task(bind=True)
-# pylint: disable=unused-argument
-def update_catalog_metadata_task(self, catalog_query_id):
+@shared_task
+def update_catalog_metadata_task(catalog_query_id):
     update_contentmetadata_from_discovery(catalog_query_id)
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -80,7 +80,7 @@ def index_enterprise_catalog_courses_in_algolia(content_keys, algolia_fields):
 
     # extract out only the fields we care about and send to Algolia index
     algolia_objects = create_algolia_objects_from_courses(courses, algolia_fields)
-    algolia_client.partial_update_algolia_index(algolia_objects)
+    algolia_client.partially_update_index(algolia_objects)
 
 
 @shared_task
--- a/enterprise_catalog/apps/api/v1/utils.py
+++ b/enterprise_catalog/apps/api/v1/utils.py
@@ -134,7 +134,7 @@ def get_course_language(course_runs):
 
 def get_course_availability(course_runs):
     """
-    Gets the availability for a course.
+    Gets the availability for a course. Used for adding the "Availability" facet for courses in Algolia.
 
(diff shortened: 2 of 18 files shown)


enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -71,7 +70,7 @@ def get_metadata_by_query(self, catalog_query, query_params=None):
                     params=query_params
                 ).json()
                 results += response.get('results', [])
-        except JSONDecodeError as exc:
+        except Exception as exc:
```
adamstankiewicz, line 73: broadening the exception, just in case...

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -0,0 +1,164 @@
...
+                exc,
+            )
+
+    def handle(self, *args, **options):
+        """
+        Spin off tasks to fetch courses from the discovery service and index them in Algolia.
+        """
+        # find all ContentMetadata records with a content type of "course" that are
+        # also part of at least one EnterpriseCatalog
+        content_metadata = ContentMetadata.objects.filter(
```
adamstankiewicz, line 111: This will return, at most, ~4300 results on production.

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -0,0 +1,164 @@
...
+        content_keys = [metadata.content_key for metadata in content_metadata]
+
+        # set default settings to use for the index. note: this will override manual updates
+        # to the index configuration on the Algolia dashboard but ensures consistent settings.
+        self.set_algolia_index_settings()
+
+        # break up the content_keys in smaller batches, where each batch will spin off its
+        # own celery task. this should help performance and prevent errors with having too
+        # many content_keys for a GET request to the discovery service's /courses endpoint
+        for keys in self.batch(content_keys, batch_size=250):
```
adamstankiewicz, line 130: Given that there are ~4300 course records within enterprise-catalog, this will generate ~18 tasks at this batch size of 250 keys per batch. This means that in each task, we're passing 250 keys at once (comma-separated) in the `keys` query parameter to the `/api/v1/courses/` API endpoint. We can probably adjust `batch_size` to optimize performance.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -292,15 +304,49 @@ def __str__(self):
...
+def get_related_enterprise_catalogs_for_content_keys(content_keys):
+    """
+    Get enterprise_catalog_uuids and enterprise_customer_uuids for each of the specified content_keys.
+    """
+    related_catalogs_for_keys = {}
+
+    content_metadata = ContentMetadata.objects.filter(content_key__in=content_keys)
+
+    catalog_queries = CatalogQuery.objects.prefetch_related('contentmetadata_set')
+    catalog_queries = catalog_queries.prefetch_related('enterprise_catalogs')
```
adamstankiewicz: By querying for enterprise catalogs through the `CatalogQuery` model, we can get everything we need in a fixed number of queries. Since we use `enterprise_catalogs` and `contentmetadat_set` below, we can optimize by pre-fetching the data for those fields.

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -0,0 +1,164 @@
...
+    'enterprise_customer_uuids',
+    'entitlements',
+    'expected_learning_items',
+    'extra_description',
+    'faq',
+    'full_description',
+    'key',  # for links to course about pages from the Learner Portal search page
+    'language',
+    'level_type',
+    'objectID',  # required by Algolia, e.g. "course-{uuid}"
```
rgraber, line 38: if you're not including programs you don't need to prefix the uuid with 'course'

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,178 @@ def is_any_course_run_enrollable(course_runs):
...
+        content_language = course_run.get('content_language')
+        if not content_language:
+            continue
+        language_name = Language.make(language=content_language).language_name()
+        languages.add(language_name)
+
+    return list(languages)
+
+
+def get_course_availability(course_runs):
```
rgraber, line 135: i think you accounted for this, but just a note that this is different logic than what's displayed on the B2C site, so you might get some inconsistencies

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
adamstankiewicz: We won't be initially supporting programs for MVP but it is likely something we'll be adding on later, so I think it makes sense to keep the prefix for the time being.

enterprise_catalog/apps/api/v1/utils.py
adamstankiewicz: Yeah, for this initial prototype, I just got the barebones functionality in place. We can adjust in an incremental approach to make these sorts of fields exactly consistent with the B2C experience if necessary.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -1,10 +1,92 @@
...
+    get_related_enterprise_catalogs_for_content_keys,
     update_contentmetadata_from_discovery,
 )
 
 
+logger = logging.getLogger(__name__)
+
+
+@shared_task(bind=True)
+# pylint: disable=unused-argument
```
iloveagent57: If you don't need `self` as a parameter, I _think_ you can get rid of `bind=True` above and get rid of the self parameter. Then you don't need the pylint disable.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -1,10 +1,92 @@
...
+    algolia_index = initialize_algolia_index(index_name, app_id, api_key)
+    if not algolia_index:
+        # without an Algolia index, we can't really do much of anything so exit early
+        return
+
+    discovery_client = DiscoveryApiClient()
+    query_params = {
+        'keys': ','.join(content_keys),
+        'limit': 100,
+        'ordering': 'key',
```
iloveagent57, line 39: Do we want to order by things in addition to `key`, like `start`?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -1,10 +1,92 @@
...
+    }
+    courses = discovery_client.get_courses(query_params=query_params)
+    logger.info(
+        'Retrieved %d courses from course-discovery for %d content keys: %s',
+        len(courses),
+        len(content_keys),
+        content_keys,
+    )
+
+    if courses is not None:
```
iloveagent57: Nit: can we just check `if courses` for truthiness/emptiness here?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -1,10 +1,92 @@
...
+        # Get related enterprise_catalog_uuids and enterprise_customer_uuids for the provided content_keys
+        related_enterprise_catalogs = get_related_enterprise_catalogs_for_content_keys(content_keys)
+        for content_key, uuids in related_enterprise_catalogs.items():
+            # add those enterprise-specific uuids to the appropriate course object within the list of courses
+            courses = add_uuids_to_courses(content_key, uuids, courses)
```
iloveagent57: It feels weird to pass all of `courses` into this, and then re-assign the result to `courses`.

enterprise_catalog/apps/api/tasks.py
iloveagent57: Nit: to lessen indentation, let's check for falsiness and early-return?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -1,10 +1,92 @@
...
+# pylint: disable=unused-argument
+def index_enterprise_catalog_courses_in_algolia(self, content_keys, index_name, app_id, api_key, algolia_fields):
+    """
+    Index a batch of course content into Algolia.
+
+    Arguments:
+        content_keys (list): a list of content_keys to index in Algolia
+        index_name (str): name of an Algolia index
+        app_id (str): an APPLICATION_ID for Algolia
+        api_key (str): an API_KEY for Algolia
```
iloveagent57: Nit: add line for `algolia_fields`.

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,178 @@ def is_any_course_run_enrollable(course_runs):
...
+        dict: an Algolia index or None if we cannot initialize an Algolia index
+    """
+    if not index_name:
+        logger.error('Could not initialize Algolia index due to missing index name.')
+        # can't do much without an index_name, so return None
+        return None
+
+    algolia_client = None
+    if app_id and api_key:
+        algolia_client = SearchClient.create(app_id, api_key)
```
iloveagent57: It's going to be worth adding a little `aloglia_client()` helper function composed of these two lines.

iloveagent57: ...and you'll probably want to read `app_id` and `api_key` from settings.  Which means your management command function doesn't have to worry about those params.

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,178 @@ def is_any_course_run_enrollable(course_runs):
...
+    Returns:
+        list: a list of courses with the uuids added on to the appropriate course
+    """
+    # find index in courses where course key matches content_key
+    course_index = next(
+        (index for (index, d) in enumerate(courses) if d['key'] == content_key),
+        None
+    )
+    if course_index is not None:
+        course = courses[course_index].copy()
```
iloveagent57: Nit: you _may_ want to use deepcopy instead.

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,178 @@ def is_any_course_run_enrollable(course_runs):
...
+    """
+    Gets the availability for a course.
+
+    Arguments:
+        course_runs (list): list of course runs for a course
+
+    Returns:
+        list: a list of availabilities for those course runs (e.g., "Upcoming")
+    """
+    availability = set()
```
iloveagent57, line 151: How are you tying this back to the course runs?  Do you really want a dict?

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,178 @@ def is_any_course_run_enrollable(course_runs):
...
+
+    for course_run in course_runs:
+        run_availability = course_run.get('availability', '').lower()
+        if run_availability == 'current':
+            availability.add('Available now')
+        elif run_availability == 'upcoming':
+            availability.add('Upcoming')
+        else:
+            availability.add('Archived')
```
iloveagent57: Nit: you can make this code more "data-driven" and clean up these magic strings with something like:
```python
DEFAULT_COURSE_AVAILABILITY = 'Archived'
COURSE_AVAILABILITY_MESSAGES = {
    'current': 'Available Now',
    'upcoming': 'Upcoming',
}
...
for course_run in course_runs:
    run_availability = course_run.get('availability', '').lower()
    availability.add(COURSE_AVAILABILITY_MESSAGES.get(availability, DEFAULT_COURSE_AVAILABILITY))
```

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,178 @@ def is_any_course_run_enrollable(course_runs):
...
+            'enterprise_customer_uuids': uuids.get('enterprise_customer_uuids', []),
+        })
+        courses[course_index] = course
+    else:
+        logger.error(
+            'Could not find course with content key %s from course-discovery discovery',
+            content_key,
+        )
+
+    return courses
```
iloveagent57: I don't have a concrete recommendation right now, but the fact that you're
1) plucking an index out of a list, and
2) modifying a data structure that's passed into a function
...is a bit of a code smell :). Happy to talk through that more tomorrow.

iloveagent57 (commented):
Gotta leave the computer for a little bit, but here's a few comments to start.

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: Yup, looks like it according to the docs!

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: I don't see `start` as an ordering option (or anything related to dates) in this list  for the /api/v1/courses endpoint.

image

enterprise_catalog/apps/api/v1/utils.py
adamstankiewicz: This is a field that gets faceted within Algolia, so it should be a list. The list is added as key to the course object. In the return, I am casting it as a list, though it is initialized as a set in order to prevent duplicates when iterating through the course runs.

adamstankiewicz:
@iloveagent57 Feel free to take another look. Addressed your feedback and added a new `AlgoliaSearchClient` class.

enterprise_catalog/apps/api/tasks.py
iloveagent57: Sorry, I was thinking of course runs, not courses.

enterprise_catalog/apps/api/v1/utils.py
iloveagent57: 👍 Gotchya.  Could you add a line to the docstring like "Useful for determining the `availability` facet of a course object in Algolia." ?

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,134 @@ def is_any_course_run_enrollable(course_runs):
...
+    Transforms a course into an Algolia object.
+
+    Arguments:
+        course (dict): a course dict
+        algolia_fields (list): list of fields to extract from the course
+
+    Returns:
+        dict: a dictionary containing only the fields noted in algolia_fields
+    """
+    course = copy.deepcopy(course)
```
iloveagent57: Nit: parameter/variable reassignment - maybe `searchable_course`?

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,134 @@ def is_any_course_run_enrollable(course_runs):
...
+    published_course_runs = list(
+        filter(lambda d: d['status'].lower() == 'published', course.get('course_runs', []))
```
iloveagent57: Nit: you can filter using a list comprehension like so
```python
published_course_runs = [
    course_run for course_run in searchable_course.get('course_runs', [])
    if course_run['status'].lower() == 'published'
]
```

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,134 @@ def is_any_course_run_enrollable(course_runs):
...
+    for course_run in course_runs:
+        run_availability = course_run.get('availability', '').lower()
+        availability.add(
+            COURSE_AVAILABILITY_MESSAGES.get(run_availability, DEFAULT_COURSE_AVAILABILITY)
+        )
+
+    return list(availability)
+
+
+def get_algolia_object_from_course(course, algolia_fields):
```
iloveagent57: Nit: make this private with a leading `_`, since (I think) it's only used by the `create_algolia_objects_from_courses` function below.

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -68,3 +75,134 @@ def is_any_course_run_enrollable(course_runs):
...
+        courses (list): list of courses
+        algolia_fields (list): list of fields to extract from courses
+
+    Returns:
+        list: a list of Algolia objects containing only the fields noted in algolia_fields
+    """
+    if not algolia_fields:
+        algolia_fields = []
+
+    algolia_objects = list()
```
iloveagent57: Python convention nit: `algolia_objects = []`

iloveagent57: Actually, this could be a single expression:
```python
algolia_objects = [
    get_algolia_object_from_course(course, algolia_fields)
    for course in courses
]
```

enterprise_catalog/apps/api/v1/utils.py
iloveagent57: Even tinier nit, bordering on pet-peeve: `get` is superfluous in this method name...you're not setting algolia objects from a course anywhere ")

enterprise_catalog/apps/api/v1/utils.py
adamstankiewicz: > Nit: parameter/variable reassignment - maybe `searchable_course`?

D'oh, good catch!

enterprise_catalog/apps/api/v1/utils.py
adamstankiewicz: Much better, not sure why I made it more complex 👍

enterprise_catalog/apps/api_client/algolia.py
```diff
@@ -0,0 +1,118 @@
...
+    """
+    Object builds an API client to make calls to an Algolia index.
+    """
+
+    ALGOLIA_APPLICATION_ID = settings.ALGOLIA.get('APPLICATION_ID')
+    ALGOLIA_API_KEY = settings.ALGOLIA.get('API_KEY')
+    ALGOLIA_INDEX_NAME = settings.ALGOLIA.get('INDEX_NAME')
+
+    def __init__(self):
+        self.client = None
```
iloveagent57: Nit: make this private, since the whole point of `AlgoliaSearchClient` is to provide an abstraction over the third-party `algoliasearch.SearchClient`.

enterprise_catalog/apps/api_client/algolia.py
```diff
@@ -0,0 +1,118 @@
...
+        try:
+            self.algolia_index.set_settings(index_settings)
+        except Exception as exc:  # pylint: disable=broad-except
+            logger.error(
+                'Unable to set settings for Algolia\'s %s index: %s',
+                self.ALGOLIA_INDEX_NAME,
+                exc,
+            )
+
+    def partial_update_algolia_index(self, algolia_objects):
```
iloveagent57: `partially_update_index` maybe?

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -83,3 +85,53 @@ def get_metadata_by_query(self, catalog_query, query_params=None):
...
+
+        Returns:
+            list: a list of the results, or None if there was an error calling the discovery service.
+        """
+        if query_params is None:
+            query_params = {}
+
+        query_params.update({
+            **query_params,
+        })
```
iloveagent57: This is kinda confusing.  How about
```python
request_params = {}
request_params.update(query_params or {})
...
response = self.client.get(self.COURSES_ENDPOINT, params=request_params).json()
```

iloveagent57 (commented):
New batch of suggestions, not quite to the bottom yet :)

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -83,3 +85,53 @@ def get_metadata_by_query(self, catalog_query, query_params=None):
...
+
+            # Traverse all results and concatenate results together
+            while response.get('next'):
+                offset += OFFSET_SIZE
+                query_params.update({'offset': offset})
+                response = self.client.get(
+                    self.COURSES_ENDPOINT,
+                    params=query_params,
+                ).json()
+                results += response.get('results', [])
```
iloveagent57: I wonder if you want to add a `traverse_pagination` optional parameter like we do in other places?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -292,15 +304,49 @@ def __str__(self):
...
 
+def get_related_enterprise_catalogs_for_content_keys(content_keys):
+    """
+    Get enterprise_catalog_uuids and enterprise_customer_uuids for each of the specified content_keys.
+    """
+    related_catalogs_for_keys = {}
+
+    content_metadata = ContentMetadata.objects.filter(content_key__in=content_keys)
+
+    catalog_queries = CatalogQuery.objects.prefetch_related('contentmetadata_set')
```
iloveagent57: I think you can pass multiple arguments to `prefetch_related()`, i.e. `prefetch_related('contentmetadata_set', 'enterprise_catalogs')`

enterprise_catalog/apps/api/tasks.py
```diff
@@ -1,11 +1,88 @@
...
 
 
-@shared_task(bind=True)
-# pylint: disable=unused-argument
-def update_catalog_metadata_task(self, catalog_query_id):
+logger = logging.getLogger(__name__)
+
+
+@shared_task
+def index_enterprise_catalog_courses_in_algolia(content_keys, algolia_fields):
```
iloveagent57: You could make this a `LoggedTask` if you install edx-celeryutils: https://github.com/edx/edx-celeryutils/blob/master/celery_utils/logged_task.py

iloveagent57 (commented):
Last batch of comments, looks great.

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: Oooh, I will look that up! Good to know if that is the case. Thanks!

enterprise_catalog/apps/api_client/discovery.py
adamstankiewicz: I held off on adding `traverse_pagination` here given that we don't have it (yet) for the calls to /search/all in `get_metadata_by_query` either.

Given that the callers of both `get_courses` and `get_metadata_by_query` always want to traverse the paginated response, I don't know if we need to support `traverse_pagination`quite yet.

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: Looks like it! Will update.

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: Looks worthwhile to include! I will update both of our async tasks to use `LoggedTask`.

iloveagent57 (approved):

PR #131, opened 2020-06-11
Refactor update_content_metadata and reindex_algolia

adamstankiewicz:
## Description

Refactors `update_content_metadata` command to run a group of tasks in parallel to perform the ContentMetadata <> CatalogQuery associations from /search/all data. Then, once all tasks finish, a callback to another task is performed. 

This follow-up task then iterates through all /api/v1/courses, finding only the courses that match existing ContentMetadata records. Each record's json_metadata is updated to use full course metadata.

This means ContentMetadata records will always have the full course metadata 🎉 

Also introduced in this PR is a refactor of `reindex_algolia` command which now indexes all ContentMetadata records of type "course" into Algolia, adding on enterprise-related fields.

## Ticket Link

## Post-review


enterprise_catalog/apps/api/tasks.py
```diff
@@ -21,62 +22,123 @@
 
 
 @shared_task(base=LoggedTask)
-def index_enterprise_catalog_courses_in_algolia(content_keys, algolia_fields):
+def update_full_content_metadata_task(*args, **kwargs):
```
adamstankiewicz: TODO: docstring

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -91,37 +92,42 @@ def get_courses(self, query_params=None):
...
-                query_params.update({'offset': offset})
-                response = self.client.get(self.COURSES_ENDPOINT, params=request_params).json()
-                results += response.get('results', [])
+                request_params.update({'offset': offset})
+                LOGGER.info('Retrieving courses from course-discovery for offset {}...'.format(offset))
+                response = self.client.get(
+                    self.COURSES_ENDPOINT,
+                    params=request_params,
+                ).json()
+                courses += response.get('results', [])
```
chavesj: here's another block of repeated code that can be broken out into a utility method (see lines 107-112)

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -78,62 +67,19 @@
...
-            logger.info(message, len(content_keys), ALGOLIA_INDEX_NAME, keys)
+        index_enterprise_catalog_courses_in_algolia_task.delay(
+            algolia_fields=ALGOLIA_FIELDS,
+            algolia_settings=ALGOLIA_INDEX_SETTINGS,
+        )
+        message = (
+            'Spinning off task index_enterprise_catalog_courses_in_algolia_task from'
+            ' the reindex_algolia command to reindex course data in Algolia.'
+        )
+        logger.info(message)
```
chavesj: some serious cleanup here 🧽  🙂

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -78,62 +67,19 @@
...
-            index_enterprise_catalog_courses_in_algolia.delay(
-                content_keys=keys,
-                algolia_fields=ALGOLIA_FIELDS,
-            )
-            message = (
-                'Spinning off index_enterprise_catalog_courses_in_algolia from reindex_algolia command'
-                ' to add %d courses to Algolia\'s %s index: %s'
-            )
-            logger.info(message, len(content_keys), ALGOLIA_INDEX_NAME, keys)
+        index_enterprise_catalog_courses_in_algolia_task.delay(
```
chavesj: (This comment applies to every call to delay()) - delay() returns an AsyncResult object which contains the task ID. It's probably worth logging the task ID in your message below

enterprise_catalog/apps/api/tasks.py
```diff
@@ -21,62 +22,123 @@
...
-    if not courses:
-        logger.error(
-            'No courses retrieved from course-discovery for content keys: %s',
-            content_keys,
+    if not content_metadata:
+        message = (
+            'There are no ContentMetadata records of content type "%s" that are'
+            ' part of at least one EnterpriseCatalog.'
         )
+        logger.error(message, COURSE)
```
chavesj: this can be broken out into a method since it's used both here and in update_full_content_metadata_task

enterprise_catalog/settings/base.py
```diff
@@ -270,6 +270,7 @@
 CELERY_MESSAGE_COMPRESSION = 'gzip'
 
 # Results configuration
+CELERY_RESULT_BACKEND = os.environ.get('CELERY_RESULT_BACKEND', '')
```
bbaker6225: Is this set in edx-internal already?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -336,6 +338,25 @@ def get_related_enterprise_catalogs_for_content_keys(content_keys):
     return related_catalogs_for_keys
 
 
+def course_metadata_used_by_at_least_one_catalog():
+    # find all ContentMetadata records with a content type of "course" that are
+    # also part of at least one EnterpriseCatalog
```
bbaker6225: Nit: Might be worth turning this comment into a docstring for the overall method?

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
```diff
@@ -17,13 +20,48 @@ class Command(BaseCommand):
...
+        # find all CatalogQuery records used by at least one EnterpriseCatalog to avoid
         # calling /search/all/ for a CatalogQuery that is not currently used by any catalogs.
-        for catalog_query in CatalogQuery.objects.filter(enterprise_catalogs__isnull=False).distinct():
-            update_catalog_metadata_task.delay(catalog_query_id=catalog_query.id)
+        catalog_queries = CatalogQuery.objects.filter(enterprise_catalogs__isnull=False).distinct()
+
+        # create a group of celery tasks that run in parallel to create/update ContentMetadata records
+        # and associate those with the appropriate CatalogQuery(s). once all those tasks succeed, run a
+        # callback to update the json_metadata of ContentMetadata records with content type "course"
+        # with the full course metadata from /courses/.
```
bbaker6225, line 47: I imagine it's fine, but just want to check and see if you've run into any issue with potentially updating the same `ContentMetadata` at the same time since different queries can reference the same metadata

bbaker6225 (commented):
Looks good overall, just a few small comments

chavesj (approved):
LGTM 🚀

enterprise_catalog/settings/base.py
chavesj: Good catch @bbaker6225 , need a PR in edx-internal before this can be merged

chavesj (changes requested):

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
adamstankiewicz: Yeah, I haven't come across this issue in testing locally, but it also crossed my mind. It shouldn't have any impact on the association logic since that's still unique for each `CatalogQuery`. The create/update of `ContentMetadata` records should also be fine. If one task creates a new `ContentMetadata` record, another task would then just be updating the newly created one.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -21,62 +22,106 @@
...
+    updated_metadata_count = 0
+    for course_metadata in courses_in_content_metadata:
+        content_key = course_metadata.get('key')
+        try:
+            metadata_record = ContentMetadata.objects.get(content_key=content_key)
+        except ContentMetadata.DoesNotExist:
+            logger.error('Could not find ContentMetadata record for content_key %s.', content_key)
+            continue
+
+        metadata_record.json_metadata = course_metadata
```
bbaker6225: Are there any fields that we might be losing here that we want? To name one, I think `aggregation_key` is only on the search endpoints and I believe we use that

enterprise_catalog/settings/base.py
adamstankiewicz: @bbaker6225 Ah, right. Any guidance on what the value of this would be for stage/prod?

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: Hmm, within enterprise-catalog, we only use `aggregation_key` for the /search/all endpoint and parsing its results. Nothing within enterprise-catalog relies on that field existing within `json_metadata` on ContentMetadata records.

That said, I'm unsure off the top of my head about the downstream impacts of no longer including `aggregation_key`. If it's no longer included, callers of the `get_content_metadata` endpoint will no longer `aggregation_key` (or any other search-only fields).

I will look into this more, and follow-up.

enterprise_catalog/settings/base.py
bbaker6225: I'm seeing `'redis://edx-prod-queues.6sxrym.ng.0001.use1.cache.amazonaws.com:6379'` as the value elsewhere (such as for VEM). I'm pretty sure that's the shared one that we want to use, but it's probably worth throwing a question in the sre channel to check

chavesj: another option: since the result backend will be the same (for the foreseeable future) as the broker URL you can just move this statement below the BROKER_URL string formatting statement and set it equal to BROKER_URL. Wouldn't need an edx-internal PR that way

enterprise_catalog/apps/api/tasks.py
brittneyexline: i would grep for aggregation key in the edx-enterprise to see if anything there depends on it, since that is the main internal production code that uses get_content_metadata. otherwise i don't see it being a big deal. alternatively, you could merge the dictionaries and write that value instead of overwriting entirely.

enterprise_catalog/settings/base.py
adamstankiewicz: Updated!

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: The only use of `aggregation_key` within edx-enterprise seems to be in tests, e.g. creating fake API responses from the catalog details endpoint (or `get_content_metadata` equivalent in enterprise-catalog).

image

From this, I don't believe edx-enterprise necessarily depends on it. That said, the transmit_content_metadata job currently sends `aggregation_key` to our integrated channels (e.g., Degreed). That job depends on `get_content_metadata` and the data within `json_metadata`. 

Given that we are sending this data to 3rd party systems, it probably makes sense to merge the dictionaries in case any of our customers rely on `aggregation_key`.

bbaker6225: Yeah I agree that we might as well merge dictionaries

chavesj (approved):
noice, celery settings looking much cleaner 🧽

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: Updated! It now merges original json_metadata with new course_metadata to not remove any potentially critical keys.

adamstankiewicz:
@bbaker6225 @chavesj tagged you both on an edx-internal PR (https://github.com/edx/edx-internal/pull/2136) to add a non-recurring Jenkins job for the new management command `update_full_content_metadata` in case we ever need to run it.

PR #135, opened 2020-06-15
Perform a bulk update to reduce DB calls and simplify a query to prevent an error

adamstankiewicz:
## Description

* Perform one bulk update rather than thousands of individual updates when updating the `json_metadata` field with full course metadata.
* Simplify a queryset due to a "hash join" error by removing the pre-optimization of filtering for only `ContentMetadata` records used by at least one catalog as it's not really that crucial of a filter anyways.
  * The "hash join" error happened in the reindex_algolia management command, likely due to having 4 tables joined together in one query, one of which having a sizable amount of data for that many joins.

## Post-review


chavesj (approved):
LGTM

PR #136, opened 2020-06-15
add batch_size kwarg to bulk update of json_metadata

adamstankiewicz:
## Description

In a previous PR, I switched from doing individual updates thousands of time to a single bulk_update. However, I believe this is causing MySQL to timeout and throw this error:

image

In this PR, I am adding `batch_size` to bulk_update to only update 100 objects at a time. Not sure what the "correct" batch size should be here, but 100 feels reasonable.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/utils.py
+++ b/enterprise_catalog/apps/api/v1/utils.py
@@ -155,7 +155,7 @@ def _algolia_object_from_course(course, algolia_fields):
     searchable_course = copy.deepcopy(course)
     published_course_runs = [
         course_run for course_run in searchable_course.get('course_runs', [])
-        if course_run['status'].lower() == 'published'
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -155,7 +155,7 @@ def _algolia_object_from_course(course, algolia_fields):
     searchable_course = copy.deepcopy(course)
     published_course_runs = [
         course_run for course_run in searchable_course.get('course_runs', [])
-        if course_run['status'].lower() == 'published'
+        if course_run.get(['status'], '').lower() == 'published'
```
adamstankiewicz, line 158: prevents potential `KeyError` if/when `status` is not an attribute of a course run.

brittneyexline (approved):

iloveagent57 (approved):

PR #140, opened 2020-06-15
Fix syntax error with course_run.get()

adamstankiewicz:
## Description

Fixing `course_run.get(['status'], '')` to be `course_run.get('status', '')`

## Post-review


iloveagent57 (approved):
🤦

PR #142, opened 2020-06-16
clean up reindex-algolia job to spin off multiple tasks instead of just one

adamstankiewicz:
## Description

Before, the `reindex-algolia` job before was only spinning off a single task to do all the work of reindexing data into Algolia. However, this caused the task to linger for 21 minutes and spike the memory usage / CPU of the worker app instances significantly (up to nearly 5GB on the memory 😬).

Instead, I am now batching the `content_keys` of ContentMetadata records to reindex courses in chunks. I replicated all of the ContentMetadata records on stage in my local DB so that I could run the management command "at scale" locally. 

At a batch_size of 250 content_keys, the management command spins off several tasks, completing each in 15-35 seconds for a total of ~4 minutes to reindex all the courses in Algolia.

I'm also trimming down the amount of data for each course object within Algolia since while iterating through my copy of stage data, a few courses failed to be indexed due to being over the max allowable payload size from Algolia (i.e., max 10kb per object).

image

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -13,7 +13,6 @@ from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.catalog.models import (
     ContentMetadata,
     content_metadata_with_type_course,
-    related_enterprise_catalogs_for_content_metadata,
     update_contentmetadata_from_discovery,
 )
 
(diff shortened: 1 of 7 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -84,45 +83,48 @@ def update_full_content_metadata_task(*args, **kwargs):  # pylint: disable=unuse
...
-    content_metadata = content_metadata_with_type_course()
-
-    # find related enterprise_catalog_uuids and enterprise_customer_uuids for each ContentMetadata record
-    related_enterprise_catalogs = related_enterprise_catalogs_for_content_metadata(content_metadata)
+    # retrieve ContentMetadata records in bulk
+    content_metadata_records = ContentMetadata.objects.in_bulk(
+        content_keys,
+        field_name='content_key'
+    ).values()
+    content_metadata = list(content_metadata_records)
```
bbaker6225: Nit: The naming convention with `record` here seems a little strange. Maybe just wrap the bulk retrieve in `list`?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -305,39 +305,6 @@ def __str__(self):
         )
 
 
-def related_enterprise_catalogs_for_content_metadata(content_metadata):
```
bbaker6225, line 308: 🙏

bbaker6225 (approved):

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: Agreed :)

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -141,6 +141,72 @@ def get_course_availability(course_runs):
...
+
+    for program in programs:
+        program_type = program.get('type')
+        if program_type:
+            program_types.add(program_type)
+
+    return list(program_types)
+
+
+def get_course_subjects(course):
```
iloveagent57, line 188: Future small refactoring idea: these three functions do the same thing, the only difference is around the names of keys.

iloveagent57 (approved):

PR #143, opened 2020-06-16
Consider catalog and customer associations for catalogs that only contain course runs in Algolia indexing

adamstankiewicz:
## Description

Before, the `reindex-algolia` job was only retrieving the associated enterprise_catalog_uuids and enterprise_customer_uuids for catalogs/customers that explicitly have courses in their catalog(s). This means we were missing all the associations where a course's course run may be in an enterprise's catalog(s), but not the top-level course.

This PR ensures that for any given course, we account for catalogs/customers that have any of that course's course runs as part of its catalog as well before sending data to Algolia.

This now follows the [same logic as the `contains_content_items` endpoints](https://github.com/edx/enterprise-catalog/blob/9daaac7cf098c3353d0426a9c8ead0a89a851523/enterprise_catalog/apps/catalog/models.py#L196) which consider a course to be part of a catalog that contains only course runs if the course key matches any of the `parent_content_keys` of the course runs in the catalog.

**Example:**
Enterprise A has a single course in its catalog. Enterprise B has a single course run in its catalog, related to the course in Enterprise A's catalog. 

This means that for the course object we send to Algolia, it should contain both Enterprise A's and Enterprise B's catalog/customer uuids, not just Enterprise A's (like it was doing before).

## Ticket Link

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -10,6 +10,7 @@ from enterprise_catalog.apps.api.v1.utils import (
 )
 from enterprise_catalog.apps.api_client.algolia import AlgoliaSearchClient
 from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
+from enterprise_catalog.apps.catalog.constants import COURSE
 from enterprise_catalog.apps.catalog.models import (
     ContentMetadata,
     content_metadata_with_type_course,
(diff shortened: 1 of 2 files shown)


chavesj (approved):
The only way around is through 👍

enterprise_catalog/apps/api/tasks.py
```diff
@@ -99,31 +101,58 @@ def index_enterprise_catalog_courses_in_algolia_task(algolia_fields, content_key
...
-        content_keys,
-        field_name='content_key'
-    ).values())
-
     courses = []
-    # iterate through ContentMetadata records, retrieving the enterprise catalog uuids and
-    # enterprise customer uuids associated with each ContentMetadata record
-    for metadata_record in content_metadata:
-        enterprise_customer_uuids = set()
+    enterprise_uuids_for_courses = {}
```
iloveagent57: Nit: I'm not sure there's much value in jamming two different data structures into this dictionary, vs something like
```python
customer_uuid_by_course_key = defaultdict(set)
catalog_uuid_by_course_key = defaultdict(set)
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -99,31 +101,58 @@ def index_enterprise_catalog_courses_in_algolia_task(algolia_fields, content_key
...
+        # add to any existing enterprise catalog uuids
+        existing_catalog_uuids = existing_uuids.get('enterprise_catalog_uuids', set())
+        existing_uuids['enterprise_catalog_uuids'] = existing_catalog_uuids.union(enterprise_catalog_uuids)
+
+        # add to any existing enterprise customer uuids
+        existing_customer_uuids = existing_uuids.get('enterprise_customer_uuids', set())
+        existing_uuids['enterprise_customer_uuids'] = existing_customer_uuids.union(enterprise_customer_uuids)
+
+        # replace the enterprise-related uuids with the updates ones for the course
+        enterprise_uuids_for_courses[course_content_key] = existing_uuids
```
iloveagent57: ...and then I think this whole chunk can be shortened to:
```python
# add to any existing enterprise catalog uuids
existing_catalog_uuids = catalog_uuids_by_course_key.get(course_content_key)
existing_catalog_uuids.update(enterprise_catalog_uuids)

# add to any existing enterprise customer uuids
existing_customer_uuids = customer_uuid_by_course_key.get(course_content_key)
existing_customer_uuids.update(enterprise_customer_uuids)
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -99,31 +101,58 @@ def index_enterprise_catalog_courses_in_algolia_task(algolia_fields, content_key
...
+        enterprise_uuids = enterprise_uuids_for_courses[content_key]
         # add enterprise-related uuids to json_metadata
-        json_metadata = copy.deepcopy(metadata_record.json_metadata)
+        json_metadata = copy.deepcopy(metadata.json_metadata)
         json_metadata.update({
             'objectID': get_algolia_object_id(json_metadata.get('uuid')),
-            'enterprise_catalog_uuids': list(enterprise_catalog_uuids),
-            'enterprise_customer_uuids': list(enterprise_customer_uuids),
+            'enterprise_catalog_uuids': sorted(list(enterprise_uuids.get('enterprise_catalog_uuids', {}))),
+            'enterprise_customer_uuids': sorted(list(enterprise_uuids.get('enterprise_customer_uuids', {}))),
```
iloveagent57: ...and down here you use `customer_uuid_by_course_content_key` and `catalog_uuid_by_course_content_key`, and you can get rid of the `enterprise_uuids` variable.

iloveagent57 (commented):
I have a suggestion that I think would make all this logic easier to follow, but this all looks good otherwise.

adamstankiewicz:
@iloveagent57 Incorporated your feedback to simplify the logic through `defaultdict` :) Great suggestion.

iloveagent57 (approved):

PR #151, opened 2020-07-03
batch uuids for algolia objects and enable distinct to reduce record size

adamstankiewicz:
## Description

The course records that get indexed into the Algolia search index currently combine 2 lists of uuids: `enterprise_catalog_uuids` and `enterprise_customer_uuids`. Taken together, these significantly add to the avg record size. For example, CS50's record for production is 80kb where the recommended size (per Algolia) is 10kb and the max size is 100kb.

As more catalogs/customers are created, it won't be too long before we hit that max 100kb.

To address this issue, this PR does the following:
1. Splits up the `enterprise_catalog_uuids` and `enterprise_customer_uuids` into two otherwise duplicate course records. Each record will now have either `enterprise_catalog_uuids` or `enterprise_customer_uuids`, but not both. This trims the record size in half.
2. To trim down the record further, introduces logic to batch the catalog/customer uuids in several other duplicate course records. For example, if a course is associated with 880+ Enterprise Customers, as in the case of CS50, we will index 9 course records, each containing an individual batch of the uuids (up to 100).
3. Adds the `attributeForDistinct` and `distinct` settings to the Algolia index so Algolia knows how to group results (i.e., by the course `key`).

### Testing Instructions
1. To run locally, you need the stage Algolia API keys. Reach if out if you want these to manually test.
2. Ensure you have course records in your enterprise-catalog DB that are linked to catalog(s) and customer(s).
3. Run the `reindex_algolia` management command from the app container.
4. Verify results in Algolia.

## Ticket Link

[ENT-3059](https://openedx.atlassian.net/browse/ENT-3059)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -12,12 +12,16 @@ from enterprise_catalog.apps.api.v1.utils import (
 )
 from enterprise_catalog.apps.api_client.algolia import AlgoliaSearchClient
 from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
-from enterprise_catalog.apps.catalog.constants import COURSE
+from enterprise_catalog.apps.catalog.constants import (
+    COURSE,
+    DEFAULT_UUID_BATCH_SIZE,
+)
 from enterprise_catalog.apps.catalog.models import (
     ContentMetadata,
     content_metadata_with_type_course,
     update_contentmetadata_from_discovery,
 )
+from enterprise_catalog.apps.catalog.utils import batch
 
 
 logger = logging.getLogger(__name__)
@@ -86,7 +90,9 @@ def update_full_content_metadata_task(*args, **kwargs):  # pylint: disable=unuse
 
 
 @shared_task(base=LoggedTask)
-def index_enterprise_catalog_courses_in_algolia_task(algolia_fields, content_keys):
+def index_enterprise_catalog_courses_in_algolia_task(
+    algolia_fields, content_keys, uuid_batch_size=DEFAULT_UUID_BATCH_SIZE
+):
     """
     Index course data in Algolia with enterprise-related fields.
 
@@ -134,8 +140,12 @@ def index_enterprise_catalog_courses_in_algolia_task(algolia_fields, content_key
         customer_uuids_by_course_key[course_content_key].update(enterprise_customer_uuids)
 
     # iterate through only the courses, retrieving the enterprise-related uuids from the
-    # dictionary created above, and append each course to the list of courses with the added
-    # fields (e.g., objectID, enterprise_customer_uuids).
+    # dictionary created above. there is at least 2 duplicate course records per course,
+    # each including the catalog uuids and customer uuids respectively.
+    #
+    # if the number of uuids for both catalogs/customers exceeds uuid_batch_size, then
+    # create duplicate course records, batching the uuids (flattened records) to reduce
+    # the payload size of the Algolia objects.
     course_content_metadata = content_metadata.filter(content_type=COURSE)
     for metadata in course_content_metadata:
         content_key = metadata.content_key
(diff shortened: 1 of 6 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -134,19 +140,57 @@ def index_enterprise_catalog_courses_in_algolia_task(algolia_fields, content_key
...
+            courses.append(updated_json_metadata)
+        else:
+            customer_uuids = sorted(list(customer_uuids_by_course_key[content_key]))
+            for batch_index, customer_uuid_batch in enumerate(batch(customer_uuids, batch_size=uuid_batch_size)):
+                json_metadata_with_uuids = copy.deepcopy(json_metadata)
+                json_metadata_with_uuids.update({
+                    'objectID': '{}-customer-uuids-{}'.format(json_metadata['objectID'], batch_index),
+                    'enterprise_customer_uuids': customer_uuid_batch,
+                })
+                courses.append(json_metadata_with_uuids)
```
iloveagent57: This is really similar to the chunk of code above, you could use a helper function like:
```python
def _batched_metadata(json_metadata, sorted_uuids, uuid_key_name, obj_id_fmt):
    batched_metadata = []
    for batch_index, uuid_batch in enumerate(batch(sorted_uuids, batch_size=uuid_batch_size)):
        json_metadata_with_uuids = copy.deepcopy(json_metadata)
        json_metadata_with_uuids.update({
            'objectID': obj_id_fmt.format(json_metadata['objectID'], batch_index),
            uuid_key_name: uuid_batch,
        })
        batched_metadata.append(json_metadata_with_uuids)
    return batched_metadata

...
if customer_uuid_count <= uuid_batch_size:
....
else:
    customer_uuids = sorted(list(customer_uuids_by_course_key[content_key]))
    batched_metadata = _batched_metadata(
        json_metadata, customer_uuids, 'enterprise_customer_uuids', '{}-customer-uuids-{}'
    )
    courses.extend(batched_metadata)
```

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -35,6 +35,8 @@
 
 ACCESS_TO_ALL_ENTERPRISES_TOKEN = '*'
 
+DEFAULT_UUID_BATCH_SIZE = 100
```
iloveagent57: Nit: this is really specific to algolia, so maybe `ALGOLIA_UUID_BATCH_SIZE`?

enterprise_catalog/apps/api/tasks.py
iloveagent57: ...and actually, I think you could use that helper in the `if` clause, too - `batch()` should spit you out a single batch :). Which means you could get rid of the `if/else` altogether...

iloveagent57 (commented):
Suggestion for DRY-ing this up, but it looks good otherwise!

enterprise_catalog/apps/api/tasks.py
```diff
@@ -134,19 +140,57 @@ def index_enterprise_catalog_courses_in_algolia_task(algolia_fields, content_key
...
+
+        if catalog_uuid_count <= uuid_batch_size:
+            json_metadata_with_uuids = copy.deepcopy(json_metadata)
+            json_metadata_with_uuids.update({
+                'objectID': '{}-catalog-uuids-0'.format(json_metadata['objectID']),
+                'enterprise_catalog_uuids': sorted(list(catalog_uuids_by_course_key[content_key])),
+            })
+            courses.append(json_metadata_with_uuids)
+        else:
+            catalog_uuids = sorted(list(catalog_uuids_by_course_key[content_key]))
```
kellmarie: [INFO] Do these catalog_uuids need to be sorted?

adamstankiewicz: Technically, no. Curious if there is there a reason for them _not_ to be sorted? Figured sorting them will make finding specific uuids easier to find when looking at individual course records within Algolia.

kellmarie: meh, not any real reason other than reducing unnecessary work. If it makes it easier to find them when looking for individual records in Algolia then it's probably worth having.

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -96,31 +86,94 @@ def test_index_algolia(self, mock_search_client):
...
 
         # create expected Algolia data
-        expected_catalog_uuids = [enterprise_catalog_courses.uuid, enterprise_catalog_course_runs.uuid]
-        expected_customer_uuids = [
-            enterprise_catalog_courses.enterprise_uuid,
-            enterprise_catalog_course_runs.enterprise_uuid,
-        ]
-        expected_course_card_image = course_metadata.json_metadata.get('original_image')
-        course_owners = course_metadata.json_metadata.get('owners')
-        expected_partners = [{
```
adamstankiewicz, line 112: tests for asserting correct partners and course card image were added here: https://github.com/edx/enterprise-catalog/pull/153/files#diff-ea660126ab3a6ddb8f01028388db6409R76

adamstankiewicz:
@iloveagent57 Your feedback is now incorporated and ready for another pass :)

iloveagent57 (approved):
👨‍🍳 💋

PR #154, opened 2020-07-08
xAPI: Include Activity ID in course(run) metadata.

mattdrayer:
## Description

In order to enable third party systems to close the loop on learner enrollments, it's necessary to provide the xAPI Activity ID value up-front with the course metadata.  This allows the system to watch for related xAPI events to arrive from the edX system to the third party LRS with the corresponding Activity ID.

## Ticket Link

N/A

## Post-review


adamstankiewicz (approved):
:shipit:

PR #156, opened 2020-07-21
Implement usable Decentralized Devstack prototype

kdmccormick:
EDIT: To make it easier to keep this branch updated with master until enterprise-titans has chance to look at this, I rebased and squashed work in PR #195. Please add any further contributions to that PR.

--------------------------------------------
https://openedx.atlassian.net/wiki/spaces/ENG/pages/1653702661/Devstack+2.0+Spike+Notes
fyi @dianakhuang @adzuci 

Discovery PR: https://github.com/edx/course-discovery/pull/2667
Platform PR: https://github.com/edx/edx-platform/pull/24608


.env
```diff
@@ -0,0 +1,4 @@
+COMPOSE_PATH_SEPARATOR=:
+COMPOSE_FILE=docker-compose-supporting.yml:docker-compose.yml
+# TODO: Make sure this doesn't break existing local development 
+COMPOSE_PROJECT_NAME=decentralized-devstack
```
kdmccormick: I regret splitting up `docker-compose.yml` and `docker-compose-supporting.yml`. The fact that people were confused about the presence of the `.env` file shows that it doesn't lead to a good developer experience.

If we (1) combine `docker-compose.yml` and `docker-compose-supporting.yml` and (2) declare the decentralized_devstack network directly in the `docker-compose.yml` file, then we can just delete this file completely.

.travis.yml
```diff
@@ -21,6 +21,9 @@ matrix:
     - env: TESTNAME=translations
       script:
         - make validate_translations
+    - env: TESTNAME=shell_quality
+      script:
+        - make shellcheck
```
kdmccormick, line 26: FYI, I added this. I think it's a good thing, but the enterprise-catalog owners should probably be aware of it.

Makefile
```diff
@@ -6,7 +6,8 @@ TOX = ''
         migrate html_coverage upgrade extract_translation dummy_translations \
         compile_translations fake_translations  pull_translations \
         push_translations start-devstack open-devstack  pkg-devstack \
-        detect_changed_source_translations validate_translations
+        detect_changed_source_translations validate_translations \
+        docker.build docker.push docker.build.push shellcheck
```
kdmccormick, line 10: This file has similarly-named targets for both centralized and decentralized devstack operations. If we want to support both, we may want them to be distinctly different, and possible even in separate Makefiles.

docker-compose-supporting.yml
```diff
@@ -0,0 +1,90 @@
...
+    depends_on:
+      - mysql
+      - elasticsearch
+      - memcached
+    stdin_open: true
+    tty: true
+    environment:
+      TEST_ELASTICSEARCH_URL: http://edx.devstack.elasticsearch:9200
+      DJANGO_WATCHMAN_TIMEOUT: 30
+    image: kdmccormick96/course-discovery:latest
```
kdmccormick: (This is one of the action items in the [post-spike action plan](https://openedx.atlassian.net/wiki/spaces/ENG/pages/1685619678/Post-Spike+Action+Plan?focusedCommentId=1746436117#comment-1746436117) under **Make sure enterprise-catalog is pointed at the correct images.**).

Need to update all image names.

docker-compose-supporting.yml
```diff
@@ -0,0 +1,90 @@
+version: "3.7"
+
+# TODO: Stop naming containers.
```
kdmccormick: This TODO comment can be ignored and removed.

docker-compose-supporting.yml
```diff
@@ -0,0 +1,90 @@
+version: "3.7"
```
kdmccormick: As mentioned in another comment, we may as well merge this file into `docker-compose.yml`. Having them separate doesn't seem to add much other than confusion.

kdmccormick (commented):
two overarching comments:
* Do we want this to be runnable *alongside* the existing enterprise catalog devstack? If so, some changes need to be made here so that we don't steamroll their current devstack.
* This PR drops a bunch of `.sh` files and `.sql` files in the root of the repository. Not the best organization strategy. It would be cool if they could exist in a `devstack/` directory or something.

provision-lms.sh
```diff
@@ -0,0 +1,43 @@
...
+log_step "lms: Bringing up LMS..."
+docker-compose up --detach lms
+
+# TODO: Make sure this handles squashed migrations idempotently 
+# (e.g. enterprise/migrations/0001_squashed_0092_auto_20200312_1650.py)
+#log_step "lms: Running migrations for default database..."
+#service_exec_management lms migrate
+
+#log_step "lms: Running migrations for courseware student module history (CSMH) database..."
+#service_exec_management lms migrate --database student_module_history
```
kdmccormick: We ought to still run these in case the dump is out of date.
```suggestion
# TODO: Make sure this handles squashed migrations idempotently 
# (e.g. enterprise/migrations/0001_squashed_0092_auto_20200312_1650.py)
log_step "lms: Running migrations for default database..."
service_exec_management lms migrate

log_step "lms: Running migrations for courseware student module history (CSMH) database..."
service_exec_management lms migrate --database student_module_history
```

provision-lms.sh
```diff
@@ -0,0 +1,43 @@
...
+
+
+## TODO: can we handle assets during provisioning? and do we even need it for slim LMS? for login, probably...
+# log "Fixing missing vendor file by clearing the cache..."
+# service_exec lms rm /edx/app/edxapp/edx-platform/.prereqs_cache/Node_prereqs.sha1
+# log "Compiling static assets for LMS..."
+# service_exec lms paver update_assets lms
```
kdmccormick: We should be able to remove this, as assets are built into the LMS image now.
```suggestion
```

provision-lms.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+
+# Include utilities.
+source provisioning-utils.sh
+
+log_step "lms: Ensuring MySQL databases and users exist..."
+docker-compose exec -T mysql bash -c "mysql -uroot mysql" < provision-mysql-lms.sql
+
+log_step "lms: Adding default MySQL data from dump..."
+docker-compose exec -T mysql /usr/bin/mysql edxapp < provision-mysql-lms-data.sql
```
kdmccormick: Is it safe to load the dumps on an existing DB? If not, can we skip these steps if the DB is not empty?

provision-discovery.sh
```diff
@@ -0,0 +1,49 @@
...
+
+log_step "discovery: Create an edx superuser..."
+service_create_edx_user discovery
+
+log_step "discovery: Creating users and API applications for integrating with LMS..."
+create_lms_integration_for_service discovery 18381
+
+# TODO: can we handle assets during provisioning? and do we even need it for slim Discovery?
+## log "Compiling static assets for Discovery..."
+## service_exec discovery make static
```
kdmccormick:
```suggestion
```
We should build assets in to the image, which should allow us to remove this part.

docker-compose-supporting.yml
```diff
@@ -0,0 +1,90 @@
...
+    # We use WiredTiger in all environments. In development environments we use small files
+    # to conserve disk space, and disable the journal for a minor performance gain.
+    # See https://docs.mongodb.com/v3.0/reference/program/mongod/#options for complete details.
+    command: mongod --smallfiles --nojournal --storageEngine wiredTiger
+    container_name: edx.devstack.mongo
+    hostname: mongo.devstack.edx
+    image: mongo:${MONGO_VERSION:-3.6.17}
+    volumes:
+      - mongo_data:/data/db
+      - ./lms_mongo_dump:/data/dump
```
kdmccormick: @dianakhuang -- you generated this dump (along with the SQL) from your own Devstack, right? Would it be good to update the dumps to be from from a newly-provisioned devstack, or are you fine with essentially merging in your databases?

docker-compose-supporting.yml
kdmccormick: FYI @adzuci -- I recommend we name the image `discovery`, not `course-discovery` Most places I know of (GoCD, Splunk, configuration variable names, etc.) exclude the `course-` prefix.

provision-lms.sh
dianakhuang: I don't think it is safe to overwrite necessarily, but I also think that people should only want to run this when they want to clean their LMS data anyway. maybe a warning about it?

docker-compose-supporting.yml
dianakhuang: I am okay with merging my databases as they are. I believe they should be relatively clean.

provision-app.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+# Include utilities.
+source provisioning-utils.sh
+
+log_step "app: Bringing up container(s)..."
+docker-compose up -d app
```
kdmccormick:
```suggestion
docker-compose up --build --detach app
```
Because we aren't pulling this image from Docker Hub, we need to rebuild it during provisioning.

docker-compose.yml
```diff
@@ -1,41 +1,27 @@
...
   app:
     # Uncomment this line to use the official catalog base image
     # image: edxops/enterprise_catalog:devstack
     build:
-      context: .
-      dockerfile: Dockerfile
-    container_name: enterprise.catalog.app
+      context: "."
+    image: kdmccormick96/enterprise-catalog  # TODO: Change to openedx.
+    container_name: edx.devstack.enterprise-catalog
```
kdmccormick: Revert this diff; we are building image locally for now.

docker-compose.yml
```diff
@@ -1,41 +1,27 @@
...
-      - enterprise_catalog_mysql:/var/lib/mysql
-    # Uncomment these lines to access the database from localhost
-    # ports:
-    #  - "3307:3306"
 
   app:
     # Uncomment this line to use the official catalog base image
     # image: edxops/enterprise_catalog:devstack
     build:
-      context: .
```
kdmccormick: Add a `.dockerignore` file to make the `build` phase quicker.

provision-lms.sh
kdmccormick: @dianakhuang I think we want to make the provisioning scripts idempotent. It'd be solid if running `./provision.sh` on an already-provisioned devstack essentially just pulled new images, ran migrations, and re-ran the commands to create users (which are already idempotent). So I think we should skip the DB load if the LMS database already exists.

Of course, we'd still want the ability to destroy and re-provision from scratch. That could be its own script, or an argument to the `./provision.sh`.

docs/getting_started_with_decentralized_devstack.rst
```diff
@@ -0,0 +1,132 @@
...
+
+        $ docker-compose down --volumes && docker-compose pull --include-deps app && docker ps
+        $ ./provision.sh
+
+    4. Run the following to attach to your LMS container:
+
+    .. code-block:: bash
+
+        $ docker attach edx.devstack.lms
+
```
kdmccormick: Some of the indented sections are appearing as quotes in the rendered RST; de-denting them should fix it.

docs/getting_started_with_decentralized_devstack.rst
```diff
@@ -0,0 +1,132 @@
...
+Initialize and Provision
+------------------------
+    1. Verify that your virtual environment is active before proceeding
+    2. Clone the enterprise-catalog repo and **cd into that directory**
+    3. Run the following to provision a new enterprise catalog environment:
+
+    .. code-block:: bash
+
+        $ docker-compose down --volumes && docker-compose pull --include-deps app && docker ps
```
kdmccormick: I put this line in the video just to make make sure that we were started from a fair point for timing purpose. I think we can omit it here.

TODO: Provide a script option or Makefile target that wipes existing Devstack data.

docker-compose.yml
adzuci: Could we shorthand this to `build: .` like Demographics does? https://github.com/edx/demographics/blob/master/docker-compose.yml#L23

.env
adzuci: I agree that we should condense these to one docker-compose file.

docker-compose.yml
```diff
@@ -1,41 +1,27 @@
...
+      context: "."
+    image: kdmccormick96/enterprise-catalog  # TODO: Change to openedx.
+    container_name: edx.devstack.enterprise-catalog
     hostname: app.catalog.enterprise
     volumes:
       - .:/edx/app/enterprise_catalog/enterprise_catalog
       - ../src:/edx/src:cached
-    # This should be the same as the `CMD` in the devapp build stage
-    # of the Dockerfile
-    command: bash -c 'gunicorn --reload --workers=2 --name enterprise_catalog -b :18160 -c /edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/docker_gunicorn_configuration.py --log-file - --max-requests=1000 enterprise_catalog.wsgi:application'
```
adzuci, line 31: @kdmccormick FYI, Demographics found that their container would hand when hitting an exception so they did this: https://github.com/edx/demographics/blob/master/docker-compose.yml#L25

They also considered just calling `runserver` since the while loop us pretty hacky and you lose some debugging checks when running inside of gunicorn.

I'm also curious if we could remove the entire devapp build stage?  Demographics doesn't install dev requirements, though they may want to to get debug tools, but it is cleaner to be able to use the same image in multiple places: https://github.com/edx/enterprise-catalog/blame/master/Dockerfile#L62-L69

.env
katenels: Based on my experience trying to adopt this to registrar, I support condensing to one docker-compose file. I ran into errors after forgetting to copy over the .env file

docker-compose-supporting.yml
adzuci: https://github.com/edx/course-discovery/pull/2712/files
https://hub.docker.com/repository/docker/openedx/discovery

.env
jinder1s: This split is being removed in this ticket: https://openedx.atlassian.net/browse/ARCHBOM-1456 and in this PR: https://github.com/edx/enterprise-catalog/pull/165

docker-compose-supporting.yml
jinder1s: This file has been merged

docs/getting_started_with_decentralized_devstack.rst
```diff
@@ -0,0 +1,130 @@
...
+
+Initialize and Provision
+------------------------
+    1. Verify that your virtual environment is active and all requirements installed (`make requirements`) before proceeding
+    2. Clone the enterprise-catalog repo and **cd into that directory**
+    3. Run the following to provision a new enterprise catalog environment:
+
+    .. code-block:: bash
+
+        $ docker-compose pull --include-deps app && docker ps
```
kdmccormick: This line can be removed. provision.sh already runs `docker-compose pull --include-deps app`, and `docker ps` doesn't do anything (it just shows any containers that are currently running, which provision.sh will bring down anyway).

provision-lms.sh
jinder1s: this was handled in: Adding conditional before data dump load (#181)

jinder1s: TODO still left: adding ability to re-provision from scratch

Makefile
```diff
@@ -156,7 +166,7 @@ dev.up.build:  ## Runs docker-compose -up -d --build
 	docker-compose up -d --build
 
 dev.down: ## Kills containers and all of their data that isn't in volumes
-	docker-compose down
+	docker-compose downing
```
timmc-edx:
```suggestion
	docker-compose down
```

typo

Makefile
```diff
@@ -174,18 +184,25 @@ attach: ## Runs docker attach enterprise.catalog.app
...
-	docker push "openedx/enterprise-catalog:$$TRAVIS_COMMIT-newrelic"
+	docker push openedx/enterprise-catalog:latest
+	docker push openedx/enterprise-catalog:${GITHUB_SHA}
+	docker push openedx/enterprise-catalog:latest-devstack
+	docker push openedx/enterprise-catalog:${GITHUB_SHA}-devstack
+	docker push openedx/enterprise-catalog:latest-newrelic
+	docker push openedx/enterprise-catalog:${GITHUB_SHA}-newrelic
+
+shellcheck:
+	shellcheck *.sh -x
```
timmc-edx:
```suggestion
	shellcheck --external-sources *.sh
```

More explicit (didn't know what this option was) and put options before args.

decentralized_devstack/provisioning-utils.sh
```diff
@@ -0,0 +1,128 @@
...
+# Execute a shell command in a service's container.
+# Usage:
+#   service_exec SERVICE COMMAND [ARGS...]
+# Examples:
+#   service_exec discovery make migrate
+service_exec(){
+	service="$1"
+	shift
+	command_and_args="$*"
+	docker-compose exec -T "$service" bash -c "$command_and_args"
```
timmc-edx:
```suggestion
	docker-compose exec -T "$service" "$@"
```

I think this would be the correct invocation.

Makefile
jinder1s: What? huh, SI thought I already did that

timmc-edx: I think that was on a PR that never merged.

docs/getting_started_with_decentralized_devstack.rst
timmc-edx: I think the resolution here was to just tell people to run `docker-compose down --volumes` if they want to reset.

jinder1s:
To make it easier to keep this branch updated with master until enterprise-titans has chance to look at this, I rebased and squashed work in PR #195. Please add any further contribuations to that PR.

PR #157, opened 2020-08-13
Match B2C course indexing for algolia

bbaker6225:
Replicates the logic from course-discovery that restricts which courses
are indexed for algolia search.
Some pieces of the logic look a little bit different as we use fields
coming from the API whereas they use fields in the discovery database.
Additionally, there is some logic in discovery that the website team
felt was duplicated that has been omitted here.

ENT-2992

### DIFF
--- a/enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
+++ b/enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
@@ -67,6 +67,40 @@ ALGOLIA_INDEX_SETTINGS = {
 }
 
 
+def should_index(course_metadata):
+    """
+    Replicates the B2C index check of whether a certain course should be indexed for search.
+
+    A course should only be indexed for algolia search if it has a non-indexed advertiseable course run, at least
+    one owner, and a marketing url slug.
+    The course-discovery check that the course's partner is edX is included by default as the discovery API filters
+    to the request's site's partner.
+    The discovery check that the course has an availability level was decided to be a duplicate check that the
+    website team plans on removing.
+    Original code:
+    https://github.com/edx/course-discovery/blob/c6ac5329225e2f32cdf1d1da855d7c9d905b2576/course_discovery/apps/course_metadata/algolia_models.py#L218-L227
+
+    Args:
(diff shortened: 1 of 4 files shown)


bbaker6225:
This will have a few corresponding edx-internal PRs to create a new index, test that out, and then remove the old index once this is verified.
https://github.com/edx/edx-internal/pull/2798

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -67,6 +67,40 @@
 }
 
 
+def should_index(course_metadata):
```
adamstankiewicz: nit: maybe `should_index_course`?

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -67,6 +67,40 @@
...
+    course_runs = course_json_metadata.get('course_runs')
+    try:
+        advertised_course_run = [run for run in course_runs if run.get('uuid') == advertised_course_run_uuid][0]
+    except IndexError:
+        # If there is no advertised course run we can immediately return False
+        return False
+
+    owners = course_json_metadata.get('owners', [])
+    return (len(owners) > 0
+            and bool(course_json_metadata.get('url_slug'))
```
adamstankiewicz, line 100: No action needed here; just some additional context as this relates to the Learner Portal:

`url_slug` isn't used by the Learner Portal currently. We use the course key in the URL route instead of a slug. I'm not sure there's a way to retrieve course metadata from discovery based on a `url_slug` via the API. Not sure it's worth pursuing using `url_slug` in the URL routes within the Learner Portal anyways since the app doesn't have any SEO requirements.

adamstankiewicz (approved):

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
bbaker6225: Good callout. I also don't know of any way to lookup courses / course runs / programs by marketing url slug from the API, but since we don't have any need for SEO it's likely not that big of a deal

PR #159, opened 2020-08-18
Ignore health endpoint check in newrelic

bbaker6225:
Performance monitoring was being thrown off by many successful & fast
health check calls which was hiding the application's true performance.

ENT-3292


chavesj:
I'm only seeing requirements and factory changes, shouldn't there be a call to `ignore_transaction()` in the health view?

bbaker6225:
Whoops forgot to add that file

chavesj (approved):
🚢

PR #162, opened 2020-08-24
Make help the default Makefile command

raq929:
This makes the default make command consistent with our other repos
Adds docstrings for docker commands

## Description

Updates the Makefile to add docstrings and have the default make command be `help` instead of `test`.
This makes it consistent with our other repos, and gives developers a helpful list of commands.

## Ticket Link

N/A

## Post-review


Makefile
```diff
@@ -170,10 +170,10 @@ dev.stop: # Stops containers so they can be restarted
 %-logs: ## View the logs of the specified service container
 	docker-compose logs -f --tail=500 $*
 
-attach:
+attach: ## Runs docker attach enterprise.catalog.app
 	docker attach enterprise.catalog.app
 
-docker_build:
+docker_build: ## Builds with the latest enterprise catalo
```
adamstankiewicz: nit: "catalo" -> "catalog"

adamstankiewicz (approved):

PR #163, opened 2020-08-24
fix mismatching keys and secrets with what is provisioned

adamstankiewicz:
## Description

The provision-catalog.sh script in enterprise-catalog creates an oauth application using an underscore (`enterprise_catalog-backend-service-key`), yet the default client id/secret specified in devstack.py uses a hyphen (`enterprise-catalog-backend-service-key`). This results in a 401 until I change the default setting to use an underscore over a hyphen, or update the client id/secret to no longer use an underscore.

This PR does the former, changing the default settings to use snakecase `enterprise_catalog` in the keys/secrets over the hyphen. FWIW, license-manager also uses underscores in devstack.py settings.

## Post-review


bbaker6225 (approved):
hyphens and underscores strike again 😞

PR #164, opened 2020-08-24
ENT-3363: 'enrollment_url' points to Learner Portal course page if LP enabled

chavesj:
## Description

* Added edx-enterprise API Client to retrieve Enterprise Customer records
* Added memcached container to docker-compose file
* Updated get_content_enrollment_url to return LP course page if LP enabled
* Added EnterpriseCustomerDetails interface for caching Enterprise Customer data

## Associated PRs

* https://github.com/edx/frontend-app-learner-portal-enterprise/pull/91
* https://github.com/edx/edx-internal/pull/2943

## Ticket Link

[ENT-3363](https://openedx.atlassian.net/browse/ENT-3363)

### Containers you will need running: 
* Course Discovery
* LMS
* Enterprise Catalog
* Enterprise Learner Portal (LP)

### Enterprise Catalog Testing
Ensure the content metadata existing in your devstack has been migrated to your enterprise-catalog container. If not, run the `update_content_metadata` management command from your enterprise-catalog app shell. 

Lookup the UUID for a catalog associated with your Test Enterprise and ensure its LP is enabled. Hit the get_content_metadata endpoint in enterprise-catalog. Example URL: `http://localhost:18160/api/v1/enterprise-catalogs/<uuid>/get_content_metadata/`

Inspect the value for the `enrollment_url` key in each content item. The enrollment URL for Courses should point to the LP course page with UTM params, for Course Runs should point to the LP course page for the parent course with a `course_run_id` param and UTM params, and for Programs should be blank.

### Learner Portal Testing
**You will need an Enterprise Learner that is not yet enrolled in your local demonstration course.** Link the learner to your Test Enterprise and assign an Activated License to their email address.

**You will also need your demonstration course to have multiple course runs.** Navigate to the Discovery service Django Admin page ( localhost:18381/admin ). Create 2 new course runs for the demonstration course using the existing course run as prior art - I suggest using various start dates and prices (this will come in handy later). [Reindex elasticssearch for the Discovery service](https://github.com/edx/course-discovery#using-elasticsearch-locally). Use the update_content_metadata management command from the enterprise-catalog app shell to fetch the newly created course runs.

Use each enrollment_url for your demonstration course and new course runs. Observe the start date and price on the LP Course Page for each URL to verify that the data used for the page is correctly updated. 

Enrollment will fail due to the way course run data was created, that is expected. Publisher should have been used to create course runs in a way that would be fully testable - full enrollment will be tested on stage. Still, attempt to enroll and observe the failure flow correctly redirect you back to the LP course page with the 'enrollment_failed' param in the URL and the enrollment failure message rendered at the top. 

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -26,6 +26,7 @@ from enterprise_catalog.apps.catalog.tests.factories import (
     ContentMetadataFactory,
     EnterpriseCatalogFactory,
 )
+from enterprise_catalog.apps.catalog.utils import get_parent_content_key
 
 
 @ddt.ddt
@@ -531,6 +532,7 @@ class EnterpriseCatalogContainsContentItemsTests(APITestMixin):
         self.assert_correct_contains_response(url, False)
 
 
+@ddt.ddt
 class EnterpriseCatalogGetContentMetadataTests(APITestMixin):
     """
     Tests on the get_content_metadata endpoint
@@ -550,14 +552,17 @@ class EnterpriseCatalogGetContentMetadataTests(APITestMixin):
         """
         return reverse('api:v1:enterprise-catalog-get-content-metadata', kwargs={'uuid': enterprise_catalog.uuid})
 
-    def _get_expected_json_metadata(self, content_metadata):
+    def _get_expected_json_metadata(self, content_metadata, learner_portal_enabled):
         """
         Helper to get the expected json_metadata from the passed in content_metadata instance
         """
         content_type = content_metadata.content_type
         updated_json_metadata = content_metadata.json_metadata.copy()
 
-        enrollment_url = '{}/enterprise/{}/{}/{}/enroll/?catalog={}&utm_medium=enterprise&utm_source={}'
+        if learner_portal_enabled and content_type in (COURSE, COURSE_RUN):
+            enrollment_url = '{}/{}/course/{}?{}utm_medium=enterprise&utm_source={}'
+        else:
+            enrollment_url = '{}/enterprise/{}/{}/{}/enroll/?catalog={}&utm_medium=enterprise&utm_source={}'
         marketing_url = '{}?utm_medium=enterprise&utm_source={}'
         xapi_activity_id = '{}/xapi/activities/{}/{}'
 
@@ -571,14 +576,29 @@ class EnterpriseCatalogGetContentMetadataTests(APITestMixin):
             )
 
         if content_type in (COURSE, COURSE_RUN):
-            updated_json_metadata['enrollment_url'] = enrollment_url.format(
-                settings.LMS_BASE_URL,
-                self.enterprise_catalog.enterprise_uuid,
-                COURSE,
-                updated_json_metadata['key'],
-                self.enterprise_catalog.uuid,
-                self.enterprise_catalog.enterprise_name,
-            )
+            if learner_portal_enabled:
+                if content_type == COURSE:
+                    course_run_key_param = ''
+                    course_key = updated_json_metadata['key']
+                else:
+                    course_run_key_param = 'course_run_key={}&'.format(updated_json_metadata['key'])
+                    course_key = get_parent_content_key(updated_json_metadata)
+                updated_json_metadata['enrollment_url'] = enrollment_url.format(
+                    settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL,
+                    self.enterprise_slug,
+                    course_key,
+                    course_run_key_param,
+                    self.enterprise_catalog.enterprise_name
+                )
+            else:
+                updated_json_metadata['enrollment_url'] = enrollment_url.format(
+                    settings.LMS_BASE_URL,
+                    self.enterprise_catalog.enterprise_uuid,
+                    COURSE,
+                    updated_json_metadata['key'],
+                    self.enterprise_catalog.uuid,
+                    self.enterprise_catalog.enterprise_name,
+                )
             updated_json_metadata['xapi_activity_id'] = xapi_activity_id.format(
                 settings.LMS_BASE_URL,
                 content_type,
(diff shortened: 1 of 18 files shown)


kellmarie (approved):

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -642,10 +662,21 @@ def test_get_content_metadata_no_catalog_query(self):
...
+    )
+    def test_get_content_metadata(self, learner_portal_enabled, mock_api_client):
         """
         Verify the get_content_metadata endpoint returns all the metadata associated with a particular catalog
         """
+        mock_api_client.return_value.get_enterprise_customer.return_value = {
+            'slug': self.enterprise_slug,
+            'enable_learner_portal': learner_portal_enabled,
+        }
+        # Reset sequence to 10 to avoid sorting issue with 0, 1, 10, 2
```
iloveagent57: What? :)

enterprise_catalog/apps/api_client/constants.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+from urllib.parse import urljoin
+
+from django.conf import settings
+
+
+# Discovery API Client Constants
+SEARCH_ALL_ENDPOINT = urljoin(settings.DISCOVERY_SERVICE_API_URL, 'search/all/')
+COURSES_ENDPOINT = urljoin(settings.DISCOVERY_SERVICE_API_URL, 'courses/')
+OFFSET_SIZE = 100
```
iloveagent57: Nit: prefix these with `DISCOVERY_`, since they're no longer local to a file called `discovery.py`.

enterprise_catalog/apps/api_client/enterprise_cache.py
```diff
@@ -0,0 +1,50 @@
...
+        """
+        Initialize an Enterprise Customer details instance and load data from
+        cache or by using the Enterprise API client.
+
+        Arguments:
+            uuid (str): Unique identifier for the Enterprise Customer
+        """
+        self.uuid = uuid
+        self.customer_data = _get_enterprise_customer_data(uuid)
+
```
iloveagent57, line 28: Nit: one thing you could do here is expose all of the fields you need from `customer_data` via properties, for example:
```python
@property
def slug(self):
    return self.customer_data.get('slug')
```

This helps abstract away the fact that this class holds all of the data in a dictionary, and lets you use it as an object that has pertinent enterprise customer details as top-level members.  So instead of using it like this:
```python
enterprise_customer = EnterpriseCustomerDetails(self.enterprise_uuid).customer_data
slug = enterprise_customer.slug
```
you could do
```python
slug = EnterpriseCustomerDetails(self.enterprise_uuid).slug
```

enterprise_catalog/apps/catalog/models.py
```diff
@@ -213,32 +218,55 @@ def contains_content_keys(self, content_keys):
...
-            content_key,
-        )
         params = get_enterprise_utm_context(self.enterprise_name)
-        params['catalog'] = self.uuid
-
         if self.publish_audit_enrollment_urls:
             params['audit'] = 'true'
 
+        enterprise_customer = EnterpriseCustomerDetails(self.enterprise_uuid).customer_data
+        learner_portal_enabled = enterprise_customer.get('enable_learner_portal', False)
```
iloveagent57: Nit: another good candidate for a property on `EnterpriseCustomerDetails`:
```python
@property
def enable_learner_portal(self):
    return self.customer_data.get('enable_learner_portal', False)
```

iloveagent57 (commented):
Looks great, just a couple suggestions for ease-of-use.

enterprise_catalog/apps/api/v1/tests/test_views.py
chavesj: The ContentMetadataFactory creates content with keys that are generated using a string builder with a factory sequence (index is appended onto each content key). The results are sorted by key which creates an unexpected sorting of `[key0, key1, key10, key2, ...]` so the test fails on `self.assertEqual(actual_metadata, expected_metadata[:-1])`. By resetting the factory sequence to start at 10 we avoid that sorting issue.

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -642,10 +662,25 @@ def test_get_content_metadata_no_catalog_query(self):
...
         """
+        mock_api_client.return_value.get_enterprise_customer.return_value = {
+            'slug': self.enterprise_slug,
+            'enable_learner_portal': learner_portal_enabled,
+        }
+        # The ContentMetadataFactory creates content with keys that are generated using a string builder with a
+        # factory sequence (index is appended onto each content key). The results are sorted by key which creates
+        # an unexpected sorting of [key0, key1, key10, key2, ...] so the test fails on
+        # self.assertEqual(actual_metadata, expected_metadata[:-1]). By resetting the factory sequence to start at
+        # 10 we avoid that sorting issue.
```
iloveagent57, line 682: ❤️

iloveagent57 (approved):

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -571,14 +576,29 @@ def _get_expected_json_metadata(self, content_metadata):
...
+                    course_key = updated_json_metadata['key']
+                else:
+                    course_run_key_param = 'course_run_key={}&'.format(updated_json_metadata['key'])
+                    course_key = get_parent_content_key(updated_json_metadata)
+                updated_json_metadata['enrollment_url'] = enrollment_url.format(
+                    settings.ENTERPRISE_LEARNER_PORTAL_BASE_URL,
+                    self.enterprise_slug,
+                    course_key,
+                    course_run_key_param,
+                    self.enterprise_catalog.enterprise_name
```
adamstankiewicz, line 591: Since the `utm_source` query parameter is the `enterprise_name` but slugify'ed, should the test be reflecting the computed `slug` or the `enterprise_name` here (and here: https://github.com/edx/enterprise-catalog/pull/164/files#diff-ba816b6d4932e9e20da34f265bff31edR616)?

https://github.com/edx/enterprise-catalog/blob/24d88c726011bc47b143bda78f18f2a11b969100/enterprise_catalog/apps/api/v1/utils.py#L59

I don't think we make any assertions on the actual URL itself in our tests, but I may be mistaken. 

Also, I wonder if it's worth adding `enterprise_slug` as a property (maybe `@cached_property`?) to the `EnterpriseCatalog` for ease of access. Might need to pass in the `EnterpriseCatalog` instance to `get_enterprise_utm_context` instead of our current way of passing in `enterprise_name`.

adamstankiewicz: This comment can be disregarded for now. In chatting with @chavesj , we discovered the we are using a single word for the `enterprise_name` (see the `EnterpriseCatalogFactory` for reference), where it doesn't currently matter if we slugify the name or not in the URL since it's just a single word.

We agreed to follow-up on this in a separate PR/ticket as it's not blocking for this PR.

PR #165, opened 2020-08-24
ARCHBOM-1456: Reducing to only one docker-compose.yml file

jinder1s:
Previously, the services were defined in docker-compose.yml(core enterprise services) and docker-compose-supporting.yml(dependencies such as discovery and lms) files. Having multiple files was deemed unnecessary.

[ARCHBOM-1456](https://openedx.atlassian.net/browse/ARCHBOM-1456)
Case for removing .env:

- its main purpose was to help with definiing multiple docker-compose-{}.yml files. Since we are going down to just one file, that is not necessary.
- Its secondary purpose was to define the decentralized-devstack network by using COMPOSE_PROJECT_NAME env varilable. This was done to make sure all services could network with each other. This is unnecessary cause docker-compose creates a network for activated services at up and names it based on name of root directory(where docker-compose.yml lives). I would argue that is what we want: each docker-compose up call creating a new network for the DD it is creating. 

[COMPOSE_PROJECT_NAME](https://docs.docker.com/compose/reference/envvars/#compose_project_name).
[docker-compose networking](https://docs.docker.com/compose/networking/)

This PR does not fully resolve the question of how to make sure DD does not affect normal devstack. More work for that question will be done in ARCHBOM-1461


robrap (approved):
@jinder1s: Just not sure if you already got anyone else to agree on the change regarding the network.  What you say sounds reasonable, but I feel like I could easily be missing if there are any other implications.

PR #169, opened 2020-08-26
Adding dev.dd.{} targets to Makefile

jinder1s:
We needed some way to differentiating a command that only works with decentralized devstack(DD). My solution is dev.dd prefix to a command. For example, to build app image for DD, you need: dev.dd.build.  I choose this method cause it seemed like the smallest change necessary for people to start working with DD

My assumptions: any dev.{} command should either work for both DD and normal devstack or just for devstack. For dev.{} commands that only work with normal devstack, there will be DD equivalent.

Problems I see with this approach: there is no way to specifying a command only works on normal devstack. 

Clarification: this PR does not create all targets. I am hoping to set on a standard and will add further targets in another PR/ticket.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,13 +140,17 @@ validate_translations: fake_translations detect_changed_source_translations ## i
 # Docker commands below
 # TODO curate
 
-dev.build:
-	docker build . --tag kdmccormick96/enterprise-catalog # TODO
+dev.dd.build:
+	docker build . --target app --tag kdmccormick96/enterprise-catalog # TODO
 
-dev.push: dev.build
+dev.dd.push: dev.build
 	docker push kdmccormick96/enterprise-catalog # TODO
 
-dev.build.push: dev.build dev.push # TODO
+def.dd.pull:
+	docker-compose down --volumes && docker-compose pull --include-deps app
+
+dev.dd.provision:
+	bash ./provision.sh
 
 dev.provision:
 	bash ./provision-catalog.sh
@@ -166,7 +170,7 @@ dev.up.build:
 	docker-compose up -d --build
 
 dev.down: # Kills containers and all of their data that isn't in volumes
-	docker-compose downing
+	docker-compose down
 
 dev.destroy: dev.down #Kills containers and destroys volumes. If you get an error after running this, also run: docker volume rm portal-designer_designer_mysql
 	docker volume rm enterprise-catalog_enterprise_catalog_mysql


Makefile
```diff
@@ -166,7 +170,7 @@ dev.up.build:
 	docker-compose up -d --build
 
 dev.down: # Kills containers and all of their data that isn't in volumes
-	docker-compose downing
+	docker-compose down
```
jinder1s, line 173: I could not find a downing command, so I assumed this was a typo

Makefile
```diff
@@ -140,13 +140,17 @@ validate_translations: fake_translations detect_changed_source_translations ## i
...
+dev.dd.build:
+	docker build . --target app --tag kdmccormick96/enterprise-catalog # TODO
 
-dev.push: dev.build
+dev.dd.push: dev.build
 	docker push kdmccormick96/enterprise-catalog # TODO
 
-dev.build.push: dev.build dev.push # TODO
+def.dd.pull:
+	docker-compose down --volumes && docker-compose pull --include-deps app
```
kdmccormick, line 150: You do not want `docker-compose down --volumes` here. That's synonymous with `make destroy`; it deletes all volumes (which are what holds the databases).

Makefile
```diff
@@ -140,13 +140,17 @@ validate_translations: fake_translations detect_changed_source_translations ## i
...
-	docker build . --tag kdmccormick96/enterprise-catalog # TODO
+dev.dd.build:
+	docker build . --target app --tag kdmccormick96/enterprise-catalog # TODO
 
-dev.push: dev.build
+dev.dd.push: dev.build
 	docker push kdmccormick96/enterprise-catalog # TODO
 
-dev.build.push: dev.build dev.push # TODO
+def.dd.pull:
```
timmc-edx, line 149:
```suggestion
dev.dd.pull:
```

typo

kdmccormick (changes requested):
As I mentioned in Slack, I'm in favor of any of these:
1. Drop the `dev.` prefix, and just use `make dd.{}`
2. Drop the prefixes altogether. `make up`, `make shell`, `make pull-images`, etc. Nice and clean!
3. Don't even use Makefile targets. Almost all of these can be made into very simple docker-compose commands (`docker-compose up`, `docker-compose build`, `docker-compose pull`, etc.). I honestly think that we insulate people from learning how Docker and Docker Compose work by wrapping up these commands. It also allows us to let the commands get more complicated instead of fixing the underlying issues.

timmc-edx:
I am totally in favor of just using docker-compose commands directly rather than Makefile targets if they don't add too much verbosity. Makefile targets add discoverability, but I'm happy to leave that to a wiki page or versioned howto doc somewhere, since people will mostly need to know _workflows_, not individual targets anyhow. The other thing they add is a consistent interface, which is a more interesting benefit. But papering over differences in how different IDAs do devstack might not be a good thing.

If we do use Makefile targets, though... With centralized devstack, I don't think the prefixes make sense. But given that in DD, these targets have to coexist with other Makefile targets e.g. `requirements` and `clean`, a short prefix of some sort makes sense for setting them apart.

timmc-edx:
We'll also need to add all of these targets to the `.PHONY` list.

kdmccormick:
Btw, here is a shell script that I use to generate `.PHONY` declarations, in case you find it helpful :) https://github.com/kdmccormick/home/blob/master/scripts/src/make-phonies

It assumes that every Makefile target is phony.

jinder1s:
@timmc-edx @kdmccormick You guys have mostly convinced me. I am now onboard with asking folks to use docker-compose directly rather than go through make. I'm planning on bringing this up at arch tea time today, so please come to give futher input.  In the meantime, I'm gonna create a ticket to create a how-to doc which has a list of docker commands someone will need to use for DD. Final Note: this is definitely up for debate based on what input we get during prototyping.

PR #175, opened 2020-08-28
Cleanups to DD instructions

timmc-edx:
- Specify python version
- How to install docker-compose
- Remove demo-related commands
- Other clarifications

### DIFF
--- a/docs/getting_started_with_decentralized_devstack.rst
+++ b/docs/getting_started_with_decentralized_devstack.rst
@@ -1,7 +1,7 @@
 Getting Started
 ===============
 
-If you have not already done so, install create/activate a Python 3.8 `virtualenv`_.
+If you have not already done so, install create/activate a Python 3.5 or 3.8 `virtualenv`_.
 Unless otherwise stated, assume all terminal code below
 is executed within the virtualenv.
 
--- a/docs/getting_started_with_decentralized_devstack.rst
+++ b/docs/getting_started_with_decentralized_devstack.rst
@@ -7,13 +7,12 @@ is executed within the virtualenv.
 
 .. _virtualenv: https://virtualenvwrapper.readthedocs.org/en/latest/
 
-You will also need to install docker and docker-compose.
-The latter can be installed in the venv with `pip install docker-compose`.
+You will also need to install docker.
 
 
 Initialize and Provision
 ------------------------
-    1. Verify that your virtual environment is active before proceeding
+    1. Verify that your virtual environment is active and all requirements installed (`make requirements`) before proceeding
     2. Clone the enterprise-catalog repo and **cd into that directory**
     3. Run the following to provision a new enterprise catalog environment:
 
--- a/docs/getting_started_with_decentralized_devstack.rst
+++ b/docs/getting_started_with_decentralized_devstack.rst
@@ -1,12 +1,14 @@
 Getting Started
 ===============
 
-If you have not already done so, install create/activate a `virtualenv`_. Unless otherwise stated, assume all terminal code
-below is executed within the virtualenv.
+If you have not already done so, install create/activate a Python 3.8 `virtualenv`_.
+Unless otherwise stated, assume all terminal code below
+is executed within the virtualenv.
 
 .. _virtualenv: https://virtualenvwrapper.readthedocs.org/en/latest/
 
 You will also need to install docker and docker-compose.
+The latter can be installed in the venv with `pip install docker-compose`.
 
 
 Initialize and Provision
@@ -17,14 +19,7 @@ Initialize and Provision
 
     .. code-block:: bash
 
-        $ docker-compose down --volumes && docker-compose pull --include-deps app && docker ps
         $ ./provision.sh
-
-    4. Run the following to attach to your LMS container:
-
-    .. code-block:: bash
-
-        $ docker attach edx.devstack.lms
 
 Viewing Enterprise Catalog
 --------------------------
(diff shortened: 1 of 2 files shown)


docs/getting_started_with_decentralized_devstack.rst
```diff
@@ -52,7 +47,10 @@ Migrating Catalog Data from LMS to the Catalog Service
 ------------------------------------------------------
 You may already have enterprise catalog data persisted in your local LMS (edx-platform) database.  The edx-enterprise
 library provides a `migrate_enterprise_catalogs <https://github.com/edx/edx-enterprise/blob/master/enterprise/management/commands/migrate_enterprise_catalogs.py>`_
-management command that will copy those existing catalogs and their metadata into your local catalog service.  From your **devstack** directory, do the following:
+management command that will copy those existing catalogs and their metadata into your local catalog service.
```
timmc-edx, line 49: I'm actually not sure what's supposed to happen here... we can't have both new and old devstack fully running at the same time, so I think these instructions are incomplete.

docs/getting_started_with_decentralized_devstack.rst
```diff
@@ -17,15 +19,8 @@ Initialize and Provision
 
     .. code-block:: bash
 
-        $ docker-compose down --volumes && docker-compose pull --include-deps app && docker ps
```
robrap, line 20: Why is this not needed?  I feel I needed this. I think I may have even needed this to re-provision. Maybe we could just add `#` comments before each line, and describe what it might be useful for.

docs/getting_started_with_decentralized_devstack.rst
```diff
@@ -1,12 +1,14 @@
 Getting Started
 ===============
 
-If you have not already done so, install create/activate a `virtualenv`_. Unless otherwise stated, assume all terminal code
-below is executed within the virtualenv.
+If you have not already done so, install create/activate a Python 3.8 `virtualenv`_.
```
robrap: As Jinder mentioned, I think many of us have been using Python 3.5, since that is still what is in Production.  Maybe we should use that until the repo switches to Python 3.8 more officially?

docs/getting_started_with_decentralized_devstack.rst
```diff
@@ -1,12 +1,14 @@
...
+Unless otherwise stated, assume all terminal code below
+is executed within the virtualenv.
 
 .. _virtualenv: https://virtualenvwrapper.readthedocs.org/en/latest/
 
 You will also need to install docker and docker-compose.
+The latter can be installed in the venv with `pip install docker-compose`.
```
robrap: For people like me that skim docs, it's nice to add commands as code blocks:
```
.. code-block:: bash
    
    $ pip install docker-compose
```

docs/getting_started_with_decentralized_devstack.rst
```diff
@@ -17,15 +19,8 @@ Initialize and Provision
...
     .. code-block:: bash
 
-        $ docker-compose down --volumes && docker-compose pull --include-deps app && docker ps
         $ ./provision.sh
 
-    4. Run the following to attach to your LMS container:
-
-    .. code-block:: bash
-
-        $ docker attach edx.devstack.lms
```
robrap, line 27: Can you explain why we wouldn't want this somewhere?

docs/getting_started_with_decentralized_devstack.rst
robrap: Agreed.  We'll need to ask @adzuci or @kdmccormick.

docs/getting_started_with_decentralized_devstack.rst
timmc-edx: It occurs to me now that docker-compose should probably be part of the `dev.in` requirements.

docs/getting_started_with_decentralized_devstack.rst
timmc-edx: `provision.sh` includes `docker-compose down` and `docker-compose pull --include-deps app`, and the `--volumes` flag apparently would have the effect of destroying the DB.

docs/getting_started_with_decentralized_devstack.rst
timmc-edx: I don't think it does anything useful here. It can show logs in realtime, I think, but we have a better command for that.

docs/getting_started_with_decentralized_devstack.rst
robrap: Are you planning on making that change?

docs/getting_started_with_decentralized_devstack.rst
robrap: Got it.  Thanks.

docs/getting_started_with_decentralized_devstack.rst
robrap: I guess people can add what they need when they need it. :)

robrap (approved):
@timmc-edx: Feel free to land this and iterate on the migration help in a later PR when we learn more.  That's up to you.

docs/getting_started_with_decentralized_devstack.rst
timmc-edx: Yes, unless someone thinks it's a bad idea.  It sounds like people on Macs tend to install both at the same time in one package, whereas on Linux I install just docker -- I don't know if there's any issue with having a pip-installed version shadow an OS-installed version of docker-compose.

robrap: I don't know either.  We can test that?  But, it might just make sense for now to mention this step if you need it, and we can loop back.  I don't think this should be a blocker, and I don't want to turn it into one with bad instructions for mac users.

timmc-edx: Come to think of it, I made the same change a while back with devstack, and no one has complained. So I think it is safe.

PR #180, opened 2020-09-02
ARCHBOM-1488: Standardizing build tags and exposed ports

jinder1s:
docker targets in Makefile have been modified to match standards.

Exposed port for devstack is now **** instead of 1****. This will match port in production.

Link to the associated ticket
[Link title](https://openedx.atlassian.net/browse/ARCHBOM-1488)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,21 +184,25 @@ attach: ## Runs docker attach enterprise.catalog.app
 	docker attach enterprise.catalog.app
 
 docker_build: ## Builds with the latest enterprise catalog
-	docker build . --target app -t "openedx/enterprise-catalog:latest"
-	docker build . --target newrelic -t "openedx/enterprise-catalog:latest-newrelic"
+	docker build . --target app -t openedx/enterprise-catalog:latest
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -184,21 +184,25 @@ attach: ## Runs docker attach enterprise.catalog.app
 	docker attach enterprise.catalog.app
 
 docker_build: ## Builds with the latest enterprise catalog
-	docker build . --target app -t "openedx/enterprise-catalog:latest"
-	docker build . --target newrelic -t "openedx/enterprise-catalog:latest-newrelic"
+	docker build . --target app -t openedx/enterprise-catalog:latest
```
jmbowman, line 187: Did you confirm the quotes aren't needed?

jmbowman (approved):
Looks good as long as dropping those quotes doesn't break anything.

Makefile
jinder1s: They are not needed by docker command and its used without quotes in other repositories, so I don't think they are necessary

PR #182, opened 2020-09-02
Removing unnecessary mongo dump files and fixed command to use dump

jinder1s:
- The *.gz files have the same data as the files deleted here.

- Piping mongorestore erros to mongorestore.log
These errors occur if provision script is run a second or more time. mongorestore complains
because the data its trying to restore already exists. We are  ignoring these errors(by piping errors to mongoresotre.log)


.gitignore
```diff
@@ -86,4 +86,4 @@ private.py
 
 *.trace
 
-docs/_build/
+docs/_build/
```
timmc-edx: Spurious loss of newline here.

decentralized_devstack/provision-lms.sh
```diff
@@ -25,8 +25,12 @@ done
 log_step "lms: Creating MongoDB users..."
 docker-compose exec -T mongo bash -c "mongo" < decentralized_devstack/provision-mongo.js
 
-log_step "lms: Adding default MongoDB data..."
-service_exec mongo mongorestore --gzip /data/dump
+log_step "lms: Adding default MongoDB data, piping errors to mongorestore.log"
+# if this command is run on database with default data, it will throw out errors complaining about
+# duplicates(E11000 duplicate key error collection), we are  ignoring these errors(by piping errors to mongoresotre.log)
```
timmc-edx:
```suggestion
# duplicates (E11000 duplicate key error collection), we are ignoring these errors (by piping errors to mongorestore.log)
```
typos

decentralized_devstack/provision-lms.sh
```diff
@@ -25,8 +25,12 @@ done
...
+# because mongorestore behaves how we want it to: Not replace data that already exists
+# Possible problems: this hides real errors that occur
+service_exec mongo mongorestore --gzip /data/dump 2>mongorestore.log
```
timmc-edx: If mongo does fail to load the dump, does it exit with non-zero exit code, stopping the script?

Either way, might be worth running `head mongorestore.log` and `echo $(wc -l mongorestore.log) total stderr lines` or something.

jinder1s: I agree with head mongorestore.log, but I don't think adding info on number of lines helps, especially cause normal behaviour has so many lines to begin with.

decentralized_devstack/provision-lms.sh
jinder1s: Doing it in next commit

decentralized_devstack/provision-lms.sh
timmc-edx: Is it still noisy if we pass `--quiet`?

On the number of lines, I meant that that could be an indication to the user that there's more than just what was printed.

If it halts on error, then we'll never reach the `head mongorestore.log`. Here's an option:

```
service_exec mongo mongorestore --quiet --gzip /data/dump 2>mongorestore.log || {
  log_step "Mongo restore failed. See mongorestore.log for details."
  exit 1
}
```

Could get fancier and capture the real exit code, but this is close enough. :-)

jinder1s: actually yes, quiet is def a better option than what we are doing. My rational for not using --quiet was wrong

jinder1s: Alright, I just switched it to quiet

decentralized_devstack/provision-lms.sh
```diff
@@ -25,8 +25,12 @@ done
 log_step "lms: Creating MongoDB users..."
 docker-compose exec -T mongo bash -c "mongo" < decentralized_devstack/provision-mongo.js
 
-log_step "lms: Adding default MongoDB data..."
-service_exec mongo mongorestore --gzip /data/dump
+log_step "lms: Adding default MongoDB data, piping errors to mongorestore.log"
```
timmc-edx: Nit: This log statement is no longer true. :-)

So what's the behavior now? Quiet on success, not overly verbose on failure?

timmc-edx: (same for comment a few lines down)

jinder1s: Yep, quiet on success, thought not sure about failure

jinder1s: I would hope they have good error output if the step fails. Its weird to have errors that don't lead to failure

jinder1s:
@timmc-edx  I think this is good to go now

decentralized_devstack/provision-lms.sh
```diff
@@ -26,7 +26,11 @@ log_step "lms: Creating MongoDB users..."
 docker-compose exec -T mongo bash -c "mongo" < decentralized_devstack/provision-mongo.js
 
 log_step "lms: Adding default MongoDB data..."
-service_exec mongo mongorestore --gzip /data/dump
+# FYI, without --quiet flag below, when this command is run on database which was previously provisioned,
+# it will throw out errors complaining about duplicates(E11000 duplicate key error collection),
+# we are  ignoring these errors(using --quiet flag),
```
timmc-edx: Nit: I don't think we're really ignoring the errors -- the `--quiet` flag should only suppress info messages, not error messages or error exit code.

jinder1s: @timmc-edx I removed alot from this, what do you think of the rewording?

decentralized_devstack/provision-lms.sh
```diff
@@ -27,9 +27,7 @@ docker-compose exec -T mongo bash -c "mongo" < decentralized_devstack/provision-
 
 log_step "lms: Adding default MongoDB data..."
 # FYI, without --quiet flag below, when this command is run on database which was previously provisioned,
-# it will throw out errors complaining about duplicates(E11000 duplicate key error collection),
-# we are  ignoring these errors(using --quiet flag),
-# because mongorestore behaves how we want it to: Not replace data that already exists
+# it will complaining about duplicates(E11000 duplicate key error collection)
```
timmc-edx: It will still complain about duplicates even with `--quiet` though, right?

jinder1s: Nope, those errors are not output with --quiet flag

decentralized_devstack/provision-lms.sh
```diff
@@ -26,7 +26,9 @@ log_step "lms: Creating MongoDB users..."
 docker-compose exec -T mongo bash -c "mongo" < decentralized_devstack/provision-mongo.js
 
 log_step "lms: Adding default MongoDB data..."
-service_exec mongo mongorestore --gzip /data/dump
+# FYI, without --quiet flag below, when this command is run on database which was previously provisioned,
+# it will complaining about duplicates(E11000 duplicate key error collection)
```
timmc-edx:
```suggestion
# When mongorestore is run on database which was previously provisioned,
# it will exit with code zero (which is OK for our purposes) but also will emit
# a large number of warnings about duplicates (E11000 duplicate key error collection).
# The --quiet flag is to make the logs less noisy.
```

I think this might describe it better? I'm also OK with not continuing to rat-hole on this and merging as-is, but did want to take a crack at describing the situation a little differently.

timmc-edx (approved):

PR #186, opened 2020-09-03
Adding target to docker-compose.yml file

jinder1s:
This would allow us to build service images with "docker-compose build". And changed images to point to openedx instead of kdmccormick dockerhub

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,16 +1,12 @@
 version: "3.7"
 
-# TODO: Stop naming containers.
-
 services:
 
   app:
-    # Uncomment this line to use the official catalog base image
-    # image: edxops/enterprise_catalog:devstack
+    image: openedx/enterprise_catalog:latest-devstack
     build:
       context: "."
       target: devstack
-    image: kdmccormick96/enterprise-catalog  # TODO: Change to openedx.
     container_name: edx.devstack.enterprise-catalog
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -35,11 +32,10 @@ services:
       ENABLE_DJANGO_TOOLBAR: 1
 
   worker:
-    # Uncomment this line to use the official catalog base image
-    # image: edxops/enterprise_catalog:devstack
+    image: openedx/enterprise_catalog:latest-devstack
     build:
       context: "."
-    image: kdmccormick96/enterprise-catalog  # TODO: Change to openedx.
+      target: devstack
```
kdmccormick: the `image` key declares an image that can be pulled down from dockerhub and used.

the `build` key instructs docker-compose how to build the image locally.

I believe we'd only want one or the other--either it should be built locally (my preference) or it should get pulled down from dockerhub.

jinder1s: I definitely also prefer build to pull, but am more on the side of giving both as an option. I will think on this for a bit

jinder1s: Okay, I'm with you on just the build

jinder1s: @kdmccormick this should be good to go now

kdmccormick (approved):

PR #187, opened 2020-09-03
Adding option to destroy volumes to provision.sh

jinder1s:
Shell commands are copy pasta from https://pretzelhands.com/posts/command-line-flags

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/decentralized_devstack/provision.sh
+++ b/decentralized_devstack/provision.sh
@@ -9,13 +9,13 @@ cd -- "$DIR"/..
 source decentralized_devstack/provisioning-utils.sh
 
 # whether provision script should overwrite existing data
-SHOULD_RESET=false
+START_FROM_SCRATCH=false
 # Loop through arguments and process them
 for arg in "$@"
 do
 	case $arg in
-		--reset)
-		SHOULD_RESET=true
+		--from-scratch)
+		START_FROM_SCRATCH=true
 		shift # Remove --reset from processing
 		;;
 		*)
@@ -26,9 +26,9 @@ done
 
 log_step "Starting provisioning process..."
 
-if [ "$SHOULD_RESET" = true ] ; then
+if [ "$START_FROM_SCRATCH" = true ] ; then
 	log_step 'Bringing down existing containers and starting provisioning from scratch...'
-	docker-compose down --volume
+	docker-compose down --volumes
 else
 	log_step "Bringing down any existing containers..."
 	docker-compose down
--- a/decentralized_devstack/provision.sh
+++ b/decentralized_devstack/provision.sh
@@ -27,13 +27,13 @@ done
 log_step "Starting provisioning process..."
 
 if [ "$SHOULD_RESET" = true ] ; then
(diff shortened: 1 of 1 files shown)


decentralized_devstack/provision.sh
```diff
@@ -8,10 +8,31 @@ cd -- "$DIR"/..
 # shellcheck source=provisioning-utils.sh
 source decentralized_devstack/provisioning-utils.sh
 
+# whether provision script should overwrite existing data
+SHOULD_RESET=false
+# Loop through arguments and process them
+for arg in "$@"
+do
+	case $arg in
+		--reset)
```
jinder1s: there might be a better name for this flag.

decentralized_devstack/provision.sh
```diff
@@ -8,11 +8,32 @@ cd -- "$DIR"/..
...
+		SHOULD_RESET=true
+		shift # Remove --reset from processing
+		;;
+		*)
+		shift # Remove generic argument from processing
```
timmc-edx, line 22: Probably want to yell about unrecognized argument here and exit early so that typos are caught quickly.

timmc-edx (approved):
One suggestion, but looks fine.

timmc-edx:
On second thought, I wonder if it actually would make more sense just to have this as a separate instruction, and not have it as part of the provisioning scripts. It's a pretty short command.

decentralized_devstack/provision.sh
```diff
@@ -8,10 +8,31 @@ cd -- "$DIR"/..
...
+	esac
+done
+
 log_step "Starting provisioning process..."
 
-log_step "Bringing down any existing containers..."
-docker-compose down
+if [ "$SHOULD_RESET" = true ] ; then
+	log_step 'Bringing down existing containers and starting provisioning from scratch...'
+	docker-compose down --volume
```
kdmccormick:
```suggestion
	docker-compose down --volumes
```
missing `s` here

decentralized_devstack/provision.sh
kdmccormick: in old devstack, "reset" was non-destructive, so this might confuse an unsuspecting user into destroying their data.

I suggest something scarier, like `--destroy`, `--reprovision`, or `--from-scratch`

jinder1s:
Agreed, it might just be good to leave this to documentation. I'm gonna put off merging this till we have more documentation

kdmccormick (commented):
Nice. I like that provisioning is non-destructive by default, so that one can use the provision script to try to get back to sane state before fully reprovisioning.

kdmccormick:
> On second thought, I wonder if it actually would make more sense just to have this as a separate instruction, and not have it as part of the provisioning scripts. It's a pretty short command.

Also not a bad idea. It goes along the lines of the "don't make wrappers around simple commands" mantra that we've been rolling with.

jinder1s:
This will no longer be merged. We decided to go the route of documentation the dockerk-compose down -v command rather than creating an option for it in provision script

PR #189, opened 2020-09-04
Adding back files to use normal devstack

jinder1s:
Adding .env file to allow users to switch to decentralized devstack by uncommenting line setting COMPOSE_FILE
Moved DD docker-compose.yml file to decentralized_devstack dir
Adding back docker-compose.yml and provision-catalog.sh to allow users to continue using normal devstack


.env
```diff
@@ -0,0 +1,2 @@
+# TODO: Uncomment the following line to use Decentralized Devstack
+# COMPOSE_FILE=decentralized_devstack/.env
```
kdmccormick:
```suggestion
# COMPOSE_FILE=decentralized_devstack/docker-compose.yml
```

decentralized_devstack/docker-compose.yml
```diff
@@ -0,0 +1,145 @@
+version: "3.7"
+
```
kdmccormick: Nice file placement. I forget whether having the docker-compose.yml in a subdirectory causes weirdness or not. But I figure that you'll test it.

decentralized_devstack/docker-compose.yml
```diff
@@ -0,0 +1,145 @@
...
+    hostname: redis.devstack.edx
+    image: redis:2.8
+
+
+volumes:
+  edxapp_lms_assets:
+  discovery_assets:
+  elasticsearch_data:
+  mysql_data:
+  mongo_data:
```
kdmccormick: are these isolated from legacy devstack?

decentralized_devstack/docker-compose.yml
```diff
@@ -0,0 +1,145 @@
...
+    stdin_open: true
+    tty: true
+    environment:
+      CELERY_ALWAYS_EAGER: 'false'
+      CELERY_BROKER_TRANSPORT: redis
+      CELERY_BROKER_HOSTNAME: edx.devstack.redis:6379
+      CELERY_BROKER_VHOST: 0
+      CELERY_BROKER_PASSWORD: password
+      DJANGO_SETTINGS_MODULE: enterprise_catalog.settings.devstack
+      ENABLE_DJANGO_TOOLBAR: 1
```
kdmccormick: (unrelated to this PR): you could probably stuff these into the Dockerfile now that the docker-compose.yml is separate.

decentralized_devstack/docker-compose.yml
jinder1s: No clue, still testing the way legacy devstack and DD interact

decentralized_devstack/docker-compose.yml
jinder1s: Your intuition is correct, I have to change any relative paths for this to work

.env
jinder1s: clearly I woke up too early

.env
```diff
@@ -0,0 +1,2 @@
+# TODO: Uncomment the following line to use Decentralized Devstack
+# COMPOSE_FILE=decentralized-docker-compose.yml
```
robrap: 1. Can you add more comments at the top explaining the purpose of this file, including when and if it could be removed?
2. Should we add make commands to comment/uncomment as needed to switch back and forth between devstacks? Not sure if anything else is needed when switching.

decentralized_devstack/docker-compose.yml
jinder1s: Yes, they are. The names for volumes different between the two compose files and because of how the names are prefixed during creation(with name of currect directory), they should not conflict with volumes for devstack.

jinder1s:
Converting this to draft cause alot more things need to be done before legacy devstack works with DD

enterprise_catalog/settings/devstack.py
```diff
@@ -52,6 +52,7 @@
 
 LMS_BASE_URL = 'http://edx.devstack.lms:18000'
 DISCOVERY_SERVICE_API_URL = 'http://edx.devstack.discovery:18381/api/v1/'
+ENTERPRISE_LEARNER_PORTAL_BASE_URL = 'http://localhost:8734'
```
jinder1s, line 55: I copied this file from master, so this line came from master

docker-compose.yml
```diff
@@ -1,23 +1,41 @@
-version: "3.7"
-
+version: "2.1"
```
jinder1s, line 1: This file is a copy of docker-compose.yml from master. This should allow them to continue using normal devstack during prototype

enterprise_catalog/settings/devstack.py
```diff
@@ -39,9 +39,9 @@
     'default': {
         'ENGINE': 'django.db.backends.mysql',
         'NAME': os.environ.get('DB_NAME', 'enterprise_catalog'),
-        'USER': os.environ.get('DB_USER', 'entcatalog001'),
-        'PASSWORD': os.environ.get('DB_PASSWORD', 'password'),
-        'HOST': os.environ.get('DB_HOST', 'edx.devstack.mysql'),
+        'USER': os.environ.get('DB_USER', 'root'),
+        'PASSWORD': os.environ.get('DB_PASSWORD', ''),
+        'HOST': os.environ.get('DB_HOST', 'enterprise.catalog.mysql'),
```
jinder1s, line 44: this is undoing changes made for DD

.env
jinder1s: 1) I added the comments, let me know it is sufficient
2) I would prefer not to do any optimizing on making switch back and forth in this PR. I'm working on documentation in #190, based on learnings from that work, we'll see if further make commands are necessary.

.env
```diff
@@ -0,0 +1,9 @@
+# Purpose of file: Define environment variables that configure docker
+
+# This is added for Decentralized Devstack prototyping, specifically,
+# the variable COMPOSE_FILE allows us to define which file is used by
+# docker-compose command. Unless the line below is uncommented, legacy
+# devstack should work as normal.
+
+# TODO: Uncomment the following line to use Decentralized Devstack
+#COMPOSE_FILE=decentralized-docker-compose.yml
```
kdmccormick, line 9: :100:

Dockerfile
```diff
@@ -59,8 +59,19 @@ FROM app as newrelic
 RUN pip3 install newrelic
 CMD ["newrelic-admin", "run-program", "gunicorn", "--workers=2", "--name", "enterprise_catalog", "-c", "/edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/docker_gunicorn_configuration.py", "--log-file", "-", "--max-requests=1000", "enterprise_catalog.wsgi:application"]
 
+# Creating image which will be used by Decentralized Devstack
```
kdmccormick, line 62: Nice use of build stages.

decentralized-docker-compose.yml
```diff
@@ -0,0 +1,145 @@
...
+    stdin_open: true
+    tty: true
+    environment:
+      CELERY_ALWAYS_EAGER: 'false'
+      CELERY_BROKER_TRANSPORT: redis
+      CELERY_BROKER_HOSTNAME: edx.devstack.redis:6379
+      CELERY_BROKER_VHOST: 0
+      CELERY_BROKER_PASSWORD: password
+      DJANGO_SETTINGS_MODULE: enterprise_catalog.settings.decentralized_devstack
+      ENABLE_DJANGO_TOOLBAR: 1
```
kdmccormick: non-blocking suggestion: You could move some or all of these into either the `devstack` stage of the Dockerfile or the `decentralized_devstack.py` file.

pros:
* Keeps settings/envs in fewer places.
* Docker image is closer to being usable out-of-the-box without compose file.
* Less duplication between app and worker.
* Simpler compose file.

cons:
* Some of those settings won't work without Redis being present. So you could argue that those settings are specific to the docker-compose level, not the docker level.

.env
```diff
@@ -0,0 +1,9 @@
+# Purpose of file: Define environment variables that configure docker
+
+# This is added for Decentralized Devstack prototyping, specifically,
```
robrap, line 3: This is great. Thanks.
I’d expect a README update that explains the steps required for switching each direction. Note: could be a follow-up PR.

enterprise_catalog/settings/decentralized_devstack.py
```diff
@@ -0,0 +1,69 @@
+"""
+Settings file for decentralized devstack
```
kdmccormick: non-blocking consideration:

The corresponding settings file in edx-platform is named `devstack_decentralized.py`: https://github.com/edx/edx-platform/blob/master/lms/envs/devstack_decentralized.py . I chose the backwards name to be consistent with the other devstack settings file: `devstack_docker`, `devstack_optimized`, etc.  You may or may not want to follow that pattern.

kdmccormick (approved):
LGTM and WRFM (works right for me)

Just two suggestions; feel free to ignore them.

enterprise_catalog/settings/decentralized_devstack.py
jinder1s: that is a good idea, will switch

PR #198, opened 2020-09-24
Adding redis as dependency to worker

jinder1s:
Worker is failing in decentralized devstack without redis, this should fix that.

### DIFF
--- a/decentralized-docker-compose.yml
+++ b/decentralized-docker-compose.yml
@@ -40,6 +40,7 @@ services:
       - discovery
       - lms
       - mysql
+      - redis
     environment:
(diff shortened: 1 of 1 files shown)


bbaker6225 (commented):
Would there still be the issue since the broker is `CELERY_BROKER_HOSTNAME: edx.devstack.redis:6379`?

jinder1s:
@bbaker6225 There doesn't seem to be an issue in the logs for worker container. I tested with and without this change and adding dependency got rid of the connection error.

bbaker6225 (approved):

PR #205, opened 2020-10-05
Allow updating a specified subset of catalogs

bbaker6225:
Adds an optional kwarg `catalog_uuids` to the update_content_metadata
command which allows users to only update the ContentMetadata for the
catalogs that are associated with the catalog query of the provided
catalog.
Note that this does not change the update_full_content_metadata task to
only do that process for the subset of content associated with the
catalog queries of the enterprise catalogs that are given here. That is
being split out to try and make the refactoring cleaner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - 3.5
   - 3.8
 env:
   matrix:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,6 +1,5 @@
 language: python
 python:
-  - 3.5
```
iloveagent57, line 3: 🎉

iloveagent57 (approved):
LGTM

PR #207, opened 2020-10-06
Prefetch related objects in reindex-algolia task

bbaker6225:
Prefetches the associated catalog queries and their associated
enterprise catalogs to avoid the duplicate queries from ManyToMany and
ReverseForeignKey lookups from ContentMetadata --> CatalogQuery -->
EnterpriseCatalog.
This also drops the `enterprise_catalogs` property as that seemed to
circumvent some of Django's prefetching.
Additionally, this drops the `.values` call as that was clearing the
prefetched objects cache due to being a different query.

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -93,18 +93,6 @@ class CatalogQuery(models.Model):
             )
         )
 
-    @property
-    def enterprise_catalogs(self):
-        """
-        Helper to retrieve the enterprise catalogs associated with the catalog query.
-
-        Returns:
-            Queryset: The queryset of associated enterprise catalogs
-        """
-        if not self.enterprise_catalogs:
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -93,18 +93,6 @@ def __str__(self):
...
-        """
-        Helper to retrieve the enterprise catalogs associated with the catalog query.
-
-        Returns:
-            Queryset: The queryset of associated enterprise catalogs
-        """
-        if not self.enterprise_catalogs:
-            return EnterpriseCatalog.objects.none()
-        return self.enterprise_catalogs.all()
-
```
bbaker6225, line 107: `self.enterprise_catalogs.all()` already did this same logic as well

iloveagent57 (approved):
Hooray, ORMs, again!

chavesj (approved):
noice

PR #209, opened 2020-10-08
Removing lms sql data dump from repository

jinder1s:
and linking discovery to right mysql container

edx-platform repo now has the sql dump regularly updated, so using that sql dump instead of the one which was in this repository.
This also has work from this PR: https://github.com/edx/enterprise-catalog/pull/210

### DIFF
--- a/.env
+++ b/.env
@@ -5,5 +5,22 @@
 # docker-compose command. Unless the line below is uncommented, legacy
 # devstack should work as normal.
 
-# TODO: Uncomment the following line to use Decentralized Devstack
+# TODO: Uncomment one of the following lines with "COMPOSE_FILE" based on your use case
+
+# UNCOMMENT the following line to use latest Decentralized Devstack, this will pull latest images of each service
 #COMPOSE_FILE=decentralized-docker-compose.yml
+
+# UNCOMMENT the following line if you want to develop in local version of dependencies(services)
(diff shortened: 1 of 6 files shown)


decentralized_devstack/provision-lms.sh
```diff
@@ -12,7 +12,11 @@ else
   docker-compose exec -T mysql mysql -uroot mysql < decentralized_devstack/provision-mysql-lms.sql
 
   touch /tmp/provision-mysql-lms-data.sql
-  curl https://raw.githubusercontent.com/edx/edx-platform/master/edxapp.sql > /tmp/provision-mysql-lms-data.sql
+
+  # read variable out of .env file
+  #TODO: document further
```
adzuci: I think this line could be dropped or replaced with ~`# Pull a MySQL dump from `edx-platform` that is created via a Github Action in order to speed up the startup time and ensure expected data is available."

.env
```diff
@@ -5,5 +5,22 @@
 # docker-compose command. Unless the line below is uncommented, legacy
 # devstack should work as normal.
 
-# TODO: Uncomment the following line to use Decentralized Devstack
+# TODO: Uncomment one of the following lines with "COMPOSE_FILE" based on your use case
```
adzuci: I think this line can be dropped.

adzuci (approved):

PR #210, opened 2020-10-13
Using multiple docker-compose files to expose different developer workflow

jinder1s:
While researching devstack, it became clear that we need to support multiple developer needs:
1- Some developers don't need most latest version of dependencies, they would rather use some stable versions
2- Some developers want to just develop in their main service
3- Some developers want to develop in multiple services at the same time

In this PR, I propose a solution of using multiple docker-compose-*.yml to configure containers for different usage:
The 2nd case is taken care of by default decentralized-docker-compose.yml file
1st case would be taken care of with decentralized-docker-compose-stable.yml 
3rd case would be taken care of with decentralized-docker-compose-develop-dependencies.yml

This might potentially make complexity worse, so feel free to say that if you think that is the case.

The stable case has few potential problems in during provisioning(especially if we are using the sql data dump)


.env
```diff
@@ -5,5 +5,26 @@
 # docker-compose command. Unless the line below is uncommented, legacy
 # devstack should work as normal.
 
-# TODO: Uncomment the following line to use Decentralized Devstack
-#COMPOSE_FILE=decentralized-docker-compose.yml
+# TODO: Uncomment one of the following lines with "COMPOSE_FILE" based on your use case
+
+# Uncomment the following line to use latest Decentralized Devstack, this will pull latest images of each service
+#COMPOSE_FILE=decentralized-docker-compose.yml:decentralized-docker-compose-stable.yml
```
kdmccormick: I think you mean
```suggestion
#COMPOSE_FILE=decentralized-docker-compose.yml
```

decentralized-docker-compose-stable.yml
```diff
@@ -0,0 +1,11 @@
...
+
+services:
+
+  discovery:
+    image: openedx/discovery:${DISCOVERY_STABLE_TAG}-devstack
+
+  lms:
+    image: openedx/edx-platform:${EDX_PLATFORM_STABLE_TAG}-devstack
+
+
```
kdmccormick: In interest of keeping the number of compose files under control, I think you can skip the `-develop-dependencies.yml` file and instead do this in the main compose file:

```yaml
services:
  # ... other services
  discovery:
    # The ${X:-y} syntax tries to evaluate $X, but falls back to "y" if it is undefined
    image: openedx/discovery:${DISCOVERY_STABLE_TAG:-latest}-devstack
    container_name: edx.devstack.discovery
    hostname: discovery.devstack.edx
    # ... rest of the service definition
```

Then, users would select stable/unstable for each service by uncommenting/commenting the individual `*_STABLE_TAG` lines in `.env`.

decentralized-docker-compose-develop-dependencies.yml
```diff
@@ -0,0 +1,13 @@
...
+
+  discovery:
+    volumes:
+      - ${COURSE_DISCOVERY_PATH}:/edx/app/discovery/discovery:cached
+
+  lms:
+    volumes:
+      - ${EDX_PLATFORM_PATH}:/edx/app/edxapp/edx-platform:cached
+
+
```
kdmccormick: Good thinking using multiple compose files to split things out. I think that might be the best way to go with volumes.

You might just want to [mimic legacy devstack](https://github.com/edx/devstack/blob/master/docker-compose-host.yml) and have a single `DEVSTACK_WORKSPACE` variable for the parent folder of all one's repos.

.env
jinder1s: yep

decentralized-docker-compose-stable.yml
jinder1s: thanks for the suggestion, this is a great idea

adzuci (approved):

PR #212, opened 2020-10-14
Update manual & auto catalog refresh functionality

bbaker6225:
* Updates the refresh_metadata endpoint for a catalog to perform all the
steps of `update_content_metadata`, `update_full_content_metadata`, and
`reindex_algolia` one after the other for the content associated with
the given catalog's CatalogQuery. This is done both to bring the
endpoint back in sync with the functionality of the aforementioned
scheduled jenkins jobs, as well as to give ECS and developers an easy
way to update a single CatalogQuery at a time (used in onboarding new
EnterpriseCustomers / EnterpriseCatalogs).
* Significant refactoring of the current tasks used by the jobs to
support both flows
* Minor quality and test fixes

[ENT-2450](https://openedx.atlassian.net/browse/ENT-2450)


enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -111,3 +162,13 @@ def test_get_course_subjects(self, course_metadata, expected_subjects):
         """
         course_subjects = get_course_subjects(course_metadata)
         assert sorted(course_subjects) == sorted(expected_subjects)
+
+    @mock.patch('enterprise_catalog.apps.catalog.algolia_utils.AlgoliaSearchClient')
+    def test_get_initialized_algolia_client(self, mock_search_client):
```
bbaker6225, line 167: Other than this test, this file is just moving things and small import/naming updates

enterprise_catalog/apps/catalog/models.py
```diff
@@ -480,27 +480,21 @@ def __repr__(self):
...
     client = DiscoveryApiClient()
 
-    try:
-        catalog_query = CatalogQuery.objects.get(id=catalog_query_id)
-    except CatalogQuery.DoesNotExist:
-        catalog_query = None
-
-    if not catalog_query:
-        LOGGER.error('Could not find a CatalogQuery with id %s', catalog_query_id)
-        return
```
bbaker6225, line 502: This is pulled up since it now gets a catalog query

enterprise_catalog/apps/catalog/management/commands/tests/test_reindex_algolia.py
```diff
@@ -1,108 +1,45 @@
...
 from enterprise_catalog.apps.catalog.tests.factories import (
     ContentMetadataFactory,
 )
 
 
-@ddt.ddt
 class ReindexAlgoliaCommandTests(TestCase):
     command_name = 'reindex_algolia'
 
-    @ddt.data(
```
bbaker6225, line 24: Moved to `test_algolia_utils.py`

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -5,101 +5,18 @@
...
+)
 from enterprise_catalog.apps.catalog.models import (
     content_metadata_with_type_course,
 )
-from enterprise_catalog.apps.catalog.utils import batch
 
 
 logger = logging.getLogger(__name__)
 
-BATCH_SIZE = 250
```
bbaker6225, line 17: Most of this file was pulled out into `algolia_utils`

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+def get_initialized_algolia_client():
+    """
+    Initializes, configures, and returns an Algolia client for updating search indices
+    """
+    algolia_client = AlgoliaSearchClient()
+    algolia_client.init_index()
+    algolia_client.set_index_settings(ALGOLIA_INDEX_SETTINGS)
+    return algolia_client
```
bbaker6225, line 115: These are the only "new" functions in this file, although they just capture prior logic

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -95,7 +95,10 @@ def get_courses(self, query_params=None):
         Returns:
             list: a list of the results, or None if there was an error calling the discovery service.
         """
-        request_params = {'limit': DISCOVERY_OFFSET_SIZE}
+        request_params = {
+            'ordering': 'key',
```
bbaker6225, line 101: Moved the ordering into here since we seem to always want it anyway

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -253,7 +259,19 @@ def get_permission_object(self):
...
+        async_update_metadata_chain = chain(
+            update_catalog_metadata_task.s(catalog_query_id),
+            # Runs the `update_full_content_metadata_task` with the content keys that were associated in the
+            # `update_catalog_metadata_task` to pad the metadata from discovery's /search/all endpoint with additional
+            # data from the /courses endpoint
+            update_full_content_metadata_task.s(),
+            # Runs the indexing task with the indexable course keys that were returned from the
+            # `update_full_content_metadata_task` to index those pieces of ContentMetadata in Algolia
+            index_enterprise_catalog_courses_in_algolia_task.s(ALGOLIA_FIELDS),
+        )
```
bbaker6225, line 281: This is the meat of the PR here. The auto-magic passing of parameters through the signatures in the chain is a little too hidden for my tastes, but I've tried to comment what's going on so future folks understand

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -15,40 +15,77 @@
 )
 
 
-class EnterpriseCatalogCeleryTaskTests(TestCase):
```
bbaker6225, line 18: Mostly just separating the task tests out so we can more easily add additional testing to each one

enterprise_catalog/apps/api/tasks.py
```diff
@@ -28,45 +30,74 @@
...
+        query_params['keys'] = ','.join(course_keys_chunk)
+        courses.extend(discovery_client.get_courses(query_params=query_params))
 
-    content_keys = [metadata.content_key for metadata in content_metadata]
-    courses_in_content_metadata = [course for course in courses if course.get('key') in content_keys]
+    return courses
 
-    if not courses_in_content_metadata:
-        logger.error('Could not find any ContentMetadata records that match courses retrieved from course-discovery.')
-        return
```
bbaker6225, line 65: This bit was removed since we're now always providing keys to discovery, so we already have a `content_key` for anything that comes back

enterprise_catalog/apps/api/tasks.py
```diff
@@ -104,7 +138,9 @@ def _batched_metadata(json_metadata, sorted_uuids, uuid_key_name, obj_id_fmt, uu
 
 @shared_task(base=LoggedTask)
 def index_enterprise_catalog_courses_in_algolia_task(
-    algolia_fields, content_keys, uuid_batch_size=ALGOLIA_UUID_BATCH_SIZE
+    content_keys,
```
bbaker6225, line 151: This arg needs to be first for the magical `return_value` passing with signatures

enterprise_catalog/apps/api/tasks.py
```diff
@@ -113,94 +149,111 @@ def index_enterprise_catalog_courses_in_algolia_task(
...
+
     if not algolia_fields or not content_keys:
         logger.error('Must provide algolia_fields and content_keys as arguments.')
         return
 
-    # initialize the Algolia index
-    algolia_client = AlgoliaSearchClient()
-    algolia_client.init_index()
+    # Update the index in batches
+    for content_keys_batch in batch(content_keys, batch_size=ALGOLIA_BATCH_SIZE):
```
bbaker6225: This batching was previously done in the management command level

enterprise_catalog/apps/api/tasks.py
```diff
@@ -28,52 +30,80 @@
...
+        list of dict: Returns a list of dictionaries where each dictionary represents the course data from discovery.
+    """
+    discovery_client = DiscoveryApiClient()
 
-    # find all ContentMetadata records with a content type of "course"
-    content_metadata = content_metadata_with_type_course()
+    # Batch the course keys into smaller chunks so that we don't send too big of a request to discovery
+    batched_course_keys = batch(course_keys, batch_size=DISCOVERY_COURSE_KEY_BATCH_SIZE)
+    courses = []
+    query_params = {}
```
iloveagent57: Nit: you don't really need this outside the for loop, you could just have `query_params = {'keys': ','.join(course_keys_chunk)}` inside the body.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -28,52 +30,80 @@
...
+    logger.info('Retrieved %d courses from course-discovery.', len(courses))
+
+    # Iterate through the courses to update the json_metadata field, merging the minimal json_metadata retrieved by
+    # /search/all/ with the full json_metadata retrieved by /courses/.
     updated_metadata = []
-    for course_metadata in courses_in_content_metadata:
+    for course_metadata in courses:
         content_key = course_metadata.get('key')
         try:
             metadata_record = ContentMetadata.objects.get(content_key=content_key)
```
iloveagent57: Nit (and not a change you made): We could fetch these in advance in a single query with something like:
```python
metadata_records_by_key = ContentMetadata.objects.filter(content_key__in=[course['key'] for course in course_metadata])
for course_metadata in courses:
    content_key = course_metadata.get('key')
    metadata_record = metadata_records_by_key.get(content_key)
    if not metadata_record:
        continue
....
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -120,94 +155,111 @@ def index_enterprise_catalog_courses_in_algolia_task(
...
+    Args:
+        catalog_query_id (str): The id for the catalog query to update.
+    Returns:
+        list of str: Returns the content keys for ContentMetadata objects that were associated with the query. This is
+            passed to the `update_full_content_metadata_task` from the `EnterpriseCatalogRefreshDataFromDiscovery` view
+    """
+    try:
+        catalog_query = CatalogQuery.objects.get(id=catalog_query_id)
+    except CatalogQuery.DoesNotExist:
+        catalog_query = None
```
iloveagent57: Nit: you don't need this, since you return inside the `except` block. If the linter is complaining, you can either declare `catalog_query = None` before the `try/except`, or move lines 264-265 into the `try` block.

iloveagent57 (commented):
Couple of nits and a small suggestion for now, I'm only halfway done though :)

enterprise_catalog/apps/api/tasks.py
bbaker6225: Yep forgot this in the change

enterprise_catalog/apps/api/tasks.py
bbaker6225: I might be misunderstanding you, but the intent of this is that each time through the loop the keys query param will change so we only look at that chunk of course keys. i.e.

first loop:
`?keys=course1,course2,...`

second loop:
`?keys=course10,course11,...`

enterprise_catalog/apps/api/tasks.py
bbaker6225: Yeah that would also work. I think it's a bit more readable the way it currently is and I think that's a reasonable trade-off for a _not too bad_ query in an async task.

I'm open to changing it though if others disagree 🤷‍♂️

enterprise_catalog/apps/api/tasks.py
iloveagent57: Right, I'm just griping about readability - declaring `query_params` outside of the for loop makes me kinda think that it's accumulating more and more useful data with each pass of the loop, but really, each pass of the loop is just reassigning the single key that's stored inside it.

bbaker6225: Ah yes, I get what you're saying. Makes sense

enterprise_catalog/apps/api/tasks.py
iloveagent57: We're nearing 20k ContentMetadata records, so I think we're looking at the same number of queries (at least) when we run the periodic job to update all metadata.  Divide that by 100 (or whatever the batch size is), and you're looking at a much smaller number.  Plus it'll help these tasks finish sooner, and it's not _that_ hard to grok. :)

Also, you probably already noticed this, but that first line in my code snippet should be something like
```python
metadata_records_by_key = {
    metadata.key: metadata for metadata in 
    ContentMetadata.objects.filter(
        content_key__in=[course['key'] for course in course_metadata]
    )
}
```

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -253,7 +259,19 @@ def get_permission_object(self):
     def post(self, request, uuid):
```
iloveagent57, line 259: Could you add a docstring here (or comment below) along the lines of "...auto-magic passing of parameters through the signatures in the chain..."

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -111,27 +28,18 @@ def handle(self, *args, **options):
         Initializes and configures the settings for an Algolia index, and then spins off
         a task for each batch of content_keys to reindex course data in Algolia.
         """
-        # initialize the Algolia index
-        algolia_client = AlgoliaSearchClient()
-        algolia_client.init_index()
-
-        # configure the Algolia index
-        algolia_client.set_index_settings(ALGOLIA_INDEX_SETTINGS)
+        get_initialized_algolia_client()
```
iloveagent57, line 32: Nit: it's not really clear anymore that you're caling this to configure the index settings, maybe just keep that `# configure the Algolia index` comment.

enterprise_catalog/apps/catalog/management/utils.py
```diff
@@ -0,0 +1,10 @@
+""" Utility functions for catalog management """
+from enterprise_catalog.apps.catalog.models import ContentMetadata
+
+
+def get_all_content_keys():
+    """
+    Returns a list of content keys for all ContentMetadata objects.
+    """
+    all_content_metadata = ContentMetadata.objects.all()
+    return [content_metadata.content_key for content_metadata in all_content_metadata]
```
iloveagent57: Along the same lines as "there's about 20k of these objects"...we should consider turning this into a generator that fetches smaller batches of metadata at a time and yields the individual keys.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -28,52 +30,79 @@
...
+    Gets the list of keys of course ContentMetadata objects that need to be updated.
+
+    Args:
+        content_keys_for_updating (list of str): Content keys for ContentMetadata objects that are to be updated.
+    Returns:
+        list of str: Returns the list of keys that represent courses from the provided `content_keys_for_updating`.
     """
-    discovery_client = DiscoveryApiClient()
+    related_course_content_metadata = ContentMetadata.objects.filter(
+        content_key__in=content_keys_for_updating,
```
iloveagent57, line 43: Will this IN clause be too long if we're running the mgmt command that updates every metadata record?  I think you can only send MySQL query strings that are about 65k characters long.

iloveagent57 (commented):
Couple more comments about updating every ContentMetadata record - maybe we can tweak the tasks to batch work up before doing anything else?

enterprise_catalog/apps/api/tasks.py
bbaker6225: Good points, I'll update this

enterprise_catalog/apps/api/tasks.py
bbaker6225: Ooh I did not think about that. I think the batching idea makes sense either way though

enterprise_catalog/apps/api/v1/views.py
bbaker6225: Added a more in depth comment right above the `chain(...)` call

enterprise_catalog/apps/catalog/management/utils.py
bbaker6225: I agree that's a good idea. The only problem with that is that you can't pass generators to tasks as they aren't JSON serializable. We could loop over the generated keys and then call the tasks in a chord for each batch, but I don't know if that's the cleanest solution

bbaker6225:
@iloveagent57 thanks for the suggestions, I think I've updated everything now with the exception of `get_all_content_keys` would ideally return a generator. I don't have an immediate solution that I love, but maybe we just make everything a chord to batch tasks 🤷‍♂️

enterprise_catalog/apps/catalog/management/utils.py
iloveagent57: Yup, I just pushed a commit to make it a smaller result set with `values_list()` - I think that's good enough for now.

enterprise_catalog/apps/api/tasks.py
chavesj: ^ this seems worth including as a comment

chavesj (approved):
Besides the one nit about adding a comment, LGTM 🍝

enterprise_catalog/apps/api/v1/views.py
adamstankiewicz: 🥩 Appreciate the comments here :)

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
```diff
@@ -34,12 +34,22 @@ def _run_update_catalog_metadata_task(self, catalog_query):
...
+        """
         message = (
             'Spinning off update_full_content_metadata_task from update_content_metadata command'
             ' to replace minimal json_metadata from /search/all/ with full json_metadata from /courses/.'
         )
         logger.info(message)
-        return update_full_content_metadata_task.s()
+
+        all_content_keys = get_all_content_keys()
+        # task.si() is used as a shortcut for an immutable signature to avoid calling this with the results from the
```
adamstankiewicz, line 49: 🔥

adamstankiewicz (approved):

PR #213, opened 2020-10-16
ENT-2450 | Fix early return in update_full_content_metadata_task

iloveagent57:
## Description

When updating batches of full content metadata, don't early return just because one batch contained no keys with content_type COURSE

## Ticket Link

https://openedx.atlassian.net/browse/ENT-2450

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -97,9 +97,9 @@ def update_full_content_metadata_task(content_keys):
 
         courses = _fetch_courses_by_keys(course_keys_for_updating)
         if not courses:
-            logger.error('No courses were retrieved from course-discovery.')
-            return []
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -97,9 +97,9 @@ def update_full_content_metadata_task(content_keys):
 
         courses = _fetch_courses_by_keys(course_keys_for_updating)
         if not courses:
-            logger.error('No courses were retrieved from course-discovery.')
-            return []
-        logger.info('Retrieved %d courses from course-discovery.', len(courses))
+            logger.info('No courses were retrieved from course-discovery in this batch.')
+            continue
+        logger.info('Retrieved %d courses from course-discovery in this batch.', len(courses))
```
chavesj, line 102: it could be useful to include `content_keys` as part of these logs

chavesj (approved):

PR #214, opened 2020-10-19
Support accurate build statuses for jobs

bbaker6225:
Updates management commands to collect the results of asynchronous
tasks in order to report if they were successful, or propagate any
exception that might have been raised.

ENT-3544


chavesj (approved):
LGTM

adamstankiewicz (approved):

PR #220, opened 2020-11-05
ENT-3547: adding caching for data from the Discovery API

chavesj:
## Description
* added `CatalogQueryMetadata` interface that uses cache
* only retrieve `Course` data from the Discovery API for `ContentMetadata` records that haven't been recently updated

## Ticket Link
[ENT-3547](https://openedx.atlassian.net/browse/ENT-3547)

## Testing Instructions
* Run `./manage.py update_content_metadata` from `app-shell` in devstack multiple times.
* Observe in the `worker-logs` that CatalogQuery metadata is cached on the first run and that ContentMetadata is not retrieved from the Discovery API in subsequent runs.


enterprise_catalog/settings/base.py
```diff
@@ -327,9 +327,10 @@
 SOCIAL_AUTH_REDIRECT_IS_HTTPS = False
 STATICFILES_STORAGE = 'django.contrib.staticfiles.storage.StaticFilesStorage'
 
-# How long we keep Enterprise Customer details from edx-enterprise in cache. (seconds)
-# Default = 1 hour.
-ENTERPRISE_CUSTOMER_CACHE_TIMEOUT = 60 * 60
+# How long we keep API Client data in cache. (seconds)
+ENTERPRISE_CUSTOMER_CACHE_TIMEOUT = 60 * 60  # 1 hour
+DISCOVERY_CATALOG_QUERY_CACHE_TIMEOUT = 60 * 60  # 1 hour
+DISCOVERY_COURSE_DATA_CACHE_TIMEOUT = 60 * 60  # 1 hour
```
bbaker6225: nit: I might write out 60 * 60 as an hour and then just share it since we're using the same timeout for everything

enterprise_catalog/apps/catalog/tests/test_models.py
```diff
@@ -20,7 +20,8 @@
 class TestModels(TestCase):
     """ Models tests. """
 
-    @mock.patch('enterprise_catalog.apps.catalog.models.DiscoveryApiClient')
+    @override_settings(DISCOVERY_CATALOG_QUERY_CACHE_TIMEOUT=0)
```
bbaker6225, line 23: This feels a little clunky to me. It might be worth having a way to more easily enable/disable the cache both for local development and tests

enterprise_catalog/apps/catalog/models.py
```diff
@@ -482,33 +482,25 @@ def __repr__(self):
...
-    metadata = client.get_metadata_by_query(catalog_query, query_params=query_params)
+
+    # metadata will be an empty dict if unavailable from cache or API.
+    metadata = CatalogQueryMetadata(catalog_query).metadata
 
     # associate content metadata with a catalog query only when we get valid results
     # back from the discovery service. if metadata is `None`, an error occurred while
     # calling discovery and we should not proceed with the below association logic.
-    if metadata is not None:
+    if metadata != {}:
```
bbaker6225: Maybe just `if metadata`?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -17,11 +17,11 @@
     get_enterprise_utm_context,
     update_query_parameters,
 )
-from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.api_client.enterprise_cache import (
     EnterpriseCustomerDetails,
 )
 
+from ..api_client.discovery_cache import CatalogQueryMetadata
```
bbaker6225: nit: It doesn't seem like it's actually enforced in pylint anymore, but I think we generally try not to do this sort of relative import that this file has a few of (and instead do like `from enterprise_catalog.apps.api_client.enterprise_cache import EnterpriseCustomerDetails`)

I think it was something for open source (?), but mostly just a consistency nit

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
```diff
@@ -50,10 +50,29 @@ def _run_update_full_content_metadata_task(self, *args, **kwargs):
         # https://docs.celeryproject.org/en/master/userguide/canvas.html#immutability
         return update_full_content_metadata_task.si(all_content_keys)
 
+    def add_arguments(self, parser):
+        # Argument to specify catalogs to update
+        parser.add_argument(
+            '--catalog_uuids',
+            nargs='+',
+        )
```
bbaker6225, line 56: Is there a reason you're adding this back? There's a decent amount of work to make this an option in jenkins, and I would suggest keeping the logic on the simpler side if it's not adding much value

enterprise_catalog/apps/api_client/discovery_cache.py
```diff
@@ -0,0 +1,73 @@
+"""
+Interfaces to the discovery API using a volatile cache.
+"""
+from logging import getLogger
+
+from django.conf import settings
+from django.core.cache import cache
+
+from .constants import DISCOVERY_CATALOG_QUERY_CACHE_KEY_TPL
+from .discovery import DiscoveryApiClient
```
bbaker6225: nit: I shall continue to be a grumpy old man about the inconsistency

enterprise_catalog/apps/api_client/discovery_cache.py
```diff
@@ -0,0 +1,73 @@
...
+            client = DiscoveryApiClient()
+            query_params = {
+                # Omit non-active course runs from the course-discovery results
+                'exclude_expired_course_run': True,
+                # Increase number of results per page for the course-discovery response
+                'page_size': 100,
+                # Ensure paginated results are consistently ordered by `aggregation_key` and `start`
+                'ordering': 'aggregation_key,start',
+            }
+            catalog_query_data = client.get_metadata_by_query(catalog_query, query_params=query_params)
```
bbaker6225: Almost a nit, but I think having this full implementation here is a bit too business-logicy for a cache file. You can probably just move it into the discovery client layer?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -55,15 +60,35 @@ def _fetch_courses_by_keys(course_keys):
...
         list of dict: Returns a list of dictionaries where each dictionary represents the course data from discovery.
     """
+    courses = []
+    course_keys_to_fetch = []
     discovery_client = DiscoveryApiClient()
 
+    # Check for each course key in cache. If cache hit, remove key from list to be requested from Discovery API.
+    for key in course_keys:
+        cache_key = DISCOVERY_COURSE_DATA_CACHE_KEY_TPL.format(key=key)
+        course_data = cache.get(cache_key)
```
bbaker6225: This will probably work, but I think I might recommend against dumping these large chunks of discovery data into the cache. Especially since we already have the course metadata stored in the database, so we can use that like a cache (maybe if the content metadata's modified time is within the cache timeout range or something?). 
The reason I even bring this up here and not for the `search/all/` metadata is that `course-discovery`and some other parts of edX have run into issues trying to cache chunks around and above the size of the `/v1/courses/` entries which actually resulted in slower times when trying to get things from memcached

enterprise_catalog/apps/api_client/discovery_cache.py
bbaker6225: I'll note that I don't _really_ care though

enterprise_catalog/apps/catalog/tests/test_models.py
chavesj: I would advise against disabling cache locally as the default, if someone wants to disable for their own development they should edit `CACHES` in the local.py settings file. I've also tried using the Django dummy cache in the test.py settings file and that doesn't solve the problem of cache being used when you run the test locally. My opinion is to either leave this as is or override the `CACHES` setting here to use the Django dummy cache (which effectively are the same).

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
chavesj: oh lol I was adding it because the option was present in the docstring, figured it was something small that got missed. will remove 👍

enterprise_catalog/apps/api_client/discovery_cache.py
chavesj: since this is the only foreseeable use case for `DiscoveryApiClient.get_metadata_by_query` I think it's acceptable to leave here for now

enterprise_catalog/apps/api/tasks.py
chavesj: sounds good, I'll add an `updated_at` field to the `CatalogQuery` model 👍

chavesj: Sounds good, I ended up removing the use of cache and instead used the `TimeStampedModel` `modified` field of `ContentMetadata` as a way of checking against `DISCOVERY_COURSE_DATA_CACHE_TIMEOUT`

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
bbaker6225: whoops that was me forgetting to remove it from the docstring then, thanks

enterprise_catalog/apps/api_client/discovery_cache.py
bbaker6225: You're right that it probably won't ever be used elsewhere. I more meant that we can hide a lot of the implementation here and just have
```python
catalog_query_data = client.get_metadata_by_query(catalog_query)
```
and then inside `get_metadata_by_query` we define those query parameters. Not a big deal though

enterprise_catalog/apps/catalog/tests/test_models.py
bbaker6225: ok fine with me

enterprise_catalog/apps/api/tasks.py
```diff
@@ -55,15 +58,32 @@ def _fetch_courses_by_keys(course_keys):
     Returns:
         list of dict: Returns a list of dictionaries where each dictionary represents the course data from discovery.
     """
+    courses = []
+    course_keys_to_fetch = []
     discovery_client = DiscoveryApiClient()
+    timeout_seconds = settings.DISCOVERY_COURSE_DATA_CACHE_TIMEOUT
+
+    # Check for each course key in DB. If recently updated, remove key from list to be requested from Discovery API.
```
bbaker6225: Maybe something like "Update course keys list to only request data for ContentMetadata objects that have not changed recently"?

bbaker6225 (approved):
I think this is in a pretty good state now 👍 . Once it goes in I think it's a good idea to monitor the cache usage in AWS for a few job runs / catalog updates and see how it does

enterprise_catalog/apps/api/tasks.py
chavesj: Reworded to "Populate a new list of course keys that haven't been updated recently to request from the Discovery API"

enterprise_catalog/apps/api_client/discovery_cache.py
chavesj: ahh okay I see now, updating 👍

PR #222, opened 2020-11-23
Map system ent. admin role to catalog learner

bbaker6225:
This is to ensure that enterprise admins with the system enterprise
admin role are properly mapped to a feature role in enterprise-catalog.
This mapping allows admins to have the same catalog access as their
learners, granting access to the `get_content_metadata` and
`contains_content_items` endpoints used in the learner portal.

ENT-3672


chavesj (approved):
LGTM

PR #228, opened 2020-12-14
Management command processing changes

bbaker6225:
- Bump all task timeout limits again
- Propagate exceptions from eager task applications to more quickly
  catch errors
- Run `reindex_algolia` and `update_full_content_metadata` management
  commands synchronously. This change is to try and ensure that the
  jobs fail in jenkins if there is an error in any one of the tasks.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,9 +183,6 @@ docker_build: ## Builds with the latest enterprise catalog
 	docker build . --target devstack -t openedx/enterprise-catalog:latest-devstack
 	docker build . --target newrelic -t openedx/enterprise-catalog:latest-newrelic
 
-clean:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -183,9 +183,6 @@ docker_build: ## Builds with the latest enterprise catalog
 	docker build . --target devstack -t openedx/enterprise-catalog:latest-devstack
 	docker build . --target newrelic -t openedx/enterprise-catalog:latest-newrelic
 
-clean:
```
bbaker6225, line 186: there is already a `clean` target that does this and more

iloveagent57 (approved):

PR #229, opened 2020-12-15
Add base task for auto-retrys on known exceptions

bbaker6225:
ENT-3594

Example:
image

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -4,8 +4,10 @@ from collections import defaultdict
 from datetime import timedelta
 
 from celery import shared_task
+from celery.exceptions import SoftTimeLimitExceeded
 from celery_utils.logged_task import LoggedTask
 from django.conf import settings
+from django.db import IntegrityError
 from django.db.models import Prefetch, Q
 from django.utils import timezone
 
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -87,7 +89,25 @@ def _fetch_courses_by_keys(course_keys):
...
-@shared_task(base=LoggedTask)
+class LoggedTaskWithRetry(LoggedTask):  # pylint: disable=abstract-method
+    """
+    Shared base task that allows tasks that raise some common exceptions to retry automatically.
+
+    See https://docs.celeryproject.org/en/stable/userguide/tasks.html#automatic-retry-for-known-exceptions for
+    more documentation.
+    """
+    autoretry_for = (
+        IntegrityError,
```
kellmarie, line 100: Why does retrying on an IntegrityError work? Is there sometimes a background change with the db that just needs to finish for the task succeed on a retry?

chavesj (approved):
nice :shipit:

enterprise_catalog/apps/api/tasks.py
bbaker6225: Our theory is that the integrity errors occasionally pop up when two different processes are trying to create the same ContentMetadata object at the same time. Retrying (manually) typically fixes the issue and what I found online suggested that transactions in celery tasks (especially long ones) are an expensive and not especially common thing

kellmarie: Makes sense to me.

kellmarie (approved):

adamstankiewicz (approved):

iloveagent57 (approved):

PR #240, opened 2021-02-05
feat: prototypes a logging utility to audit memory usage

chavesj:
## Description

Adds the utility function `print_memory_snapshot` to audit memory usage

## Ticket Link

[ENT-4061](https://openedx.atlassian.net/browse/ENT-4061)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/utils.py
+++ b/enterprise_catalog/apps/catalog/utils.py
@@ -13,7 +13,10 @@ from edx_rest_framework_extensions.auth.jwt.authentication import (
 from edx_rest_framework_extensions.auth.jwt.cookies import \
     get_decoded_jwt as get_decoded_jwt_from_cookie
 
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/catalog/utils.py
```diff
@@ -104,3 +105,22 @@ def batch(iterable, batch_size=1):
...
+    log details of the n largest allocations including the line of source code
+    responsible, size of memory allocated for the given object, the count (number of
+
+
+    Arguments:
+        num_allocations (int): number of allocation details to log
+    """
+    snapshot = tracemalloc.take_snapshot()
+    stats = ''
+    for stat in snapshot.statistics('filename')[:num_allocations]:
```
iloveagent57: Suggestion: Maybe `lineno` here instead of `filename`?

iloveagent57 (commented):
Good idea.  It'd be nice to put this behind a Django Settings feature toggle, so that we could easily enable/disable on prod and stage environments.

PR #247, opened 2021-02-18
feat: adding distinct-catalog-queries endpoint

chavesj:
## Description

* Adds distinct-catalog-queries endpoint to determine the number of distinct CatalogQueries used by the given set of EnterpriseCatalog UUIDs

## Ticket Link

[ENT-4096](https://openedx.atlassian.net/browse/ENT-4096)

## Testing Instructions

* Use Postman to submit requests using the Authorization header with the JWT of an admin user
* Verify that the list of unique CatalogQuery IDs, as well as the count of unique IDs, is returned in the response

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views.py
+++ b/enterprise_catalog/apps/api/v1/views.py
@@ -364,14 +364,11 @@ class DistinctCatalogQueriesView(APIView):
         """
         Method to handle POST requests to this endpoint
         """
-        distinct_catalog_query_ids = set()
         enterprise_catalog_uuids = request.data.get('enterprise_catalog_uuids', [])
 
-        enterprise_catalogs = EnterpriseCatalog.objects.filter(
+        distinct_catalog_query_ids = EnterpriseCatalog.objects.filter(
             uuid__in=enterprise_catalog_uuids,
-        )
-        for catalog in enterprise_catalogs:
-            distinct_catalog_query_ids.add(catalog.catalog_query.id)
+        ).distinct().values_list('catalog_query__id', flat=True)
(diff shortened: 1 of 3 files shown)


kellmarie (approved):

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -341,3 +341,40 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
+        Method to handle POST requests to this endpoint
+        """
+        distinct_catalog_query_ids = set()
+        enterprise_catalog_uuids = request.data.get('enterprise_catalog_uuids', [])
+
+        enterprise_catalogs = EnterpriseCatalog.objects.filter(
+            uuid__in=enterprise_catalog_uuids,
+        )
+        for catalog in enterprise_catalogs:
+            distinct_catalog_query_ids.add(catalog.catalog_query.id)
```
iloveagent57: Nit: you could do this entirely in the ORM as 
```python
distinct_catalog_query_ids = EnterpriseCatalog.objects.filter(
    uuid__in=enterprise_catalog_uuids,
).distinct().values_list('catalog_query__id', flat=True)
```

iloveagent57 (approved):
Looks good, one small suggestion about letting the DB do more work for us.

enterprise_catalog/apps/api/v1/views.py
iloveagent57: sorry, did a quick ninja edit to update to `catalog_query__id`.

PR #248, opened 2021-02-18
Put a semaphore on the update_catalog_metadata_task, so we stop doing…

iloveagent57:
… duplicate work. ENT-4081

## Description

Makes it so that `update_catalog_metadata_task()` can successfully run on a given catalog query id at most once per hour.

## Ticket Link

[Ticket about improving catalog task situation](https://openedx.atlassian.net/browse/ENT-4081)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -1,9 +1,10 @@
 import copy
+import functools
 import logging
 from collections import defaultdict
 from datetime import timedelta
 
-from celery import shared_task
+from celery import shared_task, states
 from celery.exceptions import SoftTimeLimitExceeded
 from celery.exceptions import TimeoutError as CeleryTimeoutError
 from celery_utils.logged_task import LoggedTask
@@ -12,6 +13,7 @@ from django.db import IntegrityError
 from django.db.models import Prefetch, Q
 from django.db.utils import OperationalError
 from django.utils import timezone
+from django_celery_results.models import TaskResult
 
 from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.catalog.algolia_utils import (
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -91,6 +93,48 @@ def _fetch_courses_by_keys(course_keys):
...
+    """
+    return TaskResult.objects.filter(
+        task_name=task_object.name,
+        task_args=str(task_object.request.args),
+        task_kwargs=str(task_object.request.kwargs),
+        date_created__gte=localized_utcnow() - time_delta,
+    ).exclude(status=states.FAILURE).exists()
+
+
+def expiring_task_semaphore(time_delta=None, sentinal=None):
```
iloveagent57: This will eventually get used in other tasks, sorry if it seems like overkill at the moment.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -91,6 +93,48 @@ def _fetch_courses_by_keys(course_keys):
...
+    def decorator(task):
+        @functools.wraps(task)
+        def wrapped_task(self, *args, **kwargs):
+            if task_recently_run(self, time_delta=time_delta or timedelta(hours=1)):
+                msg_args = (self.name, self.request.id, self.request.args, self.request.kwargs)
+                logger.info((
+                    '{} task with id {} was recently run with '
+                    'args: {} kwargs: {}, task returning without updating.'
+                ).format(*msg_args))
+                return sentinal
```
adamstankiewicz: [clarification] What is the role of this `sentinal` argument/variable? I suppose it gives the decorated method visibility into whether the task returned early or if the task was properly kicked off?

adamstankiewicz (approved):
Nice approach with the decorator!

I have one non-blocking clarification question; wondering if it might be worth adding a brief comment what the `sentinal` kwarg is for in a docstring :)

enterprise_catalog/apps/api/tasks.py
iloveagent57: Thanks for pointing this out, I added a doc string and renamed it to be more clear.  And made the unit test more explicit about it, too.

adamstankiewicz: The changes to add the doc string / renaming looks good! 👍

PR #252, opened 2021-02-19
style: update parameter name for clarity

taliaedX:
## Description

The `get_catalog_list` parameter name was causing confusion as we thought it was returning _all_ catalogs regardless of matching content_ids. The parameter name has been updated to `get_catalogs_containing_specified_content_ids`

## Ticket Link

Link to the associated ticket
[Link title](https://openedx.atlassian.net/browse/ENT-3991)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views.py
+++ b/enterprise_catalog/apps/api/v1/views.py
@@ -317,9 +317,15 @@ class EnterpriseCustomerViewSet(BaseViewSet):
               description: Uuids of the programs to check availability of
               paramType: query
             - name: get_catalog_list
+              description: [Old parameter] Return a list of catalogs in which the course / program is present
+              paramType: query
+            - name: get_catalogs_containing_specified_content_ids
               description: Return a list of catalogs in which the course / program is present
               paramType: query
         """
+        get_catalogs_containing_specified_content_ids = request.GET.get(
+            'get_catalogs_containing_specified_content_ids', False
+        )
         get_catalog_list = request.GET.get('get_catalog_list', False)
         course_run_ids = unquote_course_keys(course_run_ids)
 
@@ -330,7 +336,7 @@ class EnterpriseCustomerViewSet(BaseViewSet):
             contains_content_items = catalog.contains_content_keys(course_run_ids + program_uuids)
             if contains_content_items:
                 any_catalog_contains_content_items = True
-                if not get_catalog_list:
+                if not (get_catalogs_containing_specified_content_ids or get_catalog_list):
                     # Break as soon as we find a catalog that contains the specified content
                     break
                 catalogs_that_contain_course.append(catalog.uuid)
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/views.py
```diff
@@ -316,11 +316,13 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
             - name: program_uuids
               description: Uuids of the programs to check availability of
               paramType: query
-            - name: get_catalog_list
+            - name: get_catalogs_containing_specified_content_ids
```
edx-abolger, line 322: I'm confused here: if this is a v1 api, we can't just update parameter names, right? Are these semantically versioned APIs ? Do customers use them? (There is an easy fix here where you just allow either parameter name and it will be backwards compatible, just asking for business reasons)

binodpant: Agreed in general with semantic versioning that this should be api/v2 if we want to rename param. However it may be worth asking Titans if anyone does use it? I don't see any actual usage of it outside learner portal, but you never know

https://github.com/search?q=org%3Aedx+get_catalog_list&type=code

if there are no other consumers then perhaps we can get away with a rename in the same version?

taliaedX: Good call out. I don't know that it is being used publicly. I feel like @iloveagent57 might have the most insight whether it's in use/a big deal to change?

If it does require a version change, I may close this. I don't see a single parameter update for clarification as a reason to go through an API version update.

taliaedX: As discussed post-standup, another option is to include both parameters thus making it backwards compatible. The question is whether the API is used publicly and thus whether having duplicate parameters at this stage is needless clutter or if it is necessary to not break end user usage.

iloveagent57: What's wrong with the existing parameter?  Too long?

I don't believe any customer is currently using this endpoint, you could check if it's mapped in the API manager config to verify.

taliaedX: Well I'm making the current parameter name longer...so that's not the problem.
It was deemed too vague as we thought this might be getting all catalogues rather than those specified with a matching content_id(s).

What is the api manager config? I see an api and api-compact yaml file in this repo that look vaguely mappy, but nothing in the enterprise-customer path is included in those files, so unsure if that means they aren't in the config or if it's a different config I should be looking at.

iloveagent57: yeah, those are the files to look at, they're used by AWS API Manager...somehow.  Anyway, I don't see it referenced in those files, so you're safe to assume this is only used internally/by edx.org.

:) I misread this PR, I agree that the new name is more descriptive.  I think you're fine with the current strategy of adding a new, equivalent parameter and then eventually removing the deprecated one.  Agreed that we don't need a new version of the API to support this change.

edx-abolger (approved):

PR #257, opened 2021-02-22
Exclude the task's own id from the task_recently_run() check. ENT-4018

iloveagent57:
## Description

A `TaskResult` object is written as soon as a task is received (with a status of `RECEIVED`) - which means when we check if a task with the same name/args already exists, we'll find that one does - it's the current task.  This change updates that check to exclude records with the same `task_id` as the current task.

## Ticket Link

[ENT-4081](https://openedx.atlassian.net/browse/ENT-4081)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -76,10 +76,6 @@ def _fetch_courses_by_keys(course_keys):
             continue
         if timezone.now() - content_metadata[0].modified > timedelta(seconds=timeout_seconds):
             courses.append(content_metadata[0].json_metadata)
-            logger.info(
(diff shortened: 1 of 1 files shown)


brittneyexline (approved):

enterprise_catalog/apps/api/tasks.py
```diff
@@ -76,10 +76,6 @@ def _fetch_courses_by_keys(course_keys):
             continue
         if timezone.now() - content_metadata[0].modified > timedelta(seconds=timeout_seconds):
             courses.append(content_metadata[0].json_metadata)
-            logger.info(
-                'ContentMetadata with key %s has recently been updated and will not be requested from Discovery API',
-                key,
-            )
```
iloveagent57, line 82: Removed because it's filling a lot of our logs and we don't really need it any more.

PR #258, opened 2021-02-22
Equivalent recent tasks should fail/revoke.

iloveagent57:
## Description
Tasks for which an equivalent task was recently run now move to the REVOKED status and raise an exception.
This makes the semaphored task invocation more idempotent - a task instance that is locked during some time period    doesn't "count against" future invocations when checking if a recent, successful equivalent task was run.  ENT-4081.

## Ticket Link

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
+++ b/enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
@@ -1,6 +1,6 @@
 import logging
 
-from celery import chord
+from celery import group
 from django.core.management.base import BaseCommand
 
 from enterprise_catalog.apps.api.tasks import (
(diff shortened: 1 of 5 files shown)


enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
```diff
@@ -65,26 +65,28 @@ def handle(self, *args, **options):
...
+        # and associate those with the appropriate CatalogQuery(s).
+        update_group = group(
             [
                 self._run_update_catalog_metadata_task(catalog_query)
                 for catalog_query in catalog_queries
             ]
-        )(self._run_update_full_content_metadata_task())
+        )
+        try:
+            update_group_result = update_group.apply_async().get(
```
adamstankiewicz, line 86: > "Backends use resources to store and transmit results. To ensure that resources are released, you must eventually call get() or forget() on EVERY AsyncResult instance returned after calling a task."
Reference: https://docs.celeryproject.org/en/stable/reference/celery.result.html#celery.result.AsyncResult.get

[clarification] Just to confirm/learn, we're able to call `.get()` just once here because `update_group` is returning only 1 AsyncResult instance after being called?

iloveagent57: Yeah, my reading is that a `GroupResult` operates on all of the `AsyncResults` of the group when you call `.get()`. https://docs.celeryproject.org/en/stable/userguide/canvas.html#group-results

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
```diff
@@ -64,27 +64,42 @@ def handle(self, *args, **options):
...
-        # callback to update the json_metadata of ContentMetadata records with content type "course"
-        # with the full course metadata from /courses/.
-        update_chord_task = chord(
+        # First, we create a group of celery tasks that run in parallel to create/update ContentMetadata records
+        # and associate those with the appropriate CatalogQuery(s).
+        # It's possible that one of the tasks in the group will fail with a TaskRecentlyRunError.
+        # Note that a failed task in a group does not stop the rest of the tasks in the group from running.
+        # We consider this error innocuous, and don't want an occurrence(s) of it to prevent
+        # update_full_content_metadata_task from being run.
+        # Thus we run the update_catalog_metadat_tasks in their own group, wait
```
adamstankiewicz: nit: typos

 "update_catalog_metadat_tasks" -> "update_catalog_metadata_tasks"
"updaet_full_content_metadata_task" -> "update_full_content_metadata_task"

adamstankiewicz (approved):
LGTM with a nit for some typos in the comments :)

PR #261, opened 2021-02-23
Refactor update_full_content_metadata

iloveagent57:
## Description

Refactors update_full_content_metadata_task to require no arguments and work on all recently updated (within 1 hour)
ContentMetadata records.  This task is now locked based on equivalent, recently executed invocations.  It also
now retries if there are update_catalog_metadata_tasks in unready states within the past hour. ENT-4081

## Testing Instructions

* Run `./manage.py update_{full_}content_metadata` from an enterprise-catalog app shell.
* The first invocation should do some actual work.  Subsequent invocations within the same hour should emit a message about the task(s) recently having been run.
* Test the `refresh_metadata` endpoint - make a POST request to http://localhost:18160/api/v1/enterprise-catalogs/{catalog_uuid}/refresh_metadata with a body of `{"uuid": "{catalog_uuid}"}`.
* This should return a 200/OK response with a body that gives you an async task ID.
* Use the `django_celery_results_taskresult` table in `mysql-shell` to see that new tasks were started just for this catalog's query, including a task to update algolia.
* Make the request again, and you should see a REVOKED task, because we ran too recently.
* To test the retry logic - add a temporary `sleep(60)` call toward the end of the `update_catalog_metadata_task` - now submit your POST request again.  In an `app-shell`, run `./manage.py update_full_content_metadata` - this should be blocked and you should see a retry message in the worker logs - in 5 minutes the retry should occur and return successfully.

## Ticket Link

## Post-review


enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
```diff
@@ -98,9 +97,26 @@ def handle(self, *args, **options):
...
+                full_update_result
+            ))
+        except Exception as exc:
+            if type(exc).__name__ != 'TaskRecentlyRunError':
+                raise
+            else:
+                logger.info(
+                    'update_full_content_metadata_task was recently run prior to this command, '
+                    'and was thus skipped during the execution of this command.'
+                )
```
chavesj: [nit] I've typically seen exceptions handled with the following syntax - is it necessary to use the dunder here?
`
except TaskRecentlyRunError:
    logger.info(...)
except:
    raise
`

enterprise_catalog/apps/catalog/management/commands/update_full_content_metadata.py
```diff
@@ -17,10 +17,18 @@ class Command(BaseCommand):
...
-        )
-        logger.info(message, result)
+        try:
+            result = update_full_content_metadata_task.apply_async().get()
+            message = (
+                'update_full_content_metadata task from update_full_content_metadata command finished'
+                ' successfully with result %s'
+            )
+            logger.info(message, result)
+        except Exception as exc:
```
chavesj, line 27: [nit] same question about exception handling syntax

chavesj (commented):
one nit about exception handling, besides that LGTM once quality checks pass

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
iloveagent57: Yeah, that is definitely the preferred way to do it :). There's a comment on line 90 above about a weird thing celery does that forces us to write it in this way, I'll put a "pointer comment" near this invocation, too.

enterprise_catalog/apps/catalog/management/commands/update_full_content_metadata.py
iloveagent57: Same as other reply

chavesj (approved):
🚀 🌙

adamstankiewicz (approved):
LGTM!

PR #262, opened 2021-02-24
fix: use image_url for algolia course indexing

adamstankiewicz:
## Description

Use `image_url` field in course content metadata during Algolia indexing now that this field exists and seems to have stable data (i.e., all course content in enterprise-catalog has an `image_url` field), which IIRC wasn't the case before.

**User impact: better performance when loading course search results due to loading smaller file sizes.**

The learner portal currently uses whatever image url is indexed in Algolia.

_Example:_
In this example, we can see a ~67.8% decrease in image file size by using the smaller image: 12.65 kB vs. 39.1 kB. Multiply these savings over 24 course cards visible at a time, and there's the potential for reducing the image payload for **each** page in the search results by over half a mB (634.8 kB). Note this is solely based on the example file sizes for the images shown below.

```
{
  "image_url": "https://prod-discovery.edx-cdn.org/media/course/image/5236b816-deac-4a69-9586-96ca0f4dad2a-ccead5306d4c.small.jpg",
}
```

vs.

```
"original_image": {
  "src": "https://prod-discovery.edx-cdn.org/media/course/image/e56e7660-bc94-46fd-95c5-ae3f765da60c-3c84e255cf9e.jpg",
  // ...
},
```

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tests/test_tasks.py
+++ b/enterprise_catalog/apps/api/tests/test_tasks.py
@@ -173,14 +173,14 @@ class UpdateFullContentMetadataTaskTests(TestCase):
             'uuid': metadata_1.json_metadata.get('uuid'),
             'aggregation_key': 'course:fakeX',
             'marketing_url': metadata_1.json_metadata.get('marketing_url'),
-            'original_image': metadata_1.json_metadata.get('original_image'),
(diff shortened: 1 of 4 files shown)


iloveagent57 (approved):
🎉

PR #263, opened 2021-02-24
Algolia - New Search Field for Catalog Query

taliaedX:
## Description

- Add new created and modified fields via changing CatalogQuery to Timestamped model
- Added 3 migrations: uuid field, nullable > iterate through and add to all existing catalog queries (~1800 records in prod)  > require field to be unique
- Add query uuid to algolia index task
- Update algolia attributes to include new field.

Screen Shot 2021-02-24 at 4 24 14 PM

## Ticket Link

Link to the associated ticket
[ENT-4152](https://openedx.atlassian.net/browse/ENT-4152)

## Post-review


enterprise_catalog/apps/catalog/migrations/0023_catalogquery_generate_uuids.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 2.2.18 on 2021-02-24 20:56
+
+from django.db import migrations
+import uuid
+
+def generate_uuid(apps, schema_editor):
```
adamstankiewicz, line 6: [clarification] Was it necessary to split this change into 3 separate migrations?

If we're adding `uuid` field with a default of `uuid4`, would we expect it to automagically generate uuids for us? Perhaps that's not the case when backfilling these data, but I'm not certain.

If that is the case, I feel like this could all be done in a single migration? 🤔

enterprise_catalog/apps/catalog/models.py
```diff
@@ -67,12 +67,18 @@ class CatalogQuery(models.Model):
         editable=False,
     )
 
+    uuid = models.UUIDField(
```
adamstankiewicz, line 71: Were we planning to make `uuid` the primary key for this table? Or, are we intentionally keeping `id` as the primary key?

enterprise_catalog/apps/catalog/migrations/0023_catalogquery_generate_uuids.py
taliaedX: the default uuid value actually only generates a single uuid (lol) which then throws non-unique constraint errors because of course!

There might be a way otherwise, but this 3 migration style was actually the recommendation in Django's docs.

adamstankiewicz: Got it!

enterprise_catalog/apps/catalog/models.py
taliaedX: We were discussing this and were not sure if it was necessary for it to be the primary key for the use case. Making it primary key* was throwing some errors at me (it was something with admin forms oddly enough). We also currently have a foreign key relationship between catalog query id and catalog that would need to be updated. It's certainly possible to do but would involve a few more migrations and careful assessment of all data relationships that may be affected.

So TL;DR totally possible, but we weren't sure it was necessary for the use case and if not, it's more of a hassle than it's worth (perhaps this is how all the other tables that have both id and uuid ended up as such 😅 )

jlajoie (approved):
Looks good 👍 , not sure if there's extra work in Algolia for faceting outside of that attributesForFaceting section

taliaedX:
> not sure if there's extra work in Algolia for faceting outside of that attributesForFaceting section

Sorry for the delay. @jlajoie what do you mean by this? What facets are we looking for/expecting?

PR #265, opened 2021-02-25
Refactor the reindex algolia task.

iloveagent57:
## Description

Refactor the reindex algolia task to require no arguments and work on all recently updated (within 1 hour)
ContentMetadata records.  This task is now locked based on equivalent, recently executed invocations.
It now retries if there are update_full_content_metadata_tasks in unready states within the past hour. ENT-4081

## Testing Instructions

* Temporarily comment out line 326 and 405 of tasks.py - they're the two lines that interact with the Algolia API. 
* Run `./manage.py `reindex_algolia` from an enterprise-catalog app shell.
* The first invocation should do some actual work.  Subsequent invocations within the same hour should emit a message about the task(s) recently having been run.
* Test the `refresh_metadata` endpoint - make a POST request to http://localhost:18160/api/v1/enterprise-catalogs/{catalog_uuid}/refresh_metadata with a body of `{"uuid": "{catalog_uuid}"}`.
* This should return a 200/OK response with a body that gives you an async task ID.
* Use the `django_celery_results_taskresult` table in `mysql-shell` to see that new tasks were started just for this catalog's query, including a task to update algolia.
* Make the request again, and you should see a REVOKED task, because we ran too recently.
* Delete all of the `django_celery_results_taskresult` records.
* To test the retry logic - add a temporary `sleep(60)` call toward the end of the `update_full_content_metadata_task` - now submit your POST request again.  In an `app-shell`, run `./manage.py reindex_algolia` - this should be blocked and you should see a retry message in the worker logs - in 5 minutes the retry should occur and return successfully.

## Post-review


enterprise_catalog/apps/api/tasks.py
```diff
@@ -288,36 +290,40 @@ def _batched_metadata(json_metadata, sorted_uuids, uuid_key_name, obj_id_fmt, uu
...
     exception.
+    """
+    content_keys = ContentMetadata.recently_modified_records(
+        ONE_HOUR
+    ).filter(
+        content_type=COURSE
+    ).values_list(
+        'content_key',
+        flat=True
+    )
```
adamstankiewicz, line 316: tiny nit: As `content_keys` isn't used prior to the `unready_tasks` check below, would it make sense to move this query below that check so it only needs to run when appropriate?

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -233,11 +233,10 @@ def test_update_full_metadata(self, mock_oauth_client, mock_get_indexable_course
         assert metadata_1.json_metadata != course_data_1
         assert metadata_2.json_metadata != course_data_2
 
-        mock_get_indexable_course_keys.return_value = [course_key_1, course_key_2]
-        indexable_course_keys = tasks.update_full_content_metadata_task.apply().get()
-        # Verify the task returns the course keys that are indexable as those keys are passed to the
-        # `index_enterprise_catalog_courses_in_algolia_task` from the `EnterpriseCatalogRefreshDataFromDiscovery` view.
-        assert indexable_course_keys == mock_get_indexable_course_keys.return_value
+        tasks.update_full_content_metadata_task.apply().get()
```
adamstankiewicz, line 236: [clarification] now that most of our tasks are applied with `apply_async`, is there any tradeoff/benefit not using it here in the test? Or does it not really matter within a test, except if we're mocking an existing `apply_async` , e.g.: `mock_task.apply_async.return_value.get.assert_called_once_with()`?

adamstankiewicz (commented):
Looks good, just a nit and a clarification question on a first pass. I will come back for a 👍 after I grab a coffee ☕ 😉

enterprise_catalog/apps/api/tasks.py
iloveagent57: 👍 great point, fixing in the other task, too.

enterprise_catalog/apps/api/tests/test_tasks.py
iloveagent57: Good question.  The only benefit is that, if you forget to put `--ds=enterprise_catalog.settings.test` in a manual `pytest` invocation from an app shell, `apply_async` will try to use the broker/worker.  The test settings set `CELERY_ALWAYS_EAGER` to True, it's False in our devstack settings, which are the default settings in the shell environment.  I'm torn on whether to change this test invocation.

adamstankiewicz: Hm, now I'm also torn haha! On the one hand it'd be great to call our tasks in tests the same way they get invoked in a non-test environment, but on the other hand, we don't want running `pytest` locally to use the broker/worker.

I guess for the purpose of this test, we are only trying to execute a task and is probably acceptable that it gets invoked as a synchronous blocking task with the test itself 🤷‍♂️

adamstankiewicz (approved):
LGTM. I'll leave the question of `apply` vs. `apply_async` in the test up to your best judgement :)

PR #266, opened 2021-02-26
Remove isort_check from quality/validation

iloveagent57:
...because its costs outweigh its benefits.

## Description

Having our imports sorted just right provides very minimal value, and has a high cumulative cost in PR/CI churn:

1. `pylint` already catches unused imports.
2. Generally, having imports at the top of the file and grouped according to standard python convention is something most developers get right most of the time.
3. ...and if they don't, there's very little impact.  Generally, a developer inspects import in a file by using the search function of their IDE, not via a manual eyeball scan.
4. Having properly sorted imports provides 0 business value...
5. ...it doesn't eliminate any risk.
6. ...it doesn't improve developer productivity or velocity (I'd argue it harms velocity).
7. ...it has negligible impact on code readability, given point (2) above.
8. I can't think of a single incident/bug/defect that would have been caught, if only we were running `isort` (isort has only been a tool used by edX since 2018 or so).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,14 +2,14 @@
 TOX = ''
 .PHONY: help clean piptools requirements dev_requirements \
         doc_requirementsprod_requirements static shell test coverage \
-        isort_check isort style lint quality pii_check validate \
+        style lint quality pii_check validate \
         migrate html_coverage upgrade extract_translation dummy_translations \
         compile_translations fake_translations  pull_translations \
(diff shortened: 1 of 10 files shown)


chavesj:
[trying to find best-of-both-worlds scenario] I agree that the sorting of imports shouldn't impact productivity / velocity. Would you consider using the `isort` Makefile rule in a pre-commit hook instead?

PR #267, opened 2021-02-26
feat: adds more useful info to the distinct-catalog-queries endpoint …

chavesj:
…response

## Description

* Returns a list of EnterpriseCustomerCatalog UUIDs in a dictionary, keying on associated CatalogQuery ID
* Adds commenting

## Ticket Link

[ENT-4039](https://openedx.atlassian.net/browse/ENT-4039)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views.py
+++ b/enterprise_catalog/apps/api/v1/views.py
@@ -356,14 +356,28 @@ class DistinctCatalogQueriesView(APIView):
         """
         Method to handle POST requests to this endpoint
         """
+        # Find all catalogs that match the UUIDs sent in the request data
         enterprise_catalog_uuids = request.data.get('enterprise_catalog_uuids', [])
+        enterprise_catalogs = EnterpriseCatalog.objects.filter(uuid__in=enterprise_catalog_uuids)
 
-        distinct_catalog_query_ids = EnterpriseCatalog.objects.filter(
-            uuid__in=enterprise_catalog_uuids,
-        ).distinct().values_list('catalog_query__id', flat=True)
+        # Iterate through catalogs and populate to the catalog query map
+        # Format:
+        #   key: CatalogQuery ID (int)
+        #   value: EnterpriseCustomerCatalog UUID(s) (list[str])
+        catalog_query_map = {}
+        for catalog in enterprise_catalogs:
+            query_id = catalog.catalog_query_id
+            if query_id in catalog_query_map:
+                catalog_query_map[query_id].append(str(catalog.uuid))
+            else:
+                catalog_query_map[query_id] = [str(catalog.uuid)]
 
+        # Return catalog_query_map in the response.
+        # In the case that the catalog query check fails,
+        # the map can be used to determine which catalogs
+        # map to incorrect queries.
         response_data = {
-            'count': len(distinct_catalog_query_ids),
-            'catalog_query_ids': distinct_catalog_query_ids,
+            'count': len(catalog_query_map.keys()),
+            'catalog_query_ids': catalog_query_map,
         }
         return Response(response_data, status=HTTP_200_OK)
--- a/enterprise_catalog/apps/api/v1/views.py
+++ b/enterprise_catalog/apps/api/v1/views.py
@@ -1,4 +1,4 @@
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 
 import crum
 from celery import chain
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/views.py
```diff
@@ -356,14 +356,28 @@ def post(self, request):
...
-            uuid__in=enterprise_catalog_uuids,
-        ).distinct().values_list('catalog_query__id', flat=True)
-
+        enterprise_catalogs = EnterpriseCatalog.objects.filter(uuid__in=enterprise_catalog_uuids)
+
+        # Iterate through catalogs and populate to the catalog query map
+        # Format:
+        #   key: CatalogQuery ID (int)
+        #   value: EnterpriseCustomerCatalog UUID(s) (list[str])
+        catalog_query_map = {}
```
iloveagent57: suggestion: `collections.defaultdict` is pretty great, e.g.
```python
catalogs_by_query_id = defaultdict(list)
for catalog in enterprise_catalogs:
    catalogs_by_query_id[catalog.catalog_query_id].append(str(catalog.uuid))
```

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -356,14 +356,28 @@ def post(self, request):
...
+
+        # Return catalog_query_map in the response.
+        # In the case that the catalog query check fails,
+        # the map can be used to determine which catalogs
+        # map to incorrect queries.
         response_data = {
-            'count': len(distinct_catalog_query_ids),
-            'catalog_query_ids': distinct_catalog_query_ids,
+            'count': len(catalog_query_map.keys()),
+            'catalog_query_ids': catalog_query_map,
```
iloveagent57: Should probably update the docstring for this view to reflect the change in structure of the response data.

iloveagent57: Also consider changing the `count` key to `number_distinct_query_ids` or something like that.  Or maybe eschew it , since you can figure out the count by looking at the length of `catalog_query_ids`.

Also, `catalog_query_ids` is now a bit of a misnomer, maybe `catalog_uuids_by_catalog_query_ids`, or `query_ids_to_catalog_uuids`.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -356,14 +356,28 @@ def post(self, request):
         """
         Method to handle POST requests to this endpoint
         """
+        # Find all catalogs that match the UUIDs sent in the request data
```
iloveagent57: nit: You could just move this to the method docstring and put "Args:\n" in front of it :)

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -356,14 +356,28 @@ def post(self, request):
...
-        distinct_catalog_query_ids = EnterpriseCatalog.objects.filter(
-            uuid__in=enterprise_catalog_uuids,
-        ).distinct().values_list('catalog_query__id', flat=True)
-
+        enterprise_catalogs = EnterpriseCatalog.objects.filter(uuid__in=enterprise_catalog_uuids)
+
+        # Iterate through catalogs and populate to the catalog query map
+        # Format:
+        #   key: CatalogQuery ID (int)
+        #   value: EnterpriseCustomerCatalog UUID(s) (list[str])
```
iloveagent57: nit: consider removing, it's repeating what the code does, and the code is very clear.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -356,14 +356,28 @@ def post(self, request):
...
+            query_id = catalog.catalog_query_id
+            if query_id in catalog_query_map:
+                catalog_query_map[query_id].append(str(catalog.uuid))
+            else:
+                catalog_query_map[query_id] = [str(catalog.uuid)]
+
+        # Return catalog_query_map in the response.
+        # In the case that the catalog query check fails,
+        # the map can be used to determine which catalogs
+        # map to incorrect queries.
```
iloveagent57: nit: this is informative enough to be part of the view's docstring.

iloveagent57 (commented):
A bunch of minor points.  Also, this endpoint is now kind of weird - if I give it one catalog id, I can find out about all of the other catalog ids that use the same catalog query as the catalog id I provided as input (I know it's an Admin-only endpoint, so probably not a huge concern).

iloveagent57 (approved):

PR #270, opened 2021-03-01
Avoid algolia task doomspiral

iloveagent57:
Do not auto-retry tasks that failed due to timeout, especially for ones that failed due to a soft timeout.  This can lead to a doom-loop of costly, repetitive work.

Background:
I noticed in the logs that the same algolia reindex task was retried 5 times earlier this afternoon, and it did the same set of work each time.  What happens is something like this:
* Task instance 1 does a lot of hard work.  It does it for 8 minutes, which triggers a soft timeout error.  We retry automatically on soft timeout errors.  task instance 1 keeps running, but a retry is queued up.
* 1 minute has passed, instance 1 is done by now.
* task instance 2 starts, it does a lot of hard work.  It does it for 8 minutes, which triggers a soft timeout error.  We retry automatically on soft timeout errors.  task instance 2 keeps running, but a retry is queued up.
* (...you get the idea...)
* We've hit our max number of retries (5) and stop queueing up more work.

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -5,8 +5,7 @@ from collections import defaultdict
 from datetime import timedelta
 
 from celery import shared_task, states
-from celery.exceptions import Ignore, SoftTimeLimitExceeded
-from celery.exceptions import TimeoutError as CeleryTimeoutError
+from celery.exceptions import Ignore
 from celery_utils.logged_task import LoggedTask
 from django.db import IntegrityError
 from django.db.models import Prefetch, Q
@@ -175,8 +174,6 @@ class LoggedTaskWithRetry(LoggedTask):  # pylint: disable=abstract-method
     """
     autoretry_for = (
         IntegrityError,
-        SoftTimeLimitExceeded,
-        CeleryTimeoutError,
         OperationalError,
     )
     retry_kwargs = {'max_retries': 5}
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -175,8 +174,6 @@ class LoggedTaskWithRetry(LoggedTask):  # pylint: disable=abstract-method
     """
     autoretry_for = (
         IntegrityError,
-        SoftTimeLimitExceeded,
-        CeleryTimeoutError,
```
adamstankiewicz, line 179: [clarification] just to confirm, do we want to stop auto-retry for `SoftTimeLimitExceeded` and `CeleryTimeoutError` for the other tasks, or are we OK with those tasks also no longer retrying when time limits are reached?

iloveagent57: I think we want to stop for other tasks, too.  Retrying tasks for task timeouts feels like a bad pattern, now that I've stopped to think about it.

adamstankiewicz: Got it, cool. Is the reasoning here that if a task is already timing out, it's ~unlikely it'll _not_~ likely to timeout on a successive retry?

iloveagent57: Yeah, exactly.

(Unless we wrote our tasks in a more idempotent way - then this pattern might make sense.)

adamstankiewicz (approved):

PR #271, opened 2021-03-01
Mark recently reindexed course metadata records

iloveagent57:
...so that subsequent batches of algolia reindexing don't reindex them a second time.

This is a minor but hopefully valuable optimization.  Because our algolia task looks for metadata records to update in batches, _and_ because it looks for records based on key **or** parent key, it's possible that we'll update the same course record in algolia multiple times across multiple batches within a single task.

The solution here is to mark recently-reindexed course metadata records in the cache and skip over them if they're present.

Why use the cache instead of a `set()` called `recently_updated_course_records` within the task body?
Answer: this is a little of a belt-and-suspenders approach.  In case we get into a situation where more task invocations are happening than we'd like, this is a sort-of failback, as long as those extra task invocations occur within the cache expiry period - which is 30 minutes in this PR.

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -8,6 +8,7 @@ from celery import shared_task, states
 from celery.exceptions import Ignore, SoftTimeLimitExceeded
 from celery.exceptions import TimeoutError as CeleryTimeoutError
 from celery_utils.logged_task import LoggedTask
+from django.core.cache import cache
 from django.db import IntegrityError
 from django.db.models import Prefetch, Q
 from django.db.utils import OperationalError
@@ -324,6 +325,7 @@ def _reindex_algolia(content_keys):
     algolia_client = get_initialized_algolia_client()
 
     # Update the index in batches
+    logger.info('There are {} total content keys to update in the Algolia index.'.format(len(content_keys)))
     for content_keys_batch in batch(content_keys, batch_size=TASK_BATCH_SIZE):
         courses = []
         catalog_uuids_by_course_key = defaultdict(set)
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -372,6 +374,9 @@ def _reindex_algolia(content_keys):
         course_content_metadata = content_metadata.filter(content_type=COURSE)
         for metadata in course_content_metadata:
             content_key = metadata.content_key
+            if _was_recently_indexed(content_key):
```
adamstankiewicz, line 377: nit: Would it make sense to filter out / exclude recently indexed content keys before we iterated through all `course_content_metadata`? I guess we'd be iterating through `course_content_metadata` twice at that point, though so maybe a moot point 🤷‍♂️

adamstankiewicz (approved):
I think the approach of using the django cache here is great! I agree that using the cache feels a bit safer than using a local `set()` within the task itself to handle the case you outlined in the PR description.

enterprise_catalog/apps/api/tasks.py
iloveagent57: I like to make a kind of "cache sandwich" out of the code, so it's easier to tell what's going on.

cache sandwich, serves 1:
Top slice of bread ~= check if something's in the cache, maybe early return
Meat/veggies ~= do the work
Bottom slice of bread ~= put something in the cache

adamstankiewicz: Yeah, having the check whether something's in the cache nested within the iteration probably does help with readability / general understanding.

Also, side note: In a world with cached sandwiches, if something is already in the cache, is your sandwich literally just a slice of bread? 🍞 🥖

PR #275, opened 2021-03-12
feat: add language name to Algolia index course records

adamstankiewicz:
## Description

* Parses the `content_language` field from the advertised course run into a human-readable string (e.g., "en-us" to "English"), which gets added to the Algoia-indexed course records. This is done through the `langcodes` package which was recently removed since it was previously throwing an error (see [ENT-4163](https://openedx.atlassian.net/browse/ENT-4163) for details). However, this error was happening because the necessary `language_data` package wasn't installed previously. With `language_data` installed, the previous error should no longer occur but we are still catching an exception should a `LanguageTagError` occur.
* Increases test coverage for algolia_utils.py

## Ticket Link

[ENT-2990](https://openedx.atlassian.net/browse/ENT-2990)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -195,12 +195,12 @@ def get_course_availability(course):
         'upcoming': 'Upcoming',
     }
     course_runs = course.get('course_runs') or []
-    active_course_runs = filter(_is_course_run_active, course_runs)
+    active_course_runs = [run for run in course_runs if _is_course_run_active(run)]
     availability = set()
     for course_run in active_course_runs:
-        run_availability = course_run.get('availability', '').lower()
+        run_availability = course_run.get('availability') or ''
         availability.add(
-            COURSE_AVAILABILITY_MESSAGES.get(run_availability, DEFAULT_COURSE_AVAILABILITY)
+            COURSE_AVAILABILITY_MESSAGES.get(run_availability.lower(), DEFAULT_COURSE_AVAILABILITY)
         )
     return list(availability)
 
@@ -343,7 +343,8 @@ def _is_course_run_active(course_run):
     Returns:
         bool: Whether the specified course run is "active" (i.e., published, enrollable, marketable)
     """
-    is_published = course_run.get('status', '').lower() == 'published'
+    course_run_status = course_run.get('status') or ''
+    is_published = course_run_status.lower() == 'published'
     is_enrollable = course_run.get('is_enrollable', False)
     is_marketable = course_run.get('is_marketable', False)
 
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -1,7 +1,13 @@
 import copy
+import logging
+
+from langcodes import Language
+from langcodes.tag_parser import LanguageTagError
 
 from enterprise_catalog.apps.api_client.algolia import AlgoliaSearchClient
 
+
+logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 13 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -91,7 +99,7 @@ def _should_index_course(course_metadata):
     if advertised_course_run is None:
         return False
 
-    owners = course_json_metadata.get('owners', [])
+    owners = course_json_metadata.get('owners') or []
```
kellmarie, line 102: Does leaving it how it originally was not work?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -134,12 +145,46 @@ def get_algolia_object_id(uuid):
...
+
+    parsed_language_name = None
+    try:
+        language = Language.get(content_language)
+        if language.is_valid():
+            parsed_language_name = language.language_name()
+    except LanguageTagError:
+        course_run_id = advertised_course_run.get('key')
+        logger.exception(
+            'Could not parse content_language {content_language} for course run {course_run_id}'.format(
```
iloveagent57: nit: maybe `{content_language!r}` to get the rep of the language, in case it's an empty string or spaces or something like that (and since Ned just posted about this trick recently).  e.g.
```
>>> x = '    '
>>> print('this is the string {x!r}'.format(x=x))
this is the string '    '
>>> print('this is the string {x}'.format(x=x))
this is the string
```

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -149,15 +194,14 @@ def get_course_availability(course_runs):
         'current': 'Available Now',
         'upcoming': 'Upcoming',
     }
-
+    course_runs = course.get('course_runs') or []
+    active_course_runs = filter(_is_course_run_active, course_runs)
```
iloveagent57: Nit: use a list comprehension "because it's more pythonic" e.g.
```python
active_course_runs = [run for run in course_runs if _is_course_run_active(run)]
```

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -289,6 +333,23 @@ def get_advertised_course_run(course):
...
+    """
+    Determines whether a course run is "active" based on whether the run in published, enrollable, and marketable.
+
+    Arguments:
+        course_run (dict): a dict representing a single course run
+
+    Returns:
+        bool: Whether the specified course run is "active" (i.e., published, enrollable, marketable)
+    """
+    is_published = course_run.get('status', '').lower() == 'published'
```
iloveagent57: Nit: you might run into the same type of problem here that you coded around earlier, that `course_run['status']` is set to `None`, and then `None.lower()` will raise an exception.

iloveagent57 (approved):
Couple of nits, looks good otherwise.

enterprise_catalog/apps/catalog/algolia_utils.py
adamstankiewicz:
```python
course_json_metadata = {'owners': None}
owners = course_json_metadata.get('owners', [])
```

In the above example, `owners` will actually be `None` instead of an empty list. `.get('owners', [])` will only fall back to an empty list when `owners` is not part of the course json metadata to begin with.

By moving the empty list fall back as an `or []`, we can still default to an empty list in this situation.

enterprise_catalog/apps/catalog/algolia_utils.py
adamstankiewicz: Agreed this may be useful :) Hooray for applying learnings from others/Slack haha.

PR #278, opened 2021-03-25
remove langcodes in favor of reading parsed language name from course metadata

adamstankiewicz:
## Description

Due to the format of the Chinese languages on edx.org/search, the `langcodes` approach to parse language codes into human-readable language names did not match exactly with the `LanguageTag` model data stored in the course-discovery database. For example, "Chinese - Mandarin" (edx.org) vs. "Mandarin Chinese" (langcodes), which throws off the A-Z sorting/grouping of languages.

Instead, the formatted language name is added to the `CourseRunSerializer` in course-discovery such that enterprise-catalog gets the name "for free" via the `/courses/` API endpoint without needing to parse a language code or duplicate the `LanguageTag` model from course-discovery.

See https://github.com/edx/course-discovery/pull/2990.

## Ticket Link

[ENT-2990](https://openedx.atlassian.net/browse/ENT-2990)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/constants.py
+++ b/enterprise_catalog/apps/catalog/constants.py
@@ -1,10 +1,9 @@
 import json
 
-import waffle
+from enterprise_catalog.apps.catalog.waffle import (
+    DISABLE_MODEL_ADMIN_CHANGES_SWITCH,
+)
 
-
-# Waffle Switches
-DISABLE_MODEL_ADMIN_CHANGES = 'disable_model_admin_changes'
 
 # ContentMetadata content_type choices
 COURSE = 'course'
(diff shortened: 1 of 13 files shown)


enterprise_catalog/apps/catalog/constants.py
```diff
@@ -1,11 +1,10 @@
 import json
 
-import waffle
+from enterprise_catalog.apps.catalog.waffle import (
```
adamstankiewicz, line 3: Upgrading to `edx-lint` v5 throws a pylint error on `import waffle`, pushing the use of `edx_toggles` instead.

iloveagent57 (approved):

PR #279, opened 2021-03-25
chore: Removing Decentralized Devstack prototype

jinder1s:
We are no longer going forward with decentralized devstack. Instead, we
have opted for making improvements to devstack itself. Wider announce of
this decision will be made shortly. This PR does some of the necessary
cleanup of the prototype.

https://openedx.atlassian.net/browse/ARCHBOM-1722


docs/decisions/0003-deprecating-decentralized-devstack.rst
```diff
@@ -0,0 +1,24 @@
...
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+A while back, edX considered replacing `Devstack`_ with Decentralized Devstack, a framework for devstack where each IDA would have and own its own devstack. To test out the new framework, a prototype was created for enterprise-catalog. After some user testing and some ideation, it was decided that edX would not try to move towads Decentralized Devstack and instead would opt to do improvements on `Devstack`_.
```
robrap: 1. “A while back...” can be replaced with something like “In this earlier ADR for adding...”.
2. Remove extra “and own”
3. Consider replacing “edX“  with “we”.
4. Maybe we could add a little more detail about what didn’t work well with the idea and how issues that it was meant to address are now being addressed through the improvements.

feanil (approved):

feanil:
Made the updates that @robrap suggested and this looks good to me other than that so approving.

PR #280, opened 2021-03-31
fix: clean up Algolia index when content keys should no longer be indexed

adamstankiewicz:
## Description

While QA'ing the new language facet, it was observed that when a course is no longer published, enrollable, or marketable, it still remains in the Algolia index and, as a result, appears in the Learner Portal search results resulting in a poor experience where a learner can't actually enroll in the course.

This PR ensures we're only adding/updating records for courses that _should_ be in the index, and deleting any records for courses that should _not_ be in the index anymore (e.g., it's no longer published or enrollable).

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -18,6 +18,7 @@ from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.catalog.algolia_utils import (
     ALGOLIA_FIELDS,
     ALGOLIA_UUID_BATCH_SIZE,
+    configure_algolia_index,
     create_algolia_objects_from_courses,
     get_algolia_object_id,
     get_indexable_course_keys,
@@ -271,7 +272,8 @@ def _update_full_content_metadata(content_keys):
         )
 
         # record the course keys that were updated and should be indexed in Algolia by the B2C logic
-        indexable_course_keys.extend(get_indexable_course_keys(modified_content_metadata_records))
+        indexable_course_keys, __ = get_indexable_course_keys(modified_content_metadata_records)
+        indexable_course_keys.extend(indexable_course_keys)
 
     logger.info(
         '{} total course keys were updated and are ready for indexing in Algolia'.format(len(indexable_course_keys))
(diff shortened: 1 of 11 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -420,7 +424,51 @@ def _reindex_algolia(content_keys):
...
+        'Algolia index.'.format(len(content_keys))
+    )
+    for content_keys_batch in batch(content_keys, batch_size=TASK_BATCH_SIZE):
+        filters = set()
+        for content_key in content_keys_batch:
+            filters.add(f'key:"{content_key}"')
+
+        if filters:
+            content_keys_filter_query = 'OR '.join(list(filters))
+            algolia_client.delete_by({'filters': content_keys_filter_query})
```
iloveagent57: Nit: I think it'd be nice to "abstract away" the structure of the filter that the algolia client needs to delete entries from this function.  So something like
```python
for content_keys_batch in batch(content_keys, batch_size=TASK_BATCH_SIZE):
    algolia_client.delete_content_keys(content_keys_batch)
```
...and then have `delete_content_keys()` deal with building up a filter that Algolia can understand.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -306,25 +308,27 @@ def index_enterprise_catalog_courses_in_algolia_task(self):
...
-    content_keys = ContentMetadata.recently_modified_records(
-        ONE_HOUR * 2
-    ).filter(
-        content_type=COURSE
-    ).values_list(
-        'content_key',
-        flat=True
+    recently_modified_records = ContentMetadata.recently_modified_records(ONE_HOUR * 2)
+    courses_content_metadata = recently_modified_records.filter(content_type=COURSE)
+    indexable_content_keys, nonindexable_content_keys = get_indexable_course_keys(courses_content_metadata)
```
iloveagent57: Nit about naming: it reads a little funny to have `get_indexable_course_keys` return both indexable and not-indexable course keys.  Maybe something like `partition_course_keys_for_indexing()`?
(This is not a hill I'm willing to die on, FWIW)

enterprise_catalog/apps/api_client/algolia.py
```diff
@@ -44,11 +46,10 @@ def init_index(self):
...
             self.algolia_index = self._client.init_index(self.ALGOLIA_INDEX_NAME)
-        except Exception as exc:  # pylint: disable=broad-except
-            logger.error(
-                'Could not initialize %s index in Algolia: %s',
+        except AlgoliaException:
+            logger.exception(
+                'Could not initialize %s index in Algolia due to an exception.',
                 self.ALGOLIA_INDEX_NAME,
-                exc,
             )
```
iloveagent57: I think we should `raise` inside these `AlgoliaException` blocks, so that the tasks that run them don't end up being marked with a status of `success`.

enterprise_catalog/settings/base.py
```diff
@@ -354,9 +354,9 @@
 
 # Algolia
 ALGOLIA = {
-    'INDEX_NAME': '',
-    'APPLICATION_ID': '',
-    'API_KEY': '',
+    'INDEX_NAME': 'enterprise_catalog_astankiewicz',
+    'APPLICATION_ID': 'testing2MLYZ9WWU0',
+    'API_KEY': '85cb5d95d235fe5a3120d1c6a3e21766',
```
iloveagent57: no

iloveagent57 (changes requested):
Looks great, couple of suggestions.

enterprise_catalog/apps/api_client/algolia.py
adamstankiewicz: Ah, yes. Good catch, we'll definitely want to `raise` here.

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: Ha, when I was refactoring this function, I had a similar thought but never came back to update. Totally agree it reads a little funny.

enterprise_catalog/apps/api_client/algolia.py
```diff
@@ -98,14 +97,46 @@ def partially_update_index(self, algolia_objects):
...
+            return
+
+        if not content_keys:
+            # nothing left to do here
+            return
+
+        filters = set()
+        for key in content_keys:
+            filters.add(f'key:"{key}"')
+        content_keys_filter_query = 'OR '.join(list(filters))
```
iloveagent57: is there any way that this goes wrong with the colons that are in most of our course keys?  I'd guess not because of the quotes, just thinking out loud...

iloveagent57: oh, also, do you need a space in front of the `OR`?

iloveagent57 (commented):
2 minor paranoias about the deletion filter, then it's good to go.

enterprise_catalog/apps/api_client/algolia.py
adamstankiewicz: No issues on the colons that are in most of course keys due to the quotes, but yes, great catch on that missing space in front of the `OR`!

Without the space, the filter query looks something like:
```
`'key:"edX+TestX"OR key:"edX+DemoX"OR key:"edX+ExampleX"'`
```
Will add in the space.

iloveagent57 (approved):
Nice job 👍

PR #282, opened 2021-04-05
fix: refactor Algolia indexing to use ``replace_all_objects``

adamstankiewicz:
## Description

This PR is a (final) continuation of the Algolia indexing refactoring to ensure we don't have things in our Algolia index that shouldn't be there, as some search filters (e.g., content language) are exposing courses that are no longer in the catalog.

Previously, we were using `partial_update_objects` but it seems this will only create/update records with an exact match of an `objectId`. Due to how we batch the list of customer/catalog/catalog query UUIDs for each course (dynamic lists), there's no guarantee the `objectIds` will remain the same between indexing. As a result, we are leaving courses in the index that should no longer be there.

Instead, this refactor switches to use [`replace_all_objects`](https://www.algolia.com/doc/api-reference/api-methods/replace-all-objects/) which will create a temporary index with the exact objects we'd like in the index, and then, move all the objects from the temporary index into the actual index in an atomic reindex (aka no downtime).

When a course is no longer published or enrollable, we no longer have to explicitly delete content keys because we are essentially creating a brand new (temporary) index and replacing the existing one with the newly created one.

This should have a similar number of Algolia operations as we had with `partial_update_objects`.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -319,18 +319,18 @@ def index_enterprise_catalog_courses_in_algolia_task(self):
 
 def index_content_keys_in_algolia(content_keys, algolia_client):
     """
-    Adds/updates the metadata for the specified content keys in the Algolia index.
+    Determines list of Algolia objects to include in the Algolia index based on the
+    specified content keys, and replaces all existing objects with the new ones in an atomic reindex.
 
     Arguments:
-        content_keys (list): List of content_key strings.
+        content_keys (list): List of indexable content_key strings.
         algolia_client: Instance of an Algolia API client
     """
     logger.info(
-        'There are {} total content keys to add or '
-        'update in the Algolia index.'.format(len(content_keys))
+        'There are {} total content keys to include in the Algolia index.'.format(len(content_keys))
     )
+    courses = []
     for content_keys_batch in batch(content_keys, batch_size=TASK_BATCH_SIZE):
-        courses = []
         catalog_uuids_by_course_key = defaultdict(set)
         catalog_query_uuids_by_course_key = defaultdict(set)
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -319,18 +319,18 @@ def index_enterprise_catalog_courses_in_algolia_task(self):
...
-        content_keys (list): List of content_key strings.
+        content_keys (list): List of indexable content_key strings.
         algolia_client: Instance of an Algolia API client
     """
     logger.info(
-        'There are {} total content keys to add or '
-        'update in the Algolia index.'.format(len(content_keys))
+        'There are {} total content keys to include in the Algolia index.'.format(len(content_keys))
     )
+    courses = []
```
iloveagent57, line 332: This is gonna end up being a list of about 70k moderately-sized objects in prod.  I think this will be ok, as long as each record is less than maybe 5KB (which would make this list about 3.5GB of memory).  Do you happen to know the average size of a record?

iloveagent57 (commented):
My only concern is that we'd be writing one giant batch of data to the index.

enterprise_catalog/apps/api/tasks.py
adamstankiewicz: @iloveagent57 The average size of our Algolia records is currently 5.88 KB. 

Also, because we're no longer indexing courses that are not published, enrollable, or marketable, the number of records we index has significantly dropped from about that 70k to just about 12k.

iloveagent57: For posterity: I did not math correctly, and 5KB should really be 50KB.  Given the average size of our records that Adam posted, we're fine here.

iloveagent57 (approved):
🚢

PR #283, opened 2021-04-05
fix: index all ContentMetadata records of type course

adamstankiewicz:
## Description

Since we are now clearing out the Algolia index and replacing it with fresh records each time `reindex_algolia` is run, we should no longer filter ContentMetadata records by when they were modified as we could easily end up in a situation where we accidentally index 0 courses, and break the search results for all customers.

Instead of indexing only the recently modified `ContentMetadata` course objects, this PR switches to index all `ContentMetadata` course objects, regardless of when it was last modified. This will ensure we always have all courses indexed each run of the job.

## Post-review


iloveagent57 (approved):
Hmm, it does not speak highly of my unit-testing skills that you removed this line and none of the tests failed.

PR #284, opened 2021-04-08
fix: conditionally update json_metadata for courses with specific fields from /search/all data

adamstankiewicz:
## Description

The first stage of our enterprise catalog jobs are to source data from discovery's /search/all endpoint. We previously updated/created ContentMetadata object's `json_metadata` field with the metadata returned by this /search/all endpoint. However, we rely on the full course metadata for our ContentMetadata objects with content_type == "course" for Algolia indexing. This is metadata is populated via `update_full_content_metadata_task`. 

Due to our reliance on the full course metadata, we no longer want to fully replace the `json_metadata` field of existing ContentMetadata records with the /search/all metadata as we have seen several cases where the `update_content_metadata` Jenkins job times out, and the full course metadata update never occurs. This means the Algolia indexing that occurs an hour or two later does not have the necessary data for Algolia re-indexing, causing partner logo and name to be removed from our search results in the Learner Portal and elsewhere.

This refactoring also switches to using a bulk DB operation (i.e., `bulk_update`) for updating existing ContentMetadata objects and an atomic DB transaction for creating new ContentMetadata objects, which should improve task performance.

## Ticket Link

[ENT-4208](https://openedx.atlassian.net/browse/ENT-4208)

## Post-review

## Testing instructions

1. Ensure no ContentMetadata records exist in your devstack database, you have existing enterprise catalog(s) for course content, and your discovery service is running.
2. Run `./manage.py update_content_metadata` in the enterprise-catalog shell.
3. Observe ContentMetadata records having been created using the `/search/all` metadata. For courses, expect to see the merged `/search/all` metadata with the full course metadata.
4. Edit the `json_metadata` for an existing course to remove fields that only exist in the `/search/all` metadata (e.g., "aggregation_key", and re-run `./manage.py update_content_metadata` (you may need to temporarily comment out the `@expiring_task_semaphore` decorator on the tasks).
5. Observe that despite removing "aggregation_key" from an existing ContentMetadata, the `json_metadata` is populated again with the plucked fields, including "aggregation_key", from the `/search/all` metadata.

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -4,7 +4,7 @@ from logging import getLogger
 from uuid import uuid4
 
 from django.conf import settings
-from django.db import models
+from django.db import IntegrityError, models, transaction
 from django.db.models import Q
 from django.utils.translation import gettext as _
 from edx_rbac.models import UserRole, UserRoleAssignment
@@ -31,11 +31,11 @@ from enterprise_catalog.apps.catalog.constants import (
     json_serialized_course_modes,
 )
 from enterprise_catalog.apps.catalog.utils import (
+    batch,
     get_content_filter_hash,
     get_content_key,
     get_content_type,
     get_parent_content_key,
-    get_sorted_string_from_json,
     localized_utcnow,
 )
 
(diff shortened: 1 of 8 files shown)


enterprise_catalog/apps/catalog/utils.py
```diff
@@ -80,19 +80,6 @@ def get_jwt_roles(request):
     return feature_roles_from_jwt(decoded_jwt)
 
 
-def get_sorted_string_from_json(json_metadata):
```
adamstankiewicz, line 83: No longer used anywhere.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -393,12 +393,162 @@ def content_metadata_with_type_course():
...
+
+    Returns:
+        list: List of ContentMetadata objects that were created.
+    """
+    metadata_list = []
+    try:
+        with transaction.atomic():
+            for defaults in nonexisting_metadata_defaults:
+                content_metadata = ContentMetadata.objects.create(**defaults)
+                metadata_list.append(content_metadata)
```
adamstankiewicz, line 516: I attempted to use `bulk_create` here, but because it doesn't actually call `.save()`, when we associated the newly created ContentMetadata object with the catalog query later on in the task, it errored. To avoid this, I switched back to individual create operations but with an atomic DB transaction.

The updates of existing ContentMetadata records was able to switch to using `bulk_update`, though.

iloveagent57: I'm not sure if an atomic transaction helps us at all here - though I guess it doesn't hurt, either.  It shouldn't make a difference in performance - the same number (actually, more) of SQL statements get executed; the only difference is that either all of the metadata records will get created, or none of them will.

adamstankiewicz: Yeah, it's not really adding much. But I figured if something were to go awry, where the same content keys were already created in a separate batch for whatever reason, for example, at least we won't have some ContentMetadata in a batch created and others not.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -393,12 +393,164 @@ def content_metadata_with_type_course():
...
+        # Only include the json_metadata fields from /search/all that is not present in the
+        # full course metadata to avoid changing the ``get_content_metadata`` API contract.
+        fields_to_pluck = ['aggregation_key', 'content_type', 'seat_types', 'end_date', 'course_ends', 'languages']
+        entry_minimal = {}
+        for field in fields_to_pluck:
+            if value := entry.get(field):
+                entry_minimal.update({field: value})
+        if entry_minimal:
+            defaults.update({'json_metadata': entry_minimal})
+    elif (content_type != 'course' and exists) or not exists:
```
iloveagent57: Nit: I think 
```
elif not exists or (content_type != 'course'):
    ...
```
is equivalent and easier to understand.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -408,34 +560,20 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
-        content_key = get_content_key(entry)
-        defaults = {
-            'content_key': content_key,
-            'json_metadata': entry,
-            'parent_content_key': get_parent_content_key(entry),
-            'content_type': get_content_type(entry),
-        }
+    for batched_metadata in batch(metadata, batch_size=100):
+        content_keys = [get_content_key(entry) for entry in batched_metadata]
+        existing_metadata = ContentMetadata.objects.filter(content_key__in=content_keys)
```
iloveagent57, line 537: Strong suggestion: if you take this and create a dict out of it
```python
existing_metadata_by_key = {metadata.content_key: metadata for metadata in existing_metadata}
```
and use that everywhere you're currently using `existing_metadata`, your life becomes way easier in all of the helper methods.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -393,12 +393,164 @@ def content_metadata_with_type_course():
...
+    ]
+    nonexisting_metadata_defaults = [
+        _get_defaults_from_metadata(entry)
+        for entry in batched_metadata
+        if not get_content_key(entry) in existing_content_keys
+    ]
+    return existing_metadata_defaults, nonexisting_metadata_defaults
+
+
+def _find_entry_in_existing_metadata(content_key, existing_metadata):
```
iloveagent57: For instance, you could scratch this entire function.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -393,12 +393,164 @@ def content_metadata_with_type_course():
...
+        list: List of ContentMetadata objects that were updated.
+    """
+    metadata_list = []
+    for defaults in existing_metadata_defaults:
+        content_metadata = _find_entry_in_existing_metadata(defaults['content_key'], existing_metadata)
+        if content_metadata:
+            for key, value in defaults.items():
+                if key == 'json_metadata':
+                    # merge new json_metadata with old json_metadata (i.e., don't replace it fully)
+                    content_metadata.merge_json_metadata(value)
```
iloveagent57: I can't remember anymore why we need `merge_json_metadata` to make a copy before updating.  Can we just do `content_metadata.json_metadata.update(value)`?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -393,12 +393,164 @@ def content_metadata_with_type_course():
...
+        'parent_content_key': parent_content_key,
+        'content_type': content_type,
+    }
+    if content_type == 'course' and exists:
+        # Only include the json_metadata fields from /search/all that is not present in the
+        # full course metadata to avoid changing the ``get_content_metadata`` API contract.
+        fields_to_pluck = ['aggregation_key', 'content_type', 'seat_types', 'end_date', 'course_ends', 'languages']
+        entry_minimal = {}
+        for field in fields_to_pluck:
+            if value := entry.get(field):
```
iloveagent57, line 426: YES

enterprise_catalog/apps/catalog/models.py
```diff
@@ -393,12 +393,164 @@ def content_metadata_with_type_course():
...
+    content_type = get_content_type(entry)
+    defaults = {
+        'content_key': content_key,
+        'parent_content_key': parent_content_key,
+        'content_type': content_type,
+    }
+    if content_type == 'course' and exists:
+        # Only include the json_metadata fields from /search/all that is not present in the
+        # full course metadata to avoid changing the ``get_content_metadata`` API contract.
+        fields_to_pluck = ['aggregation_key', 'content_type', 'seat_types', 'end_date', 'course_ends', 'languages']
```
iloveagent57: Nit: might be good to define this as a constant, possibly in a settings file if we want to change it on the fly.

iloveagent57 (commented):
This is looking great, I think the "strong suggestion" below will help it out a bunch.

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: Ah yes, good point. I will update with this suggestion!

enterprise_catalog/settings/base.py
```diff
@@ -359,6 +359,10 @@
     'API_KEY': '',
 }
 
+# Which fields should be plucked from the /search/all course-discovery API
+# response in `update_catalog_metadata_task` for course content metadata?
+FIELDS_TO_PLUCK_FROM_SEARCH_ALL = os.environ.get('FIELDS_TO_PLUCK_FROM_SEARCH_ALL', [])
```
adamstankiewicz: https://github.com/edx/edx-internal/pull/4905

enterprise_catalog/settings/base.py
```diff
@@ -359,6 +359,10 @@
     'API_KEY': '',
 }
 
+# Which fields should be plucked from the /search/all course-discovery API
+# response in `update_catalog_metadata_task` for course content metadata?
+COURSE_FIELDS_TO_PLUCK_FROM_SEARCH_ALL = os.environ.get('COURSE_FIELDS_TO_PLUCK_FROM_SEARCH_ALL', [])
```
adamstankiewicz: https://github.com/edx/edx-internal/pull/4905

iloveagent57: Maybe we can have
```
[
    'aggregation_key',
    'content_type',
    'seat_types',
    'end_date',
    'course_ends',
    'languages',
]
```
be the default?   And then you won't have to redefine it in the devstack and test settings.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -393,12 +386,144 @@ def content_metadata_with_type_course():
...
+        'parent_content_key': parent_content_key,
+        'content_type': content_type,
+    }
+    if content_type == 'course' and exists:
+        # Only include the json_metadata fields from /search/all that is not present in the
+        # full course metadata to avoid changing the ``get_content_metadata`` API contract.
+        entry_minimal = {}
+        for field in settings.COURSE_FIELDS_TO_PLUCK_FROM_SEARCH_ALL:
+            if value := entry.get(field):
+                entry_minimal.update({field: value})
```
iloveagent57: Nit: `entry_minimal[field] = value` works; you don't need the `update({...})`.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -393,12 +386,144 @@ def content_metadata_with_type_course():
...
+                    # merge new json_metadata with old json_metadata (i.e., don't replace it fully)
+                    content_metadata.json_metadata.update(value)
+                else:
+                    # replace attributes with new values
+                    setattr(content_metadata, key, value)
+            metadata_list.append(content_metadata)
+
+    metadata_fields_to_update = ['content_key', 'parent_content_key', 'content_type', 'json_metadata']
+    existing_metadata_objs = existing_metadata_by_key.values()
+    ContentMetadata.objects.bulk_update(existing_metadata_objs, metadata_fields_to_update)
```
iloveagent57: Could we use `metadata_list` here instead of `existing_metadata_objs`?

iloveagent57 (commented):
A couple more small suggestions.

enterprise_catalog/apps/catalog/models.py
adamstankiewicz: D'oh, great catch. Yes, this should have been using `metadata_list` here.

enterprise_catalog/settings/base.py
chavesj: ✅

chavesj (approved):
LGTM, clean and easy to follow 🙏

iloveagent57 (approved):

PR #285, opened 2021-04-08
feat: Include the titles of programs as an Algolia field

iloveagent57:
...so that we can facet searches by the title of a program (rather than the type of a program).

## Description

This is the work to provide program titles as a facet-able field to Algolia.

## Ticket Link

Link to the associated ticket
[ENT-4365](https://openedx.atlassian.net/browse/ENT-4365)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -25,6 +25,7 @@ ALGOLIA_FIELDS = [
     'objectID',  # required by Algolia, e.g. "course-{uuid}"
     'partners',
     'programs',
+    'program_titles',
     'recent_enrollment_count',
     'short_description',
     'subjects',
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -232,6 +233,23 @@ def get_course_partners(course):
...
+    Arguments:
+        course (dict): a dictionary representing a course
+        field (str): the name of a field to return values of.
+    Returns:
+        list: a list of the values for a certain field in a program associated with the course.
+    """
+    programs = course.get('programs') or []
+    return list({
+        value for program in programs
+        if (value := program.get(field))
```
iloveagent57, line 249: I think this my first ever commit that includes the walrus operator.
https://www.python.org/dev/peps/pep-0572/

adamstankiewicz: Fancy, fancy. That's useful to know about. Related, Github needs a walrus emoji ;)

adamstankiewicz (approved):

PR #287, opened 2021-04-09
feat: Handle and update catalog query UUID

taliaedX:
## Description

Adds/modifies handling of query and content filter creates and updates from query parameters coming from LMS.

## Ticket Link

[Query Syncing from LMS to Enterprise Catalog](https://openedx.atlassian.net/browse/ENT-4429)

## Testing Scenarios:
The following were tested manually (and most later tested via code tests):
- new UUID; old filter > should update old filter with that UUID
- new UUID; new filter > should create a new query
- old UUID; new filter > should update existing query found by UUID with new filter
- old UUID; duplicated filter - should throw IntegrityError
- old UUID; old filter (aka no changes) - should return existing query
- no UUID, new filter > should create a new query
- no UUID, old filter > should return existing query
- no UUID, no filter > should return existing query on update catalog call

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/serializers.py
+++ b/enterprise_catalog/apps/api/v1/serializers.py
@@ -1,7 +1,7 @@
 import logging
 
 from django.db import IntegrityError
-from rest_framework import serializers
+from rest_framework import serializers, status
 
 from enterprise_catalog.apps.api.v1.utils import (
     get_enterprise_utm_context,
@@ -26,6 +26,47 @@ from enterprise_catalog.apps.catalog.utils import (
 logger = logging.getLogger(__name__)
 
 
+def find_and_modify_catalog_query(content_filter, catalog_query_uuid=None):
+    """
+    This method aims to make sure UUID and content_filter in the catalog service
+    match what Django Admin/passed in parameters have. We take the parameters as source of truth,
+    but do not want to duplicate either UUID or content filter.
+
+    Arguments:
+        content_filter(dict): filter used to pick which courses are retrieved
+        catalog_query_uuid(UUID/str): query uuid generated from LMS Django Admin.
+            - If not provided, we should be receiving a "direct" content filter.
+    Returns:
+        a CatalogQuery object.
+    """
+    if catalog_query_uuid:
+        catalog_query_from_uuid = CatalogQuery.get_by_uuid(uuid=catalog_query_uuid)
+        if catalog_query_from_uuid:
+            catalog_query_from_uuid.content_filter = content_filter
+            catalog_query_from_uuid.content_filter_hash = get_content_filter_hash(content_filter)
+            try:
+                catalog_query_from_uuid.save()
+            except IntegrityError as exc:
+                raise serializers.ValidationError(
+                    {'catalog_query': 'Content filter is not unique'},
+                    code=status.HTTP_422_UNPROCESSABLE_ENTITY
+                ) from exc
+            return catalog_query_from_uuid
+        else:
+            content_filter_from_hash, _ = CatalogQuery.objects.update_or_create(
+                content_filter_hash=get_content_filter_hash(content_filter),
+                defaults={'content_filter': content_filter, 'uuid': catalog_query_uuid}
+            )
+            return content_filter_from_hash
+    else:
+        content_filter_from_hash, _ = CatalogQuery.objects.get_or_create(
+            content_filter_hash=get_content_filter_hash(content_filter),
+            defaults={'content_filter': content_filter}
+        )
+
+        return content_filter_from_hash
+
+
 class EnterpriseCatalogSerializer(serializers.ModelSerializer):
     """
     Serializer for the `EnterpriseCatalog` model
@@ -35,6 +76,7 @@ class EnterpriseCatalogSerializer(serializers.ModelSerializer):
     enabled_course_modes = serializers.JSONField(write_only=True)
     publish_audit_enrollment_urls = serializers.BooleanField(write_only=True)
     content_filter = serializers.JSONField(write_only=True)
+    catalog_query_uuid = serializers.UUIDField(required=False)
 
     class Meta:
         model = EnterpriseCatalog
@@ -46,14 +88,13 @@ class EnterpriseCatalogSerializer(serializers.ModelSerializer):
             'enabled_course_modes',
             'publish_audit_enrollment_urls',
             'content_filter',
+            'catalog_query_uuid',
         ]
 
     def create(self, validated_data):
         content_filter = validated_data.pop('content_filter')
-        catalog_query, _ = CatalogQuery.objects.get_or_create(
-            content_filter_hash=get_content_filter_hash(content_filter),
-            defaults={'content_filter': content_filter},
-        )
+        catalog_query_uuid = validated_data.pop('catalog_query_uuid', None)
+        catalog_query = find_and_modify_catalog_query(content_filter, catalog_query_uuid)
         try:
             catalog = EnterpriseCatalog.objects.create(
                 **validated_data,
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -26,6 +26,48 @@
...
+            - If not provided, we should be receiving a "direct" content filter.
+    Returns:
+        a CatalogQuery object.
+    """
+    if catalog_query_uuid:
+        uuid_cg = CatalogQuery.objects.filter(uuid=catalog_query_uuid).first()
+        if uuid_cg:
+            uuid_cg.content_filter = content_filter
+            uuid_cg.content_filter_hash = get_content_filter_hash(content_filter)
+            uuid_cg.save()
```
taliaedX: So for our edge case, ie we get sent a "new" content filter that is actually a duplicate, how do we want to handle the inevitable IntegrityError thrown here?
Because currently it would throw a 500

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -26,6 +26,48 @@
 logger = logging.getLogger(__name__)
 
 
+def find_and_modify_catalog_query(content_filter, catalog_query_uuid=None):
```
taliaedX, line 29: This method is in here and not in a utils file because importing CatalogQuery into the utils causes a circular dependency error. If anyone knows how to resolve that I can move these around.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -26,6 +26,48 @@
...
+                    uuid=catalog_query_uuid)
+                return new_content_query
+
+    else:
+        content_filter_cg = CatalogQuery.objects.filter(content_filter_hash=get_content_filter_hash(content_filter)).first()
+        if content_filter_cg:
+            return content_filter_cg
+        else:
+            new_content_query = CatalogQuery.objects.create(content_filter_hash=get_content_filter_hash(content_filter), content_filter=content_filter)
+            return new_content_query
```
iloveagent57: Small suggestion: you could make this block a classmethod on `CatalogQuery`, something like `get_or_create_by_content_filter(cls, content_filter)`.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -26,6 +26,48 @@
...
+
+    Arguments:
+        content_filter(dict): filter used to pick which courses are retrieved
+        catalog_query_uuid(UUID/str): query uuid generated from LMS Django Admin.
+            - If not provided, we should be receiving a "direct" content filter.
+    Returns:
+        a CatalogQuery object.
+    """
+    if catalog_query_uuid:
+        uuid_cg = CatalogQuery.objects.filter(uuid=catalog_query_uuid).first()
```
iloveagent57: this can be a `get()` instead of `filter()`, then you don't need the `first()`.

enterprise_catalog/apps/api/v1/serializers.py
iloveagent57: Something like that should return a 422 response code, IMO.  You could raise a `ValidationError` (or a customer subclass thereof) in here and let that bubble up to the view, then handle it in the view.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -26,6 +26,48 @@
...
+        if uuid_cg:
+            uuid_cg.content_filter = content_filter
+            uuid_cg.content_filter_hash = get_content_filter_hash(content_filter)
+            uuid_cg.save()
+            return uuid_cg
+        else:
+            content_filter_cg = CatalogQuery.objects.filter(content_filter_hash=get_content_filter_hash(content_filter)).first()
```
iloveagent57: Same `filter()` -> `get()` thing here.

iloveagent57 (commented):
Looking good, handful of small suggestions for now.

enterprise_catalog/apps/api/v1/serializers.py
taliaedX: Get throws a DoesNotExist error though if the uuid/filter are not found. I was trying to avoid having a Try/except block on that error.
ie:
```
>>> value = CatalogQuery.objects.get(uuid='3be58e0888edbfb9869c3f6e93945bd6')
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/venv/lib/python3.8/site-packages/django/db/models/manager.py", line 82, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/venv/lib/python3.8/site-packages/django/db/models/query.py", line 406, in get
    raise self.model.DoesNotExist(
enterprise_catalog.apps.catalog.models.CatalogQuery.DoesNotExist: CatalogQuery matching query does not exist.
```

iloveagent57: Ah, right right.  Consider adding a helpful classmethod like `CatalogQuery.get_by_uuid(cls, uuid)` which returns the record, or null if it encounters `CatalogQuery.DoesNotExist`.

enterprise_catalog/apps/api/v1/serializers.py
taliaedX: Interesting. I just noticed for the EnterpriseCatalog `create` here we're just raising the IntegrityError and don't appear to be catching that error in the views either.

To handle this in the Views what would I wrap in the try/except? I see some instances where we override the HTTP methods (such as Post/Patch/get/etc.), but I'm not sure what logic would actually exist in the view function to wrap and how it's affected by us flipping which serializer we use based on the http command.

enterprise_catalog/apps/api/v1/serializers.py
taliaedX: get_or_create is already a method. From this comment I realize this whole block is just the logic of a get or create, so let me just swap to that.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -26,6 +26,41 @@
...
+
+    Arguments:
+        content_filter(dict): filter used to pick which courses are retrieved
+        catalog_query_uuid(UUID/str): query uuid generated from LMS Django Admin.
+            - If not provided, we should be receiving a "direct" content filter.
+    Returns:
+        a CatalogQuery object.
+    """
+    if catalog_query_uuid:
+        uuid_cg = CatalogQuery.get_by_uuid(uuid=catalog_query_uuid)
```
iloveagent57: Nit/suggestion: maybe `catalog_query_from_uuid`?  I don't know what the `cg` is for.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -26,6 +26,41 @@
...
+        if uuid_cg:
+            uuid_cg.content_filter = content_filter
+            uuid_cg.content_filter_hash = get_content_filter_hash(content_filter)
+            uuid_cg.save()
+            return uuid_cg
+        else:
+            content_filter_cg, _ = CatalogQuery.objects.update_or_create(
```
iloveagent57: nit: `catalog_query_from_hash`

iloveagent57 (commented):
Two naming nits - not sure if you're still in the middle of modifying.

enterprise_catalog/apps/api/v1/serializers.py
taliaedX: `cg` is my brain failing. I was trying to make an acryonym for catalog query, but your suggestion seems fine.

enterprise_catalog/apps/api/v1/serializers.py
taliaedX: Any further thoughts here @iloveagent57 ?

iloveagent57: Ah, sorry I mislead you into the view; this is one of the cases where the convenience of DRF shines.  `ValidationError` nicely lets you specify a status code that's used in the response: https://www.django-rest-framework.org/api-guide/exceptions/#validationerror
So for this duplicate case, you can do something like 
```
raise serializers.ValidationError({'catalog_query': 'UUID is not unique'}, code=status.HTTP_422_UNPROCESSABLE_ENTITY)
```

taliaedX: Mmmk so that is being caught and the error message is coming through great. However, 422 is getting reduced to just a 400. Not sure why or if that's a big issue really.

taliaedX: upon some searching, it looks like the "is_valid" run on serializers may return its own ValidationError which always returns 400. I'd have to override the `is_valid` function to get a custom error message. However, I may do a little more digging, because why have a code parameter if it doesn't use it properly.

iloveagent57: Ah, ok.  I wouldn't worry about 422 vs. 400, as long as the detail is coming through with a 4xx code.

iloveagent57 (approved):

PR #291, opened 2021-04-13
docs: add note to keep a log message consistent

adamstankiewicz:


taliaedX (approved):
Thanks for adding this! LGTM

PR #292, opened 2021-04-15
fix: makes subject, partner, program facets searchable

chavesj:
## Description

makes subject, partner, program facets searchable

## Ticket Link

Link to the associated ticket
[ENT-4460](https://openedx.atlassian.net/browse/ENT-4460)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -19,7 +19,6 @@ ALGOLIA_FIELDS = [
     'enterprise_catalog_query_uuids',
     'enterprise_customer_uuids',
     'full_description',
-    'key',  # for links to Course about pages from the Learner Portal search page
     'language',
     'level_type',
     'objectID',  # required by Algolia, e.g. "course-{uuid}"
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -54,11 +53,11 @@
         'enterprise_customer_uuids',
         'language',
         'level_type',
-        'partners.name',
+        'searchable(partners.name)',
         'programs',
```
adamstankiewicz: Probably worth making `programs` searchable as well as `program_titles` for if/when the `FEATURE_PROGRAM_TITLES_FACET` gets turned off.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -54,11 +53,11 @@
...
-        'program_titles',
+        'searchable(program_titles)',
         'searchable(skill_names)',
-        'subjects',
+        'searchable(subjects)',
         'key',  # necessary for deleting non-indexable course records from the Algolia index
```
adamstankiewicz: This is the `key` that should be removed. The `key` above is necessary for ~indexing to work~ course cards in learner portal to link to course pages based on this key from Algolia 😜

adamstankiewicz (approved):

PR #294, opened 2021-04-23
adds mgmt command argument and post data to force execute tasks using expiring_task_semaphore

chavesj:
## Description

Allows for our three management commands to be `force` run.  Useful in situations where you want to run the same `task(input)` more than once in a given hour.  Beneficial locally; will become useful in other environments in a later PR that introduces `ConfigurationModels` for these commands.

## Ticket Link

Link to the associated ticket
[4206](https://openedx.atlassian.net/browse/ENT-4206)

## Testing Instructions

Use the `make app-shell` bash shell to trigger the commands and observe any relevant output in the app-logs and worker-logs. e.g.
```
make app-shell
./manage.py update_content_metadata
./manage.py update_content_metadata # second invocation should fail because of task-locking
./manage.py update_content_metadata --force # see that this succeeds
```

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -1,5 +1,6 @@
 import copy
 import functools
+import json
 import logging
 from collections import defaultdict
 from datetime import timedelta
@@ -101,8 +102,8 @@ def task_recently_run(task_object, time_delta):
     """
     return TaskResult.objects.filter(
         task_name=task_object.name,
-        task_args=str(task_object.request.args),
-        task_kwargs=str(task_object.request.kwargs),
+        task_args=json.dumps(task_object.request.args),
+        task_kwargs=json.dumps(task_object.request.kwargs),
         date_created__gte=localized_utcnow() - time_delta,
     ).exclude(
         status__in=(states.FAILURE, states.REVOKED),
@@ -132,6 +133,8 @@ def expiring_task_semaphore(time_delta=None):
     was executed in the time between `time_delta` and now, the task moves to a REVOKED
     state and raises a `TaskRecentlyRunError`.
 
+    If task is invoked with `force` kwarg, time since last run will be ignored.
+
     The `meta` state of the task is updated
     with `exc_type` and `exc_info`; otherwise, any process that's watching for the result
     of the task (e.g. via `task.get()`) and finds that it results in a failed state
@@ -147,8 +150,9 @@ def expiring_task_semaphore(time_delta=None):
     def decorator(task):
         @functools.wraps(task)
         def wrapped_task(self, *args, **kwargs):
+            force = kwargs.get('force', False)
             delta = time_delta or ONE_HOUR
-            if task_recently_run(self, time_delta=delta):
+            if not force and task_recently_run(self, time_delta=delta):
                 msg_args = (self.name, self.request.id, self.request.args, self.request.kwargs)
                 message = (
                     '{} task with id {} was recently run with '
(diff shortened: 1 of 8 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -189,15 +193,14 @@ class LoggedTaskWithRetry(LoggedTask):  # pylint: disable=abstract-method
...
     """
     Looks up the full course metadata from discovery's `/api/v1/courses` endpoint to pad all
     ContentMetadata objects with. The course metadata is merged with the existing contents
     of the json_metadata field for each ContentMetadata record.
 
-    Note: It is especially important that this task uses the increased maximum ``CELERY_TASK_SOFT_TIME_LIMIT`` and
-    ``CELERY_TASK_TIME_LIMIT`` since the task traverses large portions of course-discovery's /courses/ endpoint, which
-    was exceeding the previous default limits, causing a SoftTimeLimitExceeded exception.
+    Args:
+        force (bool): If true, forces execution of task and ignores time since last run.
```
adamstankiewicz, line 203: To make sure I'm understanding this right, the `force` argument isn't used in the method itself since it's only used in the `@expiring_task_semaphore()` decorator?

I went back and forth a bit in my head on whether `force=True` should also bypass the `unready_tasks(..., ONE_HOUR).exists()` check in these methods as well, but agreed `force` shouldn't apply to `unready_tasks` since then we'd be in a state where there are potentially tasks overriding each other (e.g., we shouldn't force another task if a similar task is already started, but just not finished); just thinking out loud here 😄

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
```diff
@@ -42,13 +42,14 @@ def _run_update_full_content_metadata_task(self, *args, **kwargs):
...
 
     def add_arguments(self, parser):
-        # Argument to specify catalogs to update
+        # Argument to force execution of celery task, ignoring time since last execution
         parser.add_argument(
-            '--catalog_uuids',
-            nargs='+',
+            '--force',
+            default=False,
+            action='store_true',
```
adamstankiewicz, line 52: What is `action` doing here?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -258,12 +258,13 @@ def get_permission_object(self):
...
         enterprise_catalog = get_object_or_404(EnterpriseCatalog, uuid=uuid)
         catalog_query_id = enterprise_catalog.catalog_query.id
+        force = request.data.get('force', False)
 
         # Use immutable signatures so task results from a parent task are not passed as arguments to a child task.
         async_update_metadata_chain = chain(
             update_catalog_metadata_task.si(catalog_query_id),
-            update_full_content_metadata_task.si(),
-            index_enterprise_catalog_courses_in_algolia_task.si(),
+            update_full_content_metadata_task.si(force=force),  # TODO: why doesn't this work?
```
adamstankiewicz: This isn't working in that `force` isn't being passed as a kwarg? 

If that's the case, definitely odd since `si` for sure seems to accept both args and kwargs: https://docs.celeryproject.org/en/latest/_modules/celery/app/task.html#Task.si 🤔 

No immediate thoughts for a resolution yet. Does this need to be figured out prior to merge? I believe you mentioned a follow-up PR incoming for getting this to work for stage/prod as well.

enterprise_catalog/apps/api/tasks.py
iloveagent57: Yeah, exactly, only the decorator looks at that kwarg.  And good point about `unready_tasks`.

enterprise_catalog/apps/catalog/management/commands/update_content_metadata.py
iloveagent57: https://docs.python.org/3/library/argparse.html#action
In this case, if `--force` is provided as an option to the command, then `options['force']` will be `True`.  And since the default is `False`, `options['force']` will be `False` if that option isn't given.

enterprise_catalog/apps/api/v1/views.py
iloveagent57: it's actually working fine, that's a leftover TODO from pre-handoff.  I'll delete.

iloveagent57: actually, I think you saw an earlier commit - I ripped out all of the `force` changes in this file that Jeff had added, because I don't think we'll end up using it from the API.

adamstankiewicz (approved):
LGTM, though I'm curious whether the `TODO` comment is still valid/relevant:

`return update_full_content_metadata_task.si(force=kwargs.get('force', False))  # TODO: pass force argument to task`

PR #296, opened 2021-05-12
Config model to force-run mgmt commands

iloveagent57:
* Adds a configuration model to help force-run mgmt commands in stage/prod environments, so that
we can work around our task-locking logic in emergencies.
* Installs django-config-models and resolve version incompatibilities for some recently-released major versions.
* No longer runs Django 3 in tox validation, because django-config-models requires Django<3.

https://openedx.atlassian.net/browse/ENT-4206

Testing:
* Run `./manage.py migrate`
* Run any of the management commands e.g. `./manage.py update_content_metadata`
* Run it again to see the task lock.
* Go to http://localhost:18160/admin/catalog/catalogupdatecommandconfig/ and create a new config record, with `enabled` and `force` set to True.  Save it.
* Run the command again, see that it didn't task lock this time.
* Overwrite your config in http://localhost:18160/admin/catalog/catalogupdatecommandconfig/ with a new record to set `force` back to false.

## Post-review


enterprise_catalog/apps/catalog/models.py
```diff
@@ -664,3 +665,31 @@ def update_contentmetadata_from_discovery(catalog_query):
...
+        ),
+    )
+
+    @classmethod
+    def current_options(cls):
+        """
+        Returns a dictionary of options from this config model.
+        """
+        current_config = cls.current()
+        if current_config and current_config.enabled:
```
adamstankiewicz: nit: it looks like the `ConfigurationModel` has an `is_enabled` class method on it. Is that something we should use here vs. checking `enabled` ourselves?

https://github.com/edx/django-config-models/blob/master/config_models/models.py#L147

adamstankiewicz (approved):
LGTM with a non-blocking nit and after testing it locally as well.

enterprise_catalog/apps/catalog/models.py
iloveagent57: Oh good find!  Yeah, let's use that one.

iloveagent57: Actually, I'm going to leave it as-is: we have to fetch the current instance of the config anyway (i.e. `current_config.force`) and the `enabled` field isn't hidden or anything.  So I could re-write it like this
```python
if cls.is_enabled():
    return {'force': cls.current().force}
return {}
```
but I don't really see an advantage in doing so.  I guess I can reduce `if current_config and current_config.enabled` to just `if current_config.enabled`.

PR #300, opened 2021-05-27
ENT-4158: Added course skills to algolia index

irfanuddinahmad:
## Description

This PR adds the course skills to algolia index. The existing and similar skill_names section is currently not being deprecated since it is being used in a number of downstream locations, including catalog filters. Instead, the Skills section of the search/all endpoint is being added as:
```
"skills": [
                {
                    "name": "python",
                    "description": "python is a very popular language!!!"
                },
                {
                    "name": "django",
                    "description": ""
                },
                {
                    "name": "machine learning",
                    "description": "Machine Learning is one of the most growing career field right now."
                },
                {
                    "name": "data science",
                    "description": "Data Science jobs are the highest paid jobs in the market right now"
                }
            ],
```

## Ticket Link

https://openedx.atlassian.net/browse/ENT-4158

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -30,6 +30,7 @@ ALGOLIA_FIELDS = [
     'short_description',
     'subjects',
     'skill_names',
+    'skills',
     'title',
     'advertised_course_run',  # a part of the advertised course run
 ]
@@ -46,6 +47,7 @@ ALGOLIA_INDEX_SETTINGS = {
         'unordered(additional_information)',
         'partners',
         'skill_names',
+        'skills',
     ],
     'attributesForFaceting': [
         'availability',
@@ -58,6 +60,7 @@ ALGOLIA_INDEX_SETTINGS = {
         'searchable(programs)',
         'searchable(program_titles)',
         'searchable(skill_names)',
+        'searchable(skills)',
         'searchable(subjects)',
     ],
     'unretrievableAttributes': [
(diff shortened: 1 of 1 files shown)


muhammad-ammar:
@irfanuddinahmad Some questions/concerns/thoughts.

- We already have `skill_names` in algolia index.
- How the new skills different from existing, `description` attribute? 
- Wouldn't this duplicate the data at two places? 
- Will updating the existing `skill_names` break the things in `learner portal` and other places?
- Should we need to create a data structure that can also works for learner pathways?

zamanafzal (approved):

sameenfatima78:
> 
> * We already have `skill_names` in algolia index.
> * How the new skills different from existing, `description` attribute?
> * Wouldn't this duplicate the data at two places?
> * Will updating the existing `skill_names` break the things in `learner portal` and other places?
> * Should we need to create a data structure that can also works for learner pathways?

@muhammad-ammar Regarding your first 4 questions, we are planning to keep the original field as is for now because it's being used currently at two different places, Learner Portal and Publisher. Once this new field is available in Algolia, I will update the frontend code to shift on using this new field and delete the original one. This new change would also break the existing catalogs if the original field is removed, so need to have discussion with Nathan on this first. 

Regarding the last question, what suggestions do you have in mind for that?

sameenfatima78 (approved):

irfanuddinahmad:
@adamstankiewicz the Algolia object of the new Skills section is a list of dicts. Can you review if this is ok or we need to convert to another data structure. Thanks.

adamstankiewicz:
> @adamstankiewicz the Algolia object of the new Skills section is a list of dicts. Can you review if this is ok or we need to convert to another data structure. Thanks.

@irfanuddinahmad A list of dicts (JSON objects) can be indexed by Algolia. The `partners` field is doing this as well 🙂

adamstankiewicz (approved):

muhammad-ammar (approved):

PR #301, opened 2021-06-21
upgraded packages

manny-m:
## Description

Closes [ENT-4528](https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=628&modal=detail&selectedIssue=ENT-4528)


requirements/constraints.txt
```diff
@@ -16,6 +16,9 @@ Django>=2.2,<2.3
 # celery 5.0 has dropped python3.5 support.
 celery<5.0
 
+# django celesry results wants celery >=5
```
iloveagent57, line 19: [typo] "celesry" -> "celery"
You might also want to specify that django-celery-results 2.1 wants celery >=5.0

iloveagent57 (approved):
Looks good, tiniest possible nit on a comment in `constraints.txt`.

adamstankiewicz (approved):

macdiesel (commented):
@iloveagent57 I'm not sure how picky enterprise is about these things but typically I would suggest squashing these commits down in to one and getting rid of that merge commit. Thoughts?

manny-m:
> @iloveagent57 I'm not sure how picky enterprise is about these things but typically I would suggest squashing these commits down in to one and getting rid of that merge commit. Thoughts?

My github user name was not in my `.gitconfig` so I amended and had things got a little weird. Wont happen again going forward. I can rebase if necessary.

iloveagent57:
We typically try to squash down to 1 commit, but it's not a big deal if something like a merge commit goes in, too.  The thing we really want to avoid is having like, 19 different commits with messages like "checkpoint" or "fixes test" that don't get squashed.

PR #303, opened 2021-07-13
Make fewer memcached gets during get_content_metadata action requests

iloveagent57:
## Description

Hypothesis: The serialization of content metadata adds a content enrollment URL to the metadata instance - the function that determines the content enrollment URL uses an `EnterpriseCustomerDetails` instance that fetches a dictionary of customer data from memcached on each read, and there are many reads of this data (hundreds per request, somehow?).

Solution: Add a cached_property to hold this `EnterpriseCustomerDetails` instance in memory for the life of the associated `EnterpriseCatalog` instance, which should cause us to only read this data from memcached once, at the start of the request.
## Ticket Link

[The ticket](https://openedx.atlassian.net/browse/ENT-4742)

## Post-review

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -20,6 +20,7 @@ services:
     container_name: enterprise.catalog.memcached
     networks:
       - devstack_default
+    command: memcached -vv
 
   app:
     # Uncomment this line to use the official catalog base image
--- a/enterprise_catalog/settings/devstack.py
+++ b/enterprise_catalog/settings/devstack.py
@@ -72,3 +72,10 @@ CORS_ORIGIN_WHITELIST = [
     'http://localhost:18000',
     'http://localhost:18130',
 ]
+
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
+        'LOCATION': 'enterprise.catalog.memcached:11211',
+    }
+}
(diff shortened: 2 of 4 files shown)


docker-compose.yml
```diff
@@ -20,6 +20,7 @@ services:
     container_name: enterprise.catalog.memcached
     networks:
       - devstack_default
+    command: memcached -vv
```
iloveagent57, line 23: Runs memcached in verbose mode, so we can tell when reads/writes occur.

enterprise_catalog/settings/devstack.py
```diff
@@ -72,3 +72,10 @@
     'http://localhost:18000',
     'http://localhost:18130',
 ]
+
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
+        'LOCATION': 'enterprise.catalog.memcached:11211',
+    }
+}
```
iloveagent57, line 81: We were using a Local memory cache backend in devstack before this change (which meant we weren't utilizing the memcached container we've been running). 🤦

adamstankiewicz: D'oh, great catch!

adamstankiewicz (approved):
Looks good! Let's hope this help reduce the number of reads per request 😄

PR #304, opened 2021-07-23
feat: adding last modified at as a returned field to the enterprise catalog endpoint

alex-sheehan-edx:
[Jira](https://openedx.atlassian.net/browse/ENT-4643)
[Proposal of improvement](https://openedx.atlassian.net/wiki/spaces/SOL/pages/3013509195/transmit+content+metadata+hammers+the+enterprise-catalog+DB+-+make+it+stop)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/serializers.py
+++ b/enterprise_catalog/apps/api/v1/serializers.py
@@ -1,6 +1,6 @@
 import logging
 
-from django.db import IntegrityError
+from django.db import IntegrityError, models
 from rest_framework import serializers, status
 
 from enterprise_catalog.apps.api.v1.utils import (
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -89,8 +90,20 @@ class Meta:
...
+        catalog_metadata_last_modified = obj.content_metadata.aggregate(models.Max('modified')).get('modified__max')
+        catalog_last_modified = obj.modified
+        if not catalog_last_modified:
+            if catalog_metadata_last_modified:
+                return catalog_metadata_last_modified
+            return None
+        elif not catalog_metadata_last_modified:
+            return catalog_last_modified
+        return max(catalog_metadata_last_modified, catalog_last_modified)
+
```
iloveagent57: We want the serializer to return two distinct bits of information here:
1. The `modified` time of the catalog object - you can add something like `catalog_modified = serializers.DatetimeField(source='modified')` to the list of serializer fields.
2. The time any assoicated content was modified - do what you're doing here, but remove all the logic that compares to `obj.modified` from your `get_content_last_modified()` function.

iloveagent57 (changes requested):
Slight change requested, but it's close.  You also get to close https://openedx.atlassian.net/browse/ENT-4734 with this change!

iloveagent57 (approved):
Nice job!

PR #309, opened 2021-08-27
feat: Utilize catalog query titles from edx-enterprise

taliaedX:
## Description
edx-enterprise already has a field called "title" for catalog queries.
This PR makes them exist on and passable to enterprise-catalog.
Title is optional to allow for backwards compatibility with the existing API.

There is a PR incoming for the edx-enterprise side to actually send the title attribute across.
(I have it locally, but need to submit the pr and will update this when I do)

## Ticket Link
[ENT-4600: Add Enterprise Catalog Short Name to Algolia](https://openedx.atlassian.net/browse/ENT-4600)

## Post-review


enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -43,25 +43,26 @@ def find_and_modify_catalog_query(content_filter, catalog_query_uuid=None):
...
         if catalog_query_from_uuid:
             catalog_query_from_uuid.content_filter = content_filter
+            catalog_query_from_uuid.title = query_title
             catalog_query_from_uuid.content_filter_hash = get_content_filter_hash(content_filter)
             try:
                 catalog_query_from_uuid.save()
             except IntegrityError as exc:
                 raise serializers.ValidationError(
-                    {'catalog_query': 'Content filter is not unique'},
+                    {'catalog_query': 'Content filter or title is not unique'},
```
taliaedX: I'm iffy on this change.
I'm not sure if we want to save these values separately to parse their errors separately, try to parse the IntegrityError since it does mention which key had the error, or something else here.

alex-sheehan-edx: There has to be a way to parse out the offending columns within the caught integrity error, but at the same time I guess the safe bet is to do two separate try/catches and saves of the different fields  ¯\_(ツ)_/¯

alex-sheehan-edx (approved):
LGTM- I think you could change that one try/catch that you had a question on but it's all good from me

enterprise_catalog/apps/api/v1/serializers.py
taliaedX: yeah it just then requires 2 separate database operations which is also stupid. 😞

iloveagent57: The `exc` message here should have some information about the offending field, right?  Also, while you're in here, it'd be super helpful to add a `logger.exception()` - we had a bug just last week that would've benefited from it.

iloveagent57 (approved):
Looks good here, but any additional logging/error message info is very appreciated.

enterprise_catalog/apps/api/v1/serializers.py
taliaedX: It does. I wasn't sure that it was obviously parseable though or just part of the IntegrityError string, but I can double-check that.

taliaedX: oh yeah wow, without that exception level logging that error gets completely swallowed and it looks like it is otherwise successful.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -26,42 +27,48 @@
...
             catalog_query_from_uuid.content_filter_hash = get_content_filter_hash(content_filter)
             try:
                 catalog_query_from_uuid.save()
             except IntegrityError as exc:
+                error = exc.args[1]
+                column = search("(?<=for key ')(.*)(?=')", error)
+                logger.exception(f'Error occurred while saving catalog query: {error}')
                 raise serializers.ValidationError(
-                    {'catalog_query': 'Content filter is not unique'},
+                    {'catalog_query': f'{column} is not unique'},
```
taliaedX, line 57: okay yeah @iloveagent57 so from what I can tell differentiating the column requires parsing the actual error message string.

The problem comes in with the fact that we don't use the same database engine in our tests (we use SQLite) so the error message isn't formatted the same (hell it doesn't even seem to have the same attributes on the exception object) and thus my tests are failing. 😡 

Any thoughts on getting around that or should we just go back to a generic message?

PR #311, opened 2021-08-31
fix: add the setting that indicates the name of the JWT auth cookie.

iloveagent57:
## Description

This sets `USE_JWT_COOKIE` to true in the `JWT_AUTH` config for this service.  Without this key/value, JWT-based auth was effectively not used at all in the service; instead, all authentication fell back to either Session authentication (in the case of requests from a browser) or OAuth authentication (in the case of serivce-to-service requests).

### Impacts

The fact that this was previously missing accounts for why there were so few `core.User` records in the enterprise-catalog production database - Session authentication does not create user records upon successful authentication.  (Note: records are only written to the `django_session` table when data in the session has changed, which is rare in the case of enterprise-catalog: [Django-when are sessions saved?)](https://docs.djangoproject.com/en/3.2/topics/http/sessions/#when-sessions-are-saved)

How did this cause the bug below?
* One of the symptoms of the bug was that the error went away in a new browser - which likely meant that the user had to re-login in a new browser, perhaps they had a stale session cookie in the first browser they tried.

How does this fix the bug below?
* As long as the SSO flow populates a JWT cookie for the SSO'd user, this change ensures that enterprise-catalog will look for a JWT Authentication token in the cookie named `edx-jwt-cookie` and use that token to authenticate.  As long as the JWT is valid, the authentication will be successful.  Prior to this change, the drf-jwt library (upon which edx-drf-extensions/auth/jwt depends) had no way of knowing _which_ cookie to look for an auth token in.

## Ticket Link

[Learners get Axios Error when they click on a Course Tile in the Learner Portal after logging in via SSO](https://openedx.atlassian.net/browse/ENT-4885)

## Post-review

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,9 +34,8 @@ services:
       - .:/edx/app/enterprise_catalog/enterprise_catalog
       - ../src:/edx/src:cached
 
-    # This should be the same as the `CMD` in the legacy devapp build stage
-    # of the Dockerfile
-    command: bash -c 'gunicorn --reload --workers=2 --name enterprise_catalog -b :18160 -c /edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/docker_gunicorn_configuration.py --log-file - --max-requests=1000 enterprise_catalog.wsgi:application'
+    # Use the Django devserver, so that we can hot-reload code changes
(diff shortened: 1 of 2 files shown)


adamstankiewicz (approved):
LGTM :shipit:

macdiesel (approved):

PR #313, opened 2021-09-02
bug: algolia index names did not match

alex-sheehan-edx:
## Description

`banner_image_urls` & `banner_image_url` => `original_image_url` 🤗 

## Post-review


taliaedX (approved):
much closer to the original (see what I did there). I like it.

PR #315, opened 2021-09-02
feat: adding marketing url to the algolia index

alex-sheehan-edx:
## Description

Adding marketing URL to the algolia index to provide users a method of reaching the B2C course details page.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -35,6 +35,7 @@ ALGOLIA_FIELDS = [
     'advertised_course_run',  # a part of the advertised course run
     'first_enrollable_paid_seat_price',
     'original_image_url',
+    'marketing_url',
 ]
 
 # default configuration for the index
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -66,6 +66,7 @@
         'searchable(subjects)',
         'first_enrollable_paid_seat_price',
         'original_image_url',
+        'marketing_url',
```
taliaedX, line 70: doesn't this also need to be in the Algolia Fields list or no?

taliaedX (commented):
Can't tell if it's missing from a spot or fine as is

enterprise_catalog/apps/catalog/algolia_utils.py
alex-sheehan-edx: weeeeeeee thx

taliaedX (approved):
lgtm!

PR #316, opened 2021-09-09
maint: adding query titles as a facet to the algolia index

alex-sheehan-edx:
## Description

in much the same way we have the catalog query uuids as facets on algolia, we want to add query titles. These titles are unique so we don't need to worry too much about colitions. Since many stage queries have no title, I've added a catch to account for non existent titles.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -291,6 +291,27 @@ def _batched_metadata(json_metadata, sorted_uuids, uuid_key_name, obj_id_fmt):
     return batched_metadata
 
 
+def _batched_metadata_with_queries(json_metadata, sorted_queries):
+    """
+    Batched catalog queries are represented as tuples (<query uuid>, <query title>). Unzip the two fields and update
+    them together.
+    """
+    batched_metadata = []
+    for batch_index, query_batch in enumerate(batch(sorted_queries, batch_size=ALGOLIA_UUID_BATCH_SIZE)):
+        json_metadata_with_uuids = copy.deepcopy(json_metadata)
+
+        query_uuids, query_titles = list(map(list, zip(*query_batch)))
+        metadata_to_update = {
+            'objectID': f"{json_metadata['objectID']}-catalog-query-uuids-{batch_index}",
+            'enterprise_catalog_query_uuids': sorted(query_uuids),
+            # filter out `None` from `query_titles`
+            'enterprise_catalog_query_titles': sorted([title for title in query_titles if title]),
+        }
+        json_metadata_with_uuids.update(metadata_to_update)
+        batched_metadata.append(json_metadata_with_uuids)
+    return batched_metadata
+
+
 @shared_task(base=LoggedTaskWithRetry, bind=True, default_retry_delay=UNREADY_TASK_RETRY_COUNTDOWN_SECONDS)
 @expiring_task_semaphore()
 def index_enterprise_catalog_courses_in_algolia_task(self, force=False):  # pylint: disable=unused-argument
@@ -333,9 +354,8 @@ def index_content_keys_in_algolia(content_keys, algolia_client):
     courses = []
     for content_keys_batch in batch(content_keys, batch_size=TASK_BATCH_SIZE):
         catalog_uuids_by_course_key = defaultdict(set)
-        catalog_query_uuids_by_course_key = defaultdict(set)
-        catalog_query_title_by_course_key = defaultdict(set)
         customer_uuids_by_course_key = defaultdict(set)
+        catalog_queries_by_course_key = defaultdict(set)
 
         # retrieve ContentMetadata records that match the specified content_keys in the
         # content_key or parent_content_key. returns both courses and course runs.
@@ -359,14 +379,10 @@ def index_content_keys_in_algolia(content_keys, algolia_client):
             course_content_key = metadata.content_key if is_course_content_type else metadata.parent_content_key
             associated_queries = metadata.catalog_queries.all()
             enterprise_catalog_uuids = set()
-            enterprise_catalog_query_uuids = set()
-            enterprise_catalog_query_titles = set()
             enterprise_customer_uuids = set()
+            enterprise_catalog_queries = set()
             for query in associated_queries:
-                enterprise_catalog_query_uuids.add(str(query.uuid))
-                # Many stage queries have no title
-                query_title = query.title if query.title else "catalog_query_{}".format(str(query.id))
-                enterprise_catalog_query_titles.add(query_title)
+                enterprise_catalog_queries.add((str(query.uuid), query.title))
                 associated_catalogs = query.enterprise_catalogs.all()
                 for catalog in associated_catalogs:
                     enterprise_catalog_uuids.add(str(catalog.uuid))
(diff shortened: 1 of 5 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -359,9 +360,13 @@ def index_content_keys_in_algolia(content_keys, algolia_client):
             associated_queries = metadata.catalog_queries.all()
             enterprise_catalog_uuids = set()
             enterprise_catalog_query_uuids = set()
+            enterprise_catalog_query_titles = set()
             enterprise_customer_uuids = set()
             for query in associated_queries:
                 enterprise_catalog_query_uuids.add(str(query.uuid))
+                # Many stage queries have no title
+                query_title = query.title if query.title else "catalog_query_{}".format(str(query.id))
```
binodpant: I am now beginning to default to f strings if you want too!

binodpant: is the title `catalog_query_{}` format something we need specifically or just a format, is all?

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -281,6 +285,11 @@ def setUpTestData(cls):
 
         # Set up new catalog, query, and metadata for a course run
         cls.enterprise_catalog_course_runs = EnterpriseCatalogFactory()
+
+        # for testing indexing catalog queries when titles aren't present
+        cls.enterprise_catalog_course_runs.catalog_query.title = None
```
binodpant: what happens if we don't set this to even None? Is that a state we can even be in ?

binodpant (approved):
lgtm but I dont know the indexing code very well so Adam may have a better review for you

enterprise_catalog/apps/api/tasks.py
alex-sheehan-edx: ok! I'm just for consistency so if f strings are what we want to do then let's go for it!

alex-sheehan-edx: `catalog_query_{}` is simply to get unique titles for queries that have none

enterprise_catalog/apps/api/tests/test_tasks.py
alex-sheehan-edx: `title` is a part of the query model. Correct me if I'm wrong, but we should either have a value for the title or `None`, this is just testing that we have something that we can index if there's no value (common for our stage environment but not prod)

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -130,13 +130,16 @@ def create(self, validated_data):
...
+        default_query_uuid = None
         if instance.catalog_query:
             default_content_filter = instance.catalog_query.content_filter
             default_query_title = instance.catalog_query.title if hasattr(instance.catalog_query, 'title') else None
+            default_query_uuid = str(instance.catalog_query.uuid)
 
         content_filter = validated_data.get('content_filter', default_content_filter)
         query_title = validated_data.get('query_title', default_query_title)
-        catalog_query_uuid = validated_data.pop('catalog_query_uuid', None)
+        catalog_query_uuid = validated_data.pop('catalog_query_uuid', default_query_uuid)
```
alex-sheehan-edx, line 142: small unrelated bug with put requests on catalogs with instance's available

enterprise_catalog/apps/api/tasks.py
```diff
@@ -505,7 +521,7 @@ def update_catalog_metadata_task(self, catalog_query_id, force=False):  # pylint
     try:
         catalog_query = CatalogQuery.objects.get(id=catalog_query_id)
     except CatalogQuery.DoesNotExist:
-        logger.error('Could not find a CatalogQuery with id %s', catalog_query_id)
+        logger.error('Could not find a CatalofgQuery with id %s', catalog_query_id)
```
adamstankiewicz: nit: typo in `CatalofgQuery` 😉

enterprise_catalog/apps/api/tasks.py
```diff
@@ -291,6 +291,25 @@ def _batched_metadata(json_metadata, sorted_uuids, uuid_key_name, obj_id_fmt):
...
+def _batched_metadata_with_queries(json_metadata, sorted_queries):
+    """
+    Batched catalog queries are represented as tuples (<query uuid>, <query title>). Unzip the two fields and update
+    them together.
+    """
+    batched_metadata = []
+    for batch_index, query_batch in enumerate(batch(sorted_queries, batch_size=ALGOLIA_UUID_BATCH_SIZE)):
+        json_metadata_with_uuids = copy.deepcopy(json_metadata)
+
+        query_titles, query_uuids = list(map(list, zip(*query_batch)))
```
adamstankiewicz: 🚀

enterprise_catalog/apps/api/tasks.py
```diff
@@ -358,10 +377,12 @@ def index_content_keys_in_algolia(content_keys, algolia_client):
             course_content_key = metadata.content_key if is_course_content_type else metadata.parent_content_key
             associated_queries = metadata.catalog_queries.all()
             enterprise_catalog_uuids = set()
-            enterprise_catalog_query_uuids = set()
             enterprise_customer_uuids = set()
+            enterprise_catalog_queries = set()
             for query in associated_queries:
-                enterprise_catalog_query_uuids.add(str(query.uuid))
+                # Many stage queries have no title
+                query_title = query.title if query.title else f"catalog_query_{str(query.id)}"
```
adamstankiewicz: I mentioned this in Slack, but I'm wondering if we should be generating a title if one does not exist. Since this created title won't exist anywhere but in Algolia, I'm not sure if the generated titles will ever get used as a facet refinement and it may be odd if it were ever exposed in a UI.

It may make sense to keep the title `None` and filter out `None` values from `'enterprise_catalog_query_titles': sorted(query_titles),` before adding the list to the json metadata.

enterprise_catalog/apps/api/tasks.py
alex-sheehan-edx: 🤦

enterprise_catalog/apps/api/tasks.py
```diff
@@ -291,6 +291,28 @@ def _batched_metadata(json_metadata, sorted_uuids, uuid_key_name, obj_id_fmt):
...
+    for batch_index, query_batch in enumerate(batch(sorted_queries, batch_size=ALGOLIA_UUID_BATCH_SIZE)):
+        json_metadata_with_uuids = copy.deepcopy(json_metadata)
+
+        query_titles, query_uuids = list(map(list, zip(*query_batch)))
+        metadata_to_update = {
+            'objectID': f"{json_metadata['objectID']}-catalog-query-uuids-{batch_index}",
+            'enterprise_catalog_query_uuids': sorted(query_uuids),
+        }
+        if query_uuids:
+            metadata_to_update['enterprise_catalog_query_titles'] = sorted(query_titles),
```
adamstankiewicz: Following up from our Slack conversation, since `query_titles` either contains the existing query title or `None` (if one doesn't exist), let's filter our `None`'s from the `query_titles` before adding it to `metadata_to_update`; that way the list will only contain titles since it won't make sense to facet based on a title that doesn't exist.

adamstankiewicz (commented):
This is looking good to me. Just that nit about the typo and the follow-up from our Slack conversation to filter our `None` values from the list of `query_titles` 😄

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -9,6 +9,7 @@
 COURSE = 'course'
 COURSE_RUN = 'courserun'
 PROGRAM = 'program'
+QUERY_TITLE = 'abc_catalog'
```
adamstankiewicz: nit: Is this `QUERY_TITLE` constant only relevant to tests? If so, it feels odd to put it in a non-tests constants.py file. Would it make more sense in a test-specific constants.py file or just define it in the single test file that makes use of it?

adamstankiewicz (approved):
LGTM!

PR #319, opened 2021-09-10
bug: accounting for the  availabilty when indexing algolia

alex-sheehan-edx:
## Description

[Jira](https://openedx.atlassian.net/browse/ENT-4890)

courses were being indexed as archived when they had a course run switch to a `starting soon` availability because when retrieving, the state after `upcoming` (ie starting soon) wasn't being accounted for. 

Also added a unit test for when no availability is present

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
@@ -290,6 +290,27 @@ class AlgoliaUtilsTests(TestCase):
             },
             ['Archived'],
         ),
+        (
+            {
+                'course_runs': [{
+                    'status': 'published',
+                    'is_enrollable': True,
+                    'is_marketable': True,
+                    'availability': 'Starting Soon'
+                }]
+            },
+            ['Starting Soon'],
+        ),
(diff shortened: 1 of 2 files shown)


binodpant (approved):
🚀

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -290,6 +290,27 @@ def test_get_advertised_course_run(self, searchable_course, expected_course_run)
             },
             ['Archived'],
         ),
+        (
+            {
+                'course_runs': [{
+                    'status': 'published',
+                    'is_enrollable': True,
+                    'is_marketable': True,
+                    'availability': 'starting soon'
```
adamstankiewicz: nit: Should this reflect how the string looks in the API response (i.e., `Starting Soon`)? that way the test is also asserting that it's doing the proper lowercasing before finding the value in the `COURSE_AVAILABILITY_MESSAGES` map 😄

adamstankiewicz (approved):

PR #332, opened 2021-10-07
fix: include important titles in all algolia shards

johnnagro:
This is meant to be a quick-hit fix for ENT-4980 and we'll circle back with a more "correct" fix later. The "correct" fix plan at the moment is to add a new facet to algolia which contains these "important" catalog titles for display (won't be broken up or differ between shards).

TLDR on the bug: when you query a facet which is broken up between batched/sharded records, algolia only seems to return the matching batch's version of the facet data, not the aggregated value of that facet. Fine to locate a course but not okay if your display assumes you see the aggregated value.

This quick-fix essentially keeps track of "important" catalog titles (business, education, a la carte) and ensures that every algolia batch/shard for a course always has the all the titles for "important" catalogs (not just the ones which happen to exist in that batch). I added a test which which ensures every batch contains all the important catalog titles. So, no matter which batch is returned as a result, the UI can display properly (tag courses with all the important catalogs).

- [ENT-4980](https://openedx.atlassian.net/browse/ENT-4980)

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -54,6 +54,7 @@ UNREADY_TASK_RETRY_COUNTDOWN_SECONDS = 60 * 5
 # ENT-4980 every batch "shard" record in algoilia should have all of these that pertain to the course
 IMPORTANT_CATALOG_TITLES = ['A la carte', 'Business', 'Education']
 
+
 def _fetch_courses_by_keys(course_keys):
     """
     Fetches course data from discovery's /api/v1/courses endpoint for the provided course keys.
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -51,7 +51,7 @@ ONE_HOUR = timedelta(hours=1)
 
 UNREADY_TASK_RETRY_COUNTDOWN_SECONDS = 60 * 5
 
-# ENT-4980 every batch "shard" record in algoilia should have all of these that pertain to the course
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -51,6 +51,9 @@
 
 UNREADY_TASK_RETRY_COUNTDOWN_SECONDS = 60 * 5
 
+# ENT-4980 every batch "shard" record in algoilia should have all of these that pertain to the course
+IMPORTANT_CATALOG_TITLES = ['A la carte', 'Business', 'Education']
```
kiram15: NIT: I don't know if "important_catalog_titles" sends the right message, this could just be "explore_catalog_titles", also, since you have an identical comment where it's being used, not sure if you need this one (though I love that you included a ticket number - a lot of people forget)

kiram15 (approved):
LGMT, just one small thing!

enterprise_catalog/apps/api/tasks.py
johnnagro: ah much better name - i was struggling to come up with a good name for this collection. i'll tweak.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -394,16 +397,21 @@ def _batched_metadata_with_queries(json_metadata, sorted_queries):
     Batched catalog queries are represented as tuples (<query uuid>, <query title>). Unzip the two fields and update
     them together.
     """
+
+    # ENT-4980 every batch "shard" record in algoilia should have all of these that pertain to the course
```
binodpant: typo :)

PR #333, opened 2021-10-07
Create filterable upgrade_deadline associated with course_run

kiram15:
Removal of enrollment_end (as this is already being checked) and the conversion of verified track upgrade_deadline variables unix timestamps in order to have the ability to filter them in algolia.

Algolia documentation on working with dates: https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-date/

Ticket Link
Link to the associated ticket - [ENT-4986](https://openedx.atlassian.net/browse/ENT-4986)

Post-review


enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -236,6 +236,7 @@ def test_get_course_subjects(self, course_metadata, expected_subjects):
                 'pacing_type': 'instructor_paced',
                 'start': '2013-10-16T14:00:00Z',
                 'end': '2014-10-16T14:00:00Z',
+                'upgrade_deadline': 32503680000.0,
```
johnnagro, line 239: why is this a different datetime format from the other fields? and why is it a decimal?

kiram15: It's a UNIX timestamp, it's required by algolia "to transform your date attribute into Unix timestamps, as numeric values." Our other deadlines are formatted for our use, but since we will be checking if it has passed this time, we need it to a date that is recognized by algolia.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -500,6 +504,7 @@ def get_advertised_course_run(course):
         'pacing_type': full_course_run.get('pacing_type'),
         'start': full_course_run.get('start'),
         'end': full_course_run.get('end'),
+        'upgrade_deadline': _get_verified_upgrade_deadline(full_course_run),
```
adamstankiewicz, line 507: nit: we're already calling `_get_verified_upgrade_deadline` below with:
`upgrade_deadline = _get_verified_upgrade_deadline(full_course_run)`

Could we store it in a variable and reuse it in both places?

adamstankiewicz, line 507: Related, I'm confused by the existing comment below (i.e., `# only include upgrade deadline if there is a verified seat present in the course runs`) if we're going to change up this logic to always include the `upgrade_deadline`?

If there is no upgrade deadline, is it possible to still just not include the field rather than setting a default of year 3000 or does Algolia filtering require that the field exists on all Algolia objects for filtering to work?

kiram15: From the documentation: "The Algolia engine does not support filtering on null value or missing attributes." Its weird but yes, that's why the deadline does have to be set.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -247,27 +248,6 @@ def test_get_course_subjects(self, course_metadata, expected_subjects):
...
-                'advertised_course_run_uuid': ADVERTISED_COURSE_RUN_UUID
-            },
-            {
-                'key': 'course-v1:org+course+1T2021',
-                'pacing_type': 'instructor_paced',
-                'start': '2013-10-16T14:00:00Z',
-                'end': '2014-10-16T14:00:00Z',
-                'enrollment_end': '2021-09-30T23:59:00Z',
-            }
-        ),
```
adamstankiewicz, line 270: Just curious for the rationale of removing this test case?

adamstankiewicz (approved):

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
kiram15: It was a test case I had written previously, without realizing that this is being checked before this point, so there's no way this would fail.

PR #337, opened 2021-10-13
chore: fixing a single comma that lint missed

kiram15:
## Description

A single comma that lint missed is messing up our production Algolia index. 

## Post-review


johnnagro (approved):
🤸

PR #338, opened 2021-10-13
feat: new catalog endpoint that generates a diff of a catalog's content

alex-sheehan-edx:
## Description

Diff is based on provided keys. 

context- we need a new endpoint to generate 3 buckets of data based on a diff of a catalog's content and a provided list of keys. The buckets signify which provided keys are listed under the catalog, which were not, and what contents are listed but not provided.

## Ticket Link

Link to the associated ticket
[Jira ticket](https://openedx.atlassian.net/browse/ENT-5044)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -889,6 +889,114 @@ class EnterpriseCustomerViewSetTests(APITestMixin):
             kwargs={'enterprise_uuid': enterprise_uuid or self.enterprise_uuid},
         )
 
+    def _get_generate_diff_base_url(self, enterprise_catalog_uuid=None):
+        """
+        Helper to construct the base url for the catalog `generate_diff` endpoint
+        """
+        return reverse(
+            'api:v1:generate-catalog-diff',
+            kwargs={'uuid': enterprise_catalog_uuid or self.enterprise_catalog.uuid},
+        )
+
+    def test_generate_diff_unauthorized_non_catalog_learner(self):
+        """
+        Verify the generate_diff endpoint rejects users that are not catalog learners
+        """
+        self.set_up_invalid_jwt_role()
+        self.remove_role_assignments()
+        url = self._get_generate_diff_base_url()
+        response = self.client.post(url)
+        self.assertEqual(response.status_code, status.HTTP_403_FORBIDDEN)
+
+    def test_generate_diff_get_supports_up_to_max_content_keys(self):
+        """
+        Test that GET requests to generate_diff supports up to but not more than the max number of content keys.
+        """
+        content = ContentMetadataFactory()
+        self.add_metadata_to_catalog(self.enterprise_catalog, [content])
+        url = self._get_generate_diff_base_url() + '?content_keys=key'
+
+        for key in range(150):
+            url += f"&content_keys=key{key}"
+
+        response = self.client.get(url)
+        assert response.status_code == 400
+        assert response.data == 'catalog_diff GET requests supports up to 100. If more content keys required, please ' \
+                                'use a POST body.'
+
+    def test_generate_diff_get_parses_all_buckets(self):
+        """
+        Test that GET requests to the generate_diff endpoint behave the same as POST requests.
+        """
+        content = ContentMetadataFactory()
+        content2 = ContentMetadataFactory()
+        content3 = ContentMetadataFactory()
+        content4 = ContentMetadataFactory()
+
+        self.add_metadata_to_catalog(self.enterprise_catalog, [content, content2, content3, content4])
+        url = self._get_generate_diff_base_url()
+        response = self.client.post(
+            url,
+            data={'content_keys': [content.content_key, content2.content_key, 'bad+key', 'bad+key2']}
+        )
+        assert response.status_code == 200
+        response_data = response.data
+
+        for item in response_data.get('items_not_found'):
+            assert item in [{'content_key': 'bad+key'}, {'content_key': 'bad+key2'}]
+
+        for item in response_data.get('items_not_included'):
+            assert item in [{'content_key': content3.content_key}, {'content_key': content4.content_key}]
+
+        for item in response_data.get('items_found'):
+            assert item in [
+                {'content_key': content.content_key, 'date_updated': content.modified},
+                {'content_key': content2.content_key, 'date_updated': content2.modified}
+            ]
+
+    def test_generate_diff_returns_whole_catalog_w_empty_key_list(self):
+        """
+        Test that the generate_diff endpoint will return all content keys under the catalog not provided under the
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/api/v1/views.py
```diff
@@ -164,6 +164,74 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids, *
...
+            enterprise_catalog = self.get_object()
+            return str(enterprise_catalog.enterprise_uuid)
+        return None
+
+    @action(detail=True)
+    def post(self, request, **kwargs):
+        course_keys = request.data.get('course_keys')
+        return self.catalog_diff(course_keys)
+
+    @method_decorator(require_at_least_one_query_parameter('course_keys'))
```
binodpant: not clear why we need a GET at all, but since you set this up already we can go with it!

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -889,6 +889,90 @@ def _get_contains_content_base_url(self, enterprise_uuid=None):
...
+        """
+        Test that the generate_diff endpoint will return under the `items_found` bucket all content keys within the
+        catalog that were provided.
+        """
+        content = ContentMetadataFactory()
+        self.add_metadata_to_catalog(self.enterprise_catalog, [content])
+
+        url = self._get_generate_diff_base_url()
+        response = self.client.post(url, data={'course_keys': [content.content_key]})
+        assert response.data.get('items_found')[0].get('course_key') == content.content_key
```
binodpant: think it is better to use multiple values in an array here just to be stricter for the testing

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -889,6 +889,90 @@ def _get_contains_content_base_url(self, enterprise_uuid=None):
...
+    def test_generate_diff_returns_content_items_not_found(self):
+        """
+        Test that the generate_diff endpoint will return all content keys provided that were not found under the catalog
+        under the `items_not_found` bucket.
+        """
+        key = 'bad+key'
+        key2 = 'bad+key2'
+        url = self._get_generate_diff_base_url()
+        response = self.client.post(url, data={'course_keys': [key, key2]})
+        assert response.data.get('items_not_found')[0] == {key, key2}
```
binodpant: can we also test the other return values here items_found and items_not_included (will they be empty?) actually would do that for all tests for more complete testing

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -889,6 +889,90 @@ def _get_contains_content_base_url(self, enterprise_uuid=None):
...
+        Test that GET requests to the generate_diff endpoint behave the same as POST requests.
+        """
+        content = ContentMetadataFactory()
+        content2 = ContentMetadataFactory()
+        self.add_metadata_to_catalog(self.enterprise_catalog, [content, content2])
+        url = self._get_generate_diff_base_url()
+        response = self.client.post(url, data={'course_keys': [content.content_key, 'bad+key']})
+        assert response.status_code == 200
+        response_data = response.data
+        assert response_data.get('items_not_found')[0] == {'bad+key'}
```
binodpant: would recommend testing with more than one value here for each category

enterprise_catalog/apps/api/v1/views.py
alex-sheehan-edx: to allow for flexibility with consumers, not totally necessary but one or two people expressed interest so I added

binodpant (approved):
gtg!

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -164,6 +164,77 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids, *
...
+            if len(content_keys) > self.MAX_GET_CONTENT_KEYS:
+                return Response(
+                    f'catalog_diff GET requests supports up to {self.MAX_GET_CONTENT_KEYS}. If more content keys '
+                    f'required, please use a POST body.',
+                    status=HTTP_400_BAD_REQUEST
+                )
+
+        return self.catalog_diff(content_keys)
+
+    def catalog_diff(self, course_keys):
```
adamstankiewicz: nit: `content_keys` vs. `course_keys` here and elsewhere in docstrings, etc.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -164,6 +164,77 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids, *
...
+        if self.kwargs.get('uuid'):
+            enterprise_catalog = self.get_object()
+            return str(enterprise_catalog.enterprise_uuid)
+        return None
+
+    @action(detail=True)
+    def post(self, request, **kwargs):
+        course_keys = []
+        if request.data:
+            course_keys = request.data.getlist('content_keys')
```
adamstankiewicz: nit: "content_keys"

enterprise_catalog/apps/catalog/models.py
```diff
@@ -203,6 +203,50 @@ def enterprise_customer(self):
         """
         return EnterpriseCustomerDetails(self.enterprise_uuid)
 
+    def get_catalog_content_diff(self, content_keys):
+        """
+        Generate a catalog diff based on a provided list of content keys and what currently exists with the catalog's
+        metadata.
+
+        Returns:
```
adamstankiewicz, line 214: nit: Maybe also include "Arguments" to describe the `content_keys` arg?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -203,6 +203,50 @@ def enterprise_customer(self):
...
+        items_not_included = []
+        items_found = []
+        items_not_found = set()
+
+        # cannot determine if specified content keys are part of catalog when catalog
+        # query doesn't exist.
+        if not self.catalog_query:
+            return [items_not_found], items_not_included, items_found
+
+        content_keys = set(content_keys)
```
adamstankiewicz: nit: might recommend remaining `content_keys` variable to `distinct_content_keys` or similar so that we're not reusing the same name as the method argument.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -203,6 +203,50 @@ def enterprise_customer(self):
...
+        for content in self.content_metadata.all().values('modified', 'content_key'):
+            found_content_keys.add(content.get('content_key'))
```
adamstankiewicz: nit: `content.get('content_key')` is re-used in several places here. Maybe create a variable to store this value without needing to call `.get` several times.

adamstankiewicz (approved):
Just a few minor nits, otherwise looks good to me!

PR #341, opened 2021-10-25
feat: fixed content last updated and added content metadata content keys filtering

alex-sheehan-edx:
Fast follow to https://github.com/edx/enterprise-catalog/pull/338 where now we can be more specific with what we request from the catalog service + no longer not detecting when enrollment URL's have changed.

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/serializers.py
+++ b/enterprise_catalog/apps/api/v1/serializers.py
@@ -6,6 +6,7 @@ from rest_framework import serializers, status
 
 from enterprise_catalog.apps.api.v1.utils import (
     get_enterprise_utm_context,
+    get_most_recent_modified_time,
     is_any_course_run_active,
     update_query_parameters,
 )
@@ -191,7 +192,16 @@ class ContentMetadataSerializer(ImmutableStateSerializer):
         content_key = json_metadata.get('key')
         parent_content_key = get_parent_content_key(json_metadata)
 
-        json_metadata['content_last_modified'] = instance.modified
+        # The enrollment URL field of content metadata is generated on request and is determined by the status of the
+        # enterprise customer as well as the catalog. So, in order to detect when content metadata has last been
+        # modified, we have to also check the customer and the catalog's modified times.
+        modified_time = get_most_recent_modified_time(
+            instance.modified,
+            enterprise_catalog.modified,
+            enterprise_catalog.enterprise_customer.last_modified_date
+        )
+
+        json_metadata['content_last_modified'] = modified_time
 
         if marketing_url:
             marketing_url = update_query_parameters(
--- a/enterprise_catalog/apps/api/v1/tests/test_utils.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_utils.py
@@ -1,0 +1,57 @@
+from datetime import timedelta
+
+from django.test import TestCase
+
+from enterprise_catalog.apps.api.v1.utils import get_most_recent_modified_time
+from enterprise_catalog.apps.catalog.utils import localized_utcnow
+
+
+class ApiUtilsTests(TestCase):
+    """
+    Tests for the Enterprise Catalog API client utils
+    """
+
+    def test_get_most_recent_modified_time_catalog_modified(self):
+        """
+        Test that the get_most_recent_modified_time function will account for catalog modified times
+        """
+        now = localized_utcnow()
+        catalog_time = now
+        customer_time = now - timedelta(hours=1)
+        content_time = now - timedelta(hours=1)
+        most_recent_time = get_most_recent_modified_time(content_time, catalog_time, customer_time)
+        assert most_recent_time == catalog_time
+
+    def test_get_most_recent_modified_time_customer_modified(self):
+        """
+        Test that the get_most_recent_modified_time function will account for customer modified times
+        """
+        now = localized_utcnow()
+        customer_time = now
+        catalog_time = now - timedelta(hours=1)
+        content_time = now - timedelta(hours=1)
+        most_recent_time = get_most_recent_modified_time(content_time, catalog_time, customer_time)
+        assert most_recent_time == customer_time
+
+    def test_get_most_recent_modified_time_content_modified(self):
+        """
+        Test that the get_most_recent_modified_time function will account for content modified times
+        """
+        now = localized_utcnow()
+        content_time = now
+        catalog_time = now - timedelta(hours=1)
+        customer_time = now - timedelta(hours=1)
+        most_recent_time = get_most_recent_modified_time(content_time, catalog_time, customer_time)
+        assert most_recent_time == content_time
+
+    def test_get_most_recent_modified_time_accounts_for_no_customer_modified(self):
+        """
+        Test that the get_most_recent_modified_time function will account for content modified times when customer
+        modified times do not exist
+        """
+        now = localized_utcnow()
+        content_time = now
+        catalog_time = now - timedelta(hours=1)
+        customer_time = None
+        most_recent_time = get_most_recent_modified_time(content_time, catalog_time, customer_time)
+        assert most_recent_time == content_time
--- a/enterprise_catalog/apps/api/v1/urls.py
+++ b/enterprise_catalog/apps/api/v1/urls.py
@@ -12,10 +12,14 @@ app_name = 'v1'
 router = DefaultRouter()
 router.register(r'enterprise-catalogs', views.EnterpriseCatalogCRUDViewSet, basename='enterprise-catalog')
 router.register(r'enterprise-catalogs', views.EnterpriseCatalogContainsContentItems, basename='enterprise-catalog')
-router.register(r'enterprise-catalogs', views.EnterpriseCatalogGetContentMetadata, basename='enterprise-catalog')
 router.register(r'enterprise-customer', views.EnterpriseCustomerViewSet, basename='enterprise-customer')
 
 urlpatterns = [
+    url(
+        r'^enterprise-catalogs/(?P<uuid>[\S]+)/get_content_metadata',
+        views.EnterpriseCatalogGetContentMetadata.as_view({'get': 'get'}),
+        name='get-content-metadata'
+    ),
     url(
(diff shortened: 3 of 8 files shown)


enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -191,7 +192,14 @@ def to_representation(self, instance):
...
 
-        json_metadata['content_last_modified'] = instance.modified
+        modified_time = max(
+            [enterprise_catalog.modified, instance.modified]
+        )
+        customer_modified = enterprise_catalog.enterprise_customer.last_modified_date
+        if customer_modified:
+            modified_time = max([modified_time, parser.parse(customer_modified)])
+
+        json_metadata['content_last_modified'] = modified_time
```
adamstankiewicz: Beautiful 👨‍🍳 💋

johnnagro (approved):
lgtm ⛵  probably should wait for @binodpant to chime in too

enterprise_catalog/apps/api/v1/urls.py
```diff
@@ -12,10 +12,14 @@
 router = DefaultRouter()
 router.register(r'enterprise-catalogs', views.EnterpriseCatalogCRUDViewSet, basename='enterprise-catalog')
 router.register(r'enterprise-catalogs', views.EnterpriseCatalogContainsContentItems, basename='enterprise-catalog')
-router.register(r'enterprise-catalogs', views.EnterpriseCatalogGetContentMetadata, basename='enterprise-catalog')
 router.register(r'enterprise-customer', views.EnterpriseCustomerViewSet, basename='enterprise-customer')
 
 urlpatterns = [
+    url(
+        r'^enterprise-catalogs/(?P<uuid>[\S]+)/get_content_metadata',
+        views.EnterpriseCatalogGetContentMetadata.as_view({'post': 'post'}),
```
adamstankiewicz: [curious] for my own learning, why do we need to include a `post` action here, especially if the view is doing both a get and a post? Curious if this is only really needed when mapping a custom method name vs. using the standard "def get" and "def post". 🤔

example: `views.DistinctCatalogQueriesView.as_view()` below doesn't provide any arguments.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -284,18 +290,45 @@ def get_response_with_enterprise_fields(self, response):
...
+
+    @action(detail=True)
+    def get(self, request, **kwargs):
+        content_keys_filter = request.query_params.getlist('content_keys')
+        if content_keys_filter == "[]":
+            content_keys_filter = []
+        else:
+            if len(content_keys_filter) > self.MAX_GET_CONTENT_KEYS:
+                return Response(
+                    f'get_content_metadata GET requests supports up to {self.MAX_GET_CONTENT_KEYS}. If more content'
```
mattdrayer, line 307: Yay this was my idea!  :)

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -191,7 +192,14 @@ def to_representation(self, instance):
         content_key = json_metadata.get('key')
         parent_content_key = get_parent_content_key(json_metadata)
 
-        json_metadata['content_last_modified'] = instance.modified
+        modified_time = max(
```
binodpant: maybe drop a quick comment here or in docstring around what this modified timestamp represents (esp since now it's a bit more nuanced than it used to be)

enterprise_catalog/apps/catalog/models.py
```diff
@@ -239,10 +240,15 @@ def get_catalog_content_diff(self, content_keys):
         for content in self.content_metadata.all().values('modified', 'content_key'):
             content_key = content.get('content_key')
             found_content_keys.add(content_key)
+            content_modified = max([content.get('modified'), self.modified])
```
binodpant: just a thought, since this logic is also present in the serializer, is it possible to turn this into a uti function, test that just unit level, then use it in both there and here? https://github.com/edx/enterprise-catalog/pull/341/files#diff-9073e49d0f9ea982a004c58417de0420d7fe54d1b31b55c1b7b73235361e240aR200

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -260,11 +261,16 @@ def get_permission_object(self):
...
         return str(self.enterprise_catalog.enterprise_uuid)
 
-    def get_queryset(self):
+    def get_queryset(self, content_keys_filter=None):
         """
         Returns all of the json of content metadata associated with the catalog.
         """
         # Avoids ordering the content metadata by any field on that model to avoid using a temporary table / filesort
+        if content_keys_filter:
+            return self.enterprise_catalog.content_metadata.filter(
```
binodpant: nit: may it be slightly more consistent with other places if you just modify the queryset here and continue returning, in the code below? but will leave it upto you right now we are not doing any other filtering, so not a big deal

like 

```
if content_keys_filter:
    queryset = queryset.filter()

return queryset.....
```

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -284,18 +290,45 @@ def get_response_with_enterprise_fields(self, response):
         return response
 
     @action(detail=True)
-    def get_content_metadata(self, request, uuid, **kwargs):  # pylint: disable=unused-argument
+    def post(self, request, **kwargs):
+        content_keys_filter = []
```
binodpant: docstring needed?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -284,18 +290,45 @@ def get_response_with_enterprise_fields(self, response):
...
+        traverse_pagination = False
+        if request.data:
+            content_keys_filter = request.data.get('content_keys')
+            traverse_pagination = request.data.get('traverse_pagination', False)
+
+        return self.get_content_metadata(request, traverse_pagination, content_keys_filter)
+
+    @action(detail=True)
+    def get(self, request, **kwargs):
+        content_keys_filter = request.query_params.getlist('content_keys')
```
binodpant, line 301: docstring? 
how is this method different from get_content_metadata, sorry I forgot but looks like this one only returns the content keys, the other is for metadata, correct? may help to clarify in docstring

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -284,18 +290,45 @@ def get_response_with_enterprise_fields(self, response):
...
+                    f'keys required, please use a POST body.',
+                    status=HTTP_400_BAD_REQUEST
+                )
+
+        traverse_pagination = request.query_params.get('traverse_pagination', False)
+
+        return self.get_content_metadata(request, traverse_pagination, content_keys_filter)
+
+    @action(detail=True)
+    def get_content_metadata(self, request, traverse_pagination, content_keys_filter):  # pylint: disable=unused-argument
```
binodpant: can content_key_filter even be None for this method? is it always required? looks like previously we used kwargs but wren't really using them, interesting, will this change how this method is used in the browser like we use today? like `/catalogid/get_content_metadata` ?

alex-sheehan-edx: it can't be None, but if it's not provided from the user through the API then `content_kerys_filter` will just be an empty list

enterprise_catalog/apps/catalog/models.py
alex-sheehan-edx: hmmm, I'm mixed on this because yes it would be great to have a testable util function but also really it's just a `max()` call which doesn't scream needing a util function for. But let's do it!

enterprise_catalog/apps/api/v1/urls.py
alex-sheehan-edx: Hmm what I can say is that `DistinctCatalogQueriesView` seems to be getting it's positional `actions` argument from somewhere else because when I add `{'post': 'post'}` I get 

```
E   TypeError: as_view() takes 1 positional argument but 2 were given
```

However, when removing the `{'post': 'post'}`from `get_content_metadata` or `generate_diff` `as_view` calls, we get the opposite... Still not 100% sure what's going

enterprise_catalog/apps/api/v1/tests/test_utils.py
```diff
@@ -0,0 +1,56 @@
+from datetime import datetime, timedelta
+
+import pytz
+from django.test import TestCase
+
+from enterprise_catalog.apps.api.v1.utils import get_most_recent_modified_time
+
+
+class ApiUtilsTests(TestCase):
```
binodpant, line 9: this is lovely! thanks for putting this together

enterprise_catalog/apps/api/v1/tests/test_utils.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Tests for the EnterpriseCatalogCRUDViewSet
```
binodpant: comment needs fixup

binodpant (approved):
lgtm!

adamstankiewicz (approved):

enterprise_catalog/apps/api/v1/tests/test_utils.py
```diff
@@ -0,0 +1,56 @@
...
+class ApiUtilsTests(TestCase):
+    """
+    Tests for the Enterprise Catalog API client utils
+    """
+
+    def test_get_most_recent_modified_time_catalog_modified(self):
+        """
+        Test that the get_most_recent_modified_time function will account for catalog modified times
+        """
+        now = datetime.now().replace(tzinfo=pytz.UTC)
```
adamstankiewicz: Might use the existing [`localized_utcnow`](https://github.com/edx/enterprise-catalog/blob/0784811c71070e7432c5129e9d8ffd5e39511c4c/enterprise_catalog/apps/catalog/utils.py#L98) util method to get localized datetimes

adamstankiewicz (approved):
LGTM. Just one suggestion around localized datetimes 😄

PR #343, opened 2021-11-04
feat: ENT-5119 Added additional fields in Algolia index for Programs

irfanuddinahmad:
## Description

This PR adds additional fields in the Algolia index for Programs. The additional fields are Availability, Partners, Subjects, Skills, Level Types

## Ticket Link

[ENT-5119](https://openedx.atlassian.net/browse/ENT-5119)

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -3,6 +3,8 @@ import datetime
 import logging
 import time
 
+from django.utils.translation import ugettext as _
+
 from enterprise_catalog.apps.api.v1.utils import is_course_run_active
 from enterprise_catalog.apps.api_client.algolia import AlgoliaSearchClient
 from enterprise_catalog.apps.catalog.constants import (
@@ -349,14 +351,14 @@ def _get_program_course_field(program, field):
 
 def get_program_course_keys(program):
     """
-       Gets list of course keys associated with the program.
+    Gets list of course keys associated with the program.
 
-       Arguments:
-           program (dict): a dictionary representing a program.
+    Arguments:
+       program (dict): a dictionary representing a program.
 
-       Returns:
-           list: a list of course keys associated with the program.
-       """
+    Returns:
+       list: a list of course keys associated with the program.
+    """
     return _get_program_course_field(program, 'key')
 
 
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -374,6 +376,114 @@ def get_program_type(program):
...
+
+    Arguments:
+        program (dict): a dictionary representing a program.
+
+    Returns:
+        list: a union of program courses availability.
+    """
+    # Master's programs don't have courses in the same way that our other programs do.
+    program_type = program.get('type')
+    if program_type and program_type.get('slug') == 'masters':
```
HammadAhmadWaqas: I think `program_type` will be string, So is this  `program_type.get('slug') == 'masters'` expression right?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -374,6 +376,114 @@ def get_program_type(program):
...
+        return _('Available now')
+
+    availability = set()
+    for course in program.get('courses', []):
+        course_status = get_course_availability(course)
+        for status in course_status:
+            availability.add(status)
+    return list(availability)
```
HammadAhmadWaqas, line 412: Line 405 (return _('Available now')) returning a single item and here we are returning a list, 
Should not availability a single item always?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -627,7 +737,14 @@ def _algolia_object_from_product(product, algolia_fields):
     elif searchable_product.get('content_type') == PROGRAM:
         searchable_product.update({
             'course_keys': get_program_course_keys(searchable_product),
+            'programs': list(get_program_type(searchable_product)),
```
HammadAhmadWaqas: Why we are adding programs here? we have an existing key `program_type` to represent the type of the program.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -444,3 +451,206 @@ def test_configure_algolia_index(self, mock_search_client):
...
+                    'status': 'published',
+                    'is_enrollable': True,
+                    'is_marketable': True,
+                }]
+            }]},
+            ['Archived'],
+        ),
+    )
+    @ddt.unpack
+    def test_get_program_availability(self, program_metadata, expected_availability):
```
HammadAhmadWaqas, line 568: Are we checking availability for `Master` programs? as it is mentioned they are a little different!

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -374,6 +376,114 @@ def get_program_type(program):
...
+
+    Returns:
+        str: level type associated with the program.
+    """
+    level_types = []
+    for course in program.get('courses', []):
+        course_level_type = course.get('level_type')
+        if course_level_type:
+            level_types.append(course_level_type)
+    return max(set(level_types), key=level_types.count) if level_types else ''
```
HammadAhmadWaqas, line 484: why we have used `set` here? 
can we use default param to empty value like: `max(level_types, default='', key=level_types.count)` ?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -627,7 +737,14 @@ def _algolia_object_from_product(product, algolia_fields):
     elif searchable_product.get('content_type') == PROGRAM:
         searchable_product.update({
             'course_keys': get_program_course_keys(searchable_product),
+            'programs': list(get_program_type(searchable_product)),
+            'program_titles': list(get_program_title(searchable_product)),
```
HammadAhmadWaqas: why we are typecasting to list? program title should not be a singular value?

HammadAhmadWaqas (changes requested):

enterprise_catalog/apps/catalog/algolia_utils.py
irfanuddinahmad: Fixed.

enterprise_catalog/apps/catalog/algolia_utils.py
irfanuddinahmad: This is conforming to the course availability ... which is a list

enterprise_catalog/apps/catalog/algolia_utils.py
irfanuddinahmad: We want to use the same facet for courses and programs. This already exists for courses.

enterprise_catalog/apps/catalog/algolia_utils.py
irfanuddinahmad: We want to the find the "mode" of the list.

enterprise_catalog/apps/catalog/algolia_utils.py
irfanuddinahmad: We want to use the same field as is used for courses.

HammadAhmadWaqas (approved):
LGTM

muhammad-ammar (approved):
LGTM

PR #344, opened 2021-11-10
feat: Recording when queries delete content

alex-sheehan-edx:
## Description

Soft deletion is desirable from two standpoints-
1) we have a [customer feature request](https://openedx.atlassian.net/browse/ENT-4907) that would require us to track times of content metadata removal from catalog queries
2) Soft deletions are a general practice that we want to start adhering to and have started to implement in places like the [Integrated Channels](https://github.com/edx/edx-enterprise/pull/1384/files#diff-8e527cbda17b71cd26cccce8c4c077fbd4a731dc04ac83a1601a7efc57b4d451R311-R315) 

## Post-review


enterprise_catalog/apps/catalog/migrations/0028_auto_20211109_2347.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('catalog', '0027_contentmetadatatoqueries'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
+            model_name='contentmetadata',
+            name='catalog_queries',
+        ),
```
iloveagent57: you'll want to do this in a separate deployment after you deploy the code below that stops _using_ this field.

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -911,6 +912,7 @@ def setUp(self):
     def tearDown(self):
         super().tearDown()
         # clean up any stale test objects
+        ContentMetadataToQueries.all_objects.all().hard_delete()
```
alex-sheehan-edx, line 915: Ideally I can get this to where this table cascade deletes and we don't need to do this. Tips and advice appreciated!

enterprise_catalog/apps/catalog/models.py
```diff
@@ -414,7 +426,7 @@ class ContentMetadata(TimeStampedModel):
             "endpoint results, specified as a JSON object."
         )
     )
-    catalog_queries = models.ManyToManyField(CatalogQuery)
+    catalog_query_mapping = models.ManyToManyField(CatalogQuery, through='ContentMetadataToQueries')
```
iloveagent57, line 433: Would some of this become simpler if the ManyToManyField was on the CatalogQuery model?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -445,6 +465,70 @@ def __str__(self):
...
+
+class ContentMetadataToQueries(SoftDeletionModel):
+    catalog_query = models.ForeignKey(
+        CatalogQuery,
+        on_delete=models.deletion.DO_NOTHING,
+    )
+    content_metadata = models.ForeignKey(
+        ContentMetadata,
+        on_delete=models.deletion.DO_NOTHING,
+    )
```
iloveagent57, line 545: I think you can use `related_name` here to make some of the code above read a little easier.

iloveagent57 (commented):
Still reading through this, but a couple brief comments for now.

enterprise_catalog/apps/catalog/migrations/0028_auto_20211109_2347.py
alex-sheehan-edx: Good thinking!

enterprise_catalog/apps/catalog/models.py
alex-sheehan-edx: I'm not sure, I think it's the same but happy to hear other perspectives

johnnagro (approved):

enterprise_catalog/apps/catalog/migrations/0027_contentmetadatatoqueries.py
```diff
@@ -0,0 +1,44 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+import django.utils.timezone
+import model_utils.fields
+
+
+def backfill_many_to_many(apps, schema_editor):
+    content_metadata_model = apps.get_model("catalog", "ContentMetadata")
+    catalog_query_model = apps.get_model("catalog", "CatalogQuery")
```
iloveagent57: Not a blocker, but feel empowered to write these as
```
ContentMetadata = apps.get_model("catalog", "ContentMetadata")
```

enterprise_catalog/apps/catalog/models.py
```diff
@@ -414,7 +426,7 @@ class ContentMetadata(TimeStampedModel):
             "endpoint results, specified as a JSON object."
         )
     )
-    catalog_queries = models.ManyToManyField(CatalogQuery)
```
iloveagent57, line 421: Will the backfill migration work if this field isn't there?

enterprise_catalog/apps/catalog/migrations/0027_contentmetadatatoqueries.py
```diff
@@ -0,0 +1,44 @@
...
+    query_content_metadata_many_to_many = content_metadata_model.catalog_queries.through.objects.all()
```
iloveagent57: Keep in mind this is about 1 million records in prod - we might have to break this up into chunks (which isn't too hard with this table since it has an integer PK).

iloveagent57 (commented):
couple more thoughts/questions about the backfill.

enterprise_catalog/apps/catalog/migrations/0027_contentmetadatatoqueries.py
```diff
@@ -0,0 +1,44 @@
...
+    content_metadata_to_queries_model = apps.get_model("catalog", "ContentMetadataToQueries")
+    for row in query_content_metadata_many_to_many:
+        query_object = catalog_query_model.objects.filter(id=row.catalogquery_id).first()
+        content_metadata_object = content_metadata_model.objects.filter(id=row.contentmetadata_id).first()
+        content_metadata_to_queries_model.objects.update_or_create(
+            catalog_query=query_object,
+            content_metadata=content_metadata_object
+        )
```
iloveagent57: Also, maybe this is a case where writing raw SQL is ok?  All you want here is basically
```
INSERT INTO new_table (col_1, col_2) SELECT col_1, col_2 FROM old_table;
```
And then you don't have to worry about doing anything in batches, or bulk creates, etc.

enterprise_catalog/apps/catalog/models.py
alex-sheehan-edx: I believe the SQL will work (tested) because it's not dealing with the content metadata table but the through table between the two.

johnnagro (approved):
⛵  looks good. lets wait until after thanksgiving break to merge and deploy it.

enterprise_catalog/apps/catalog/migrations/0027_contentmetadatatoqueries.py
alex-sheehan-edx: done!

PR #347, opened 2021-11-29
feat: adding upcoming course runs to algolia

kiram15:
## Description
This will add in a variable to Algolia that allows us to see the number of upcoming course runs that will be shown in explore catalog's modal subheader. 

## Ticket Link

[ENT-5086](https://openedx.atlassian.net/browse/ENT-5086)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -655,7 +655,7 @@ def get_upcoming_course_runs(course):
         course (dict)
 
     Returns:
-        str: the number of course runs in the future
+        int: the number of course runs in the future
     """
     course_runs = course.get('course_runs') or []
     active_course_runs = [run for run in course_runs if is_course_run_active(run)]
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -51,6 +51,7 @@ ALGOLIA_FIELDS = [
     'title',
     'type',
     'advertised_course_run',  # a part of the advertised course run
+    'upcoming_course_runs',
     'first_enrollable_paid_seat_price',
     'original_image_url',
     'marketing_url',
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -646,6 +647,23 @@ def get_advertised_course_run(course):
...
 
+def get_upcoming_course_runs(course):
+    """
+    Get number of upcoming course runs.
+
+    Argument:
+        course (dict)
+
+    Returns:
+        str: the number of course runs in the future
```
binodpant: should this type be int or number instead of 'str' ?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -646,6 +647,23 @@ def get_advertised_course_run(course):
...
+    """
+    course_runs = course.get('course_runs') or []
+    active_course_runs = [run for run in course_runs if is_course_run_active(run)]
```
binodpant, line 661: upto you  but a slight alternative (a bit shorter) would be to edit the condition in the for comprehension to exclude if it matches the advertised_course_run uuid ! it's also fine the way it is!

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -646,6 +647,23 @@ def get_advertised_course_run(course):
     return course_run
 
 
+def get_upcoming_course_runs(course):
```
binodpant, line 650: also I am not sure myself of the correct answer, but let's confirm if we got it accurate, did you check with tNL team or discovery team on if this is the most correct way to determine 'upcoming'? e.g., can one of these active courses, have a `availability` that is something other than 'Starting soon' ? do we count those too?

binodpant (approved):
I am good as long as we are confident in the correct def of 'upcoming'

enterprise_catalog/apps/catalog/algolia_utils.py
kiram15: Yes this includes both upcoming and starting soon! Active course runs are described as "published, enrollable, and marketable" so I think we're good.

PR #358, opened 2021-12-06
maint: redeploying soft deletion properly

alex-sheehan-edx:
## Description

Previously we attempted to [ship a migration to add soft deletion](https://github.com/edx/enterprise-catalog/pull/344) to the catalog query/content metadata through table. This resulted in an unacceptably high DB load due to the exclusion of a proper db index on the new table. The [follow up deployment to add the index](https://open.spotify.com/track/509TenzZknOYbImLdT2LMu?si=6a828538ddc24f8d) was unable to be applied in prod since the table was large (~1-1.5 million + rows) and the locking migration was taking an unacceptable amount of time. Now that everything has been reverted and cleaned up, this is a release of soft deletion but with the index created **_before_** we migrate data.    

## Post-review

###### pls don't break everything again


enterprise_catalog/apps/catalog/models.py
```diff
@@ -117,6 +118,17 @@ def __str__(self):
             )
         )
 
+    @property
+    def content_metadata(self):
+        content_metadata_to_queries = self.contentmetadata_set.through.objects.filter(
+            catalog_query_id=self.id
```
alex-sheehan-edx, line 124: open question: if this is querying for `catalog_query_id`, will it be slower than the queries for content metadata due to the order of the index?

enterprise_catalog/apps/catalog/migrations/0028_contentmetadatatoqueries.py
```diff
@@ -0,0 +1,56 @@
...
+                ('deleted_at', models.DateTimeField(blank=True, null=True)),
+                ('content_metadata', models.ForeignKey(on_delete=django.db.models.deletion.DO_NOTHING, to='catalog.contentmetadata')),
+                ('catalog_query', models.ForeignKey(on_delete=django.db.models.deletion.DO_NOTHING, to='catalog.catalogquery')),
+            ],
+        ),
+        # Note- this needs to be created BEFORE we migration rows from the old through table. Additionally, the DB has
+        # issue with unapplying the unique constraint. Since the table will be dropped if this migration is reversed,
+        # we don't need to worry about removing the constraint, that's why this is written as raw sql (the add
+        # constraint method does not have a reverse sql param) and why the reverse sql is simply a noop.
+        migrations.RunSQL(
```
christopappas: I'm not remembering based on our previous conversations, but was there a reason we can't generate/express this migration from native django model makemigrations?

christopappas: I guess like, did we have reason to believe django is going to do the wrong thing?

I ask, because if we're adding a required table-drop to roll things back, in the future, if someone in the future wants to reset their DB (thinking local development or even outside of edX), that means we necessitate some manual action instead of just letting django go up and down the migration files. I don't love needing to do this (provided i'm understanding the implications correctly)

enterprise_catalog/apps/catalog/migrations/0028_contentmetadatatoqueries.py
```diff
@@ -0,0 +1,56 @@
...
+        migrations.AddField(
+            model_name='contentmetadata',
+            name='catalog_query_mapping',
+            field=models.ManyToManyField(through='catalog.ContentMetadataToQueries', to='catalog.CatalogQuery'),
+        ),
+        migrations.RunSQL(
+            sql="""
+                /* there's some funky stuff with defaults and the timestamped model */
+                INSERT INTO catalog_contentmetadatatoqueries (catalog_query_id, content_metadata_id, created, modified)
+                SELECT catalogquery_id, contentmetadata_id, CURRENT_TIMESTAMP, CURRENT_TIMESTAMP FROM catalog_contentmetadata_catalog_queries;
```
iloveagent57, line 37: Why not copy `created` from the original table, too?

enterprise_catalog/apps/catalog/models.py
iloveagent57: You should end up with two indices here - one from the FK on `catalog_query_id`, and the other from the unique constraint, which has `content_metadata_id` as its first column.  The FK index should make this query reasonably performant.

enterprise_catalog/apps/catalog/migrations/0028_contentmetadatatoqueries.py
iloveagent57: I think the reason was empirical - we tried to rollback a migration that only added the unique constraint, and Django barfed for reasons still not understood.

The "table-drop to roll things back" is the default reverse operation of the `migrations.CreateModel()` operation above - we should be able to go up and down these migrations using only Django `migrate` commands.

Dropping a table should drop all associated indices on the table, too (unless some other index on another table depends on them, which is atypical and not the case here).

iloveagent57 (commented):
I think this is good to go, but I want to check it out locally and inspect the resulting indices first (having network and devstack issues yesterday/this morning, so I'm only now in a position to do this).

enterprise_catalog/apps/catalog/migrations/0028_contentmetadatatoqueries.py
alex-sheehan-edx: In the first iteration we used the native Django method. When trying to reverse the original `AddConstraint` method, we were unable to roll it back and there was really no way to unconfuse it. Using the `RunSQL` method allows us to copy the exact SQL that the native method is generating under the hood while also specifying a downwards migration query (a noop here because the table would be dropped regardless if this migration was downgraded)

alex-sheehan-edx: >if we're adding a required table-drop to roll things back

The table drop should be the reverse when the `CreateModel` Django method is rolled back, no? The user shouldn't have to manually drop any table

alex-sheehan-edx: The intention of this `RunSQL` usage is so that we _**can**_ automatically go up and down past this migration, we just don't drop the unique constraint before we drop the table entirely. (The native Django `AddConstraint` would require the unique constraint be dropped before proceeding with a reverse migration)

enterprise_catalog/apps/catalog/migrations/0028_contentmetadatatoqueries.py
alex-sheehan-edx: oooh excellent point

iloveagent57:
OK, I was able to apply/unapply/reapply these migrations locally.  Here's the create table for the new M2M table:
```
CREATE TABLE `catalog_contentmetadatatoqueries` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `created` datetime(6) NOT NULL,
  `modified` datetime(6) NOT NULL,
  `deleted_at` datetime(6) DEFAULT NULL,
  `content_metadata_id` int(11) NOT NULL,
  `catalog_query_id` int(11) NOT NULL,
  PRIMARY KEY (`id`),
  UNIQUE KEY `unique content metadata to query` (`content_metadata_id`,`catalog_query_id`,`deleted_at`),
  KEY `catalog_contentmetad_catalog_query_id_4dc04c03_fk_catalog_c` (`catalog_query_id`),
  CONSTRAINT `catalog_contentmetad_catalog_query_id_4dc04c03_fk_catalog_c` FOREIGN KEY (`catalog_query_id`) REFERENCES `catalog_catalogquery` (`id`),
  CONSTRAINT `catalog_contentmetad_content_metadata_id_3f5c7251_fk_catalog_c` FOREIGN KEY (`content_metadata_id`) REFERENCES `catalog_contentmetadata` (`id`)
)
```
my `update_content_metadata` task finished successfully, but then I got this running `reindex_algolia`:
```
django.db.utils.ProgrammingError: (1146, "Table 'enterprise_catalog.catalog_contentmetadata_catalog_query_mapping' doesn't exist")
```

enterprise_catalog/apps/catalog/migrations/0028_contentmetadatatoqueries.py
christopappas: cool, sounds like you're all very sure about this, which is good enough for me. just wanted to make absolutely sure we wanted custom SQL here, because typically less custom stuff >>> custom stuff from a "what's going on here 1 year from now" point of view :)

iloveagent57:
Here's a more helpful part of the stacktrace:
```
enterprise.catalog.worker |   File "/edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/apps/api/tasks.py", line 601, in _reindex_algolia
enterprise.catalog.worker |     index_content_keys_in_algolia(
enterprise.catalog.worker |   File "/edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/apps/api/tasks.py", line 500, in index_content_keys_in_algolia
enterprise.catalog.worker |     for metadata in content_metadata:
```
It pertains to this chunk of code in tasks.py:
```python
        content_metadata = ContentMetadata.objects.filter(query).prefetch_related(
            Prefetch('catalog_query_mapping', queryset=catalog_queries),
            Prefetch('associated_content_metadata', queryset=associate_programs_query, to_attr='associate_programs'),
        )

        # iterate through ContentMetadata records, retrieving the enterprise_catalog_uuids
        # and enterprise_customer_uuids associated with each ContentMetadata record (either
        # a course or a course run or program), storing them in a dictionary with the related
        # content_key as a key for later retrieval. the content_key is determined by
        # the content_key field if the metadata is a `COURSE` or `PROGRAM` or by the parent_content_key
        # field if the metadata is a `COURSE_RUN`.
        for metadata in content_metadata:
```

enterprise_catalog/apps/catalog/migrations/0028_contentmetadatatoqueries.py
alex-sheehan-edx: actually- diving into this, it doesn't seem like we have a `created` value on the original through table? So I'm not sure if this is possible.

docs/decisions/0004-adding-soft-delete-to-query-metadata-relationships.rst
```diff
@@ -0,0 +1,25 @@
...
+
+Status
+------
+
+Accepted
+
+Context
+-------
+
+There has been a customer feature request that requires has required us to start capturing the times at which content is dropped from catalog queries. This has lead us to designing a soft deletion manager for models.
```
binodpant: nit: `This has led us...`

johnnagro (approved):
🙌

alex-sheehan-edx:
> Here's a more helpful part of the stacktrace:
> 
> ```
> enterprise.catalog.worker |   File "/edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/apps/api/tasks.py", line 601, in _reindex_algolia
> enterprise.catalog.worker |     index_content_keys_in_algolia(
> enterprise.catalog.worker |   File "/edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/apps/api/tasks.py", line 500, in index_content_keys_in_algolia
> enterprise.catalog.worker |     for metadata in content_metadata:
> ```
> 
> It pertains to this chunk of code in tasks.py:
> 
> ```python
>         content_metadata = ContentMetadata.objects.filter(query).prefetch_related(
>             Prefetch('catalog_query_mapping', queryset=catalog_queries),
>             Prefetch('associated_content_metadata', queryset=associate_programs_query, to_attr='associate_programs'),
>         )
> 
>         # iterate through ContentMetadata records, retrieving the enterprise_catalog_uuids
>         # and enterprise_customer_uuids associated with each ContentMetadata record (either
>         # a course or a course run or program), storing them in a dictionary with the related
>         # content_key as a key for later retrieval. the content_key is determined by
>         # the content_key field if the metadata is a `COURSE` or `PROGRAM` or by the parent_content_key
>         # field if the metadata is a `COURSE_RUN`.
>         for metadata in content_metadata:
> ```

@iloveagent57 that task failure was due to me not properly adding the through table to the model. I've updated and retested the task and it seems to be behaving appropriately

enterprise_catalog/apps/catalog/models.py
```diff
@@ -449,6 +471,83 @@ def __str__(self):
...
+        on_delete=models.deletion.DO_NOTHING,
+    )
+    content_metadata = models.ForeignKey(
+        ContentMetadata,
+        on_delete=models.deletion.DO_NOTHING,
+    )
+
+    class Meta:
+        app_label = 'catalog'
+        unique_together = ('content_metadata_id', 'catalog_query_id', 'deleted_at')
```
johnnagro, line 548: This was the key change - its an older syntax (newer way is with a constraints syntax) - it results in a unique key on the create table versus a separate constraint migration step and thus allows a clean roll forward/backwards.

iloveagent57 (approved):

PR #370, opened 2022-01-06
feat: Explore Catalog CSV Data Cleanup

johnnagro:
## Description

- use algolia search rather than browse_index for distinct records (fixes duplicate rows)
- reformat/refactor csv row building
- update tests

## References

- [ENT-5291](https://openedx.atlassian.net/browse/ENT-5291)

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -425,7 +425,7 @@ class EnterpriseCatalogCsvDataViewTests(APITestMixin):
     """
     Tests for the CatalogCsvDataView view.
     """
-    mock_algolia_hits = {'hits':[{
+    mock_algolia_hits = {'hits': [{
         'aggregation_key': 'course:MITx+18.01.2x',
         'language': 'English',
         'level_type': 'Intermediate',
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -425,7 +425,7 @@ class EnterpriseCatalogCsvDataViewTests(APITestMixin):
     """
     Tests for the CatalogCsvDataView view.
     """
-    mock_algolia_hits = [{
+    mock_algolia_hits = {'hits':[{
         'aggregation_key': 'course:MITx+18.01.2x',
         'language': 'English',
         'level_type': 'Intermediate',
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -514,19 +514,19 @@ def test_csv_row_construction_handles_missing_values(self, mock_algolia_client):
...
         url = self._get_contains_content_base_url()
         facets = 'language=English'
         response = self.client.get(f'{url}?{facets}')
         assert response.status_code == 200
         excpected_csv_data = 'Title,Partner Name,Start,End,Verified Upgrade Deadline,Program Type,Program Name,' \
                              'Pacing,Level,Price,Language,URL,Short Description,Subjects,Key,Short Key,Skills\r\n' \
-                             'Calculus 1B: Integration,Massachusetts Institute of Technology,2015-09-08T00:00:00Z,' \
-                             'No end date,No upgrade deadline,[\'Professional Certificate\'],[\'Totally Awesome ' \
-                             'Program\'],instructor_paced,Intermediate,No price,English,No url,description,' \
-                             '[\'Math\'],MITx/18.01.2x/3T2015,course:MITx+18.01.2x,"[\'Probability And Statistics\', ' \
```
kiram15, line 528: Nit: Could we add in a test case where the title or one of the skills has an apostrophe to show that they are properly handled?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -264,64 +266,57 @@ def construct_csv_row(self, hit):
...
+
+            end_date = None
+            if hit['advertised_course_run'].get('end'):
+                end_date = parser.parse(hit['advertised_course_run']['end']).strftime("%Y-%m-%d")
+            csv_row.append(end_date)
+
+            upgrade_deadline = None
+            if hit['advertised_course_run'].get('upgrade_deadline'):
+                raw_deadline = hit['advertised_course_run']['upgrade_deadline']
+                upgrade_deadline = datetime.datetime.fromtimestamp(raw_deadline).strftime("%Y-%m-%d")
```
kiram15, line 290: Thank you for making this the same format instead of the mm/dd/yy format it was earlier!

kiram15 (approved):
One nit that's not necessary to merge but otherwise LGTM!

enterprise_catalog/apps/api/v1/tests/test_views.py
johnnagro: Yeah - one of the fields is actually escaped right now, but you've given me some ideas for other tests i can add to be more thorough. I'll add more cases when we add the Excel Workbook support 👍

PR #371, opened 2022-01-07
feat: Add relevant fields to the Explore Catalog CSV Download

johnnagro:
## Description

- combine discovery metadata with algolia hits
- add new metadata to CSV download
- update tests

## References


enterprise_catalog/apps/api/v1/views.py
```diff
@@ -317,13 +323,40 @@ def construct_csv_row(self, hit):
...
         skills = [skill['name'] for skill in hit.get('skills', [])]
         csv_row.append(', '.join(skills))
+
+        discovery_course = hit.get('discovery_course', {})
+        discovery_course_runs = discovery_course.get('course_runs', [])
+
+        try:
+           discovery_first_course_run = discovery_course_runs[0]
+        except IndexError:
+           discovery_first_course_run = {}
```
binodpant: so if there's no course_runs we will still add a csv_row with null data?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -343,7 +376,19 @@ def retrieve_indexed_data(self, facets, algoliaQuery):
         algolia_hits = []
         page = algolia_client.algolia_index.search(algoliaQuery, search_options)
         while len(page['hits']) > 0:
-            algolia_hits.extend(page['hits'])
+            course_keys_chunk = [hit['key'] for hit in page.get('hits', [])]
+            query_params = {'keys': ','.join(course_keys_chunk)}
+            courses = discovery_client.get_courses(query_params=query_params)
+            course_by_key = {}
+            for course in courses:
```
binodpant, line 386: not sure I quite get this part and how this helps with the additional fields above...

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -343,7 +376,19 @@ def retrieve_indexed_data(self, facets, algoliaQuery):
...
+                course_by_key[course['key']] = course
+
+            for hit in page['hits']:
+                if course_by_key.get(hit['key']):
+                    hit['discovery_course'] = course_by_key.get(hit['key'])
+                algolia_hits.append(hit)
+
+            # algolia_hits.extend(page['hits'])
```
binodpant: remove this commented line?

enterprise_catalog/apps/api/v1/views.py
johnnagro: Correct, that is my intention.

enterprise_catalog/apps/api/v1/views.py
johnnagro: the discovery api returns a list of courses, which i want to combine with the algolia course results. to improve performance, i am building a lookup dictionary of the discovery course metadata results, keyed by the course key. then when i iterate over the algolia results (hits) i can efficiently look in the dictionary to see if there is a matching (by course key) course discovery metadata result. i add the extra discovery data to the hit dictionary and append it to the results list.

does that make sense?

binodpant: ah I see it’s an intermediate data structure, was trying to make sense of it. this clarifies it thanks!

binodpant (approved):
lgtm!

PR #374, opened 2022-01-09
Soft deletion on query to metadata through table and memory improvements to the reindexing algolia job  

alex-sheehan-edx:
relevant jira ticket: https://openedx.atlassian.net/browse/ENT-5283

Context: This PR redeploys the soft deletion update to the content metadata/query through table and restructures the algolia reindexing job to utilize said table in order to more efficiently iterate over queries and content through the Django ORM (ie uses django's `iterator()` method on top of querying the through table instead of each individual side

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review


johnnagro (approved):

enterprise_catalog/apps/api/tasks.py
```diff
@@ -488,109 +554,90 @@ def index_content_keys_in_algolia(content_keys, algolia_client):  # pylint: disa
...
-        for metadata in filtered_content_metadata:
-            content_key = metadata.content_key
+                for program in course_to_program_mapping[content_key]:
+                    catalog_queries_by_key[program.content_key].update({(str(catalog_query.uuid), catalog_query.title)})
+                    catalog_uuids_by_key[program.content_key].update(query_to_catalog_mapping[catalog_query.id])
+                    customer_uuids_by_key[program.content_key].update(query_to_enterprise_mapping[catalog_query.id])
+
+        course_membership = ContentMetadataToQueries.objects.select_related(
+            'catalog_query', 'content_metadata'
+        ).filter(query).all().iterator()
```
iloveagent57: Do you really want to iterate over memberships again, here?  I think you just want the `ContentMetadata` filtered to keys that are in this batch, right?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -470,7 +472,71 @@ def index_enterprise_catalog_in_algolia_task(self, force=False):  # pylint: disa
...
+        for course in prog.associated_content_metadata.all():
+            program_membership_by_course_key[course.content_key].add(prog)
+    return program_membership_by_course_key
+
+
+def get_catalogs_by_queries():
+    """ Prefetch catalog uuids by catalogquery.id mapping """
+    catalog_uuid_by_query_id = defaultdict(set)
+    enterprise_uuid_by_query_id = defaultdict(set)
+    for catalog in EnterpriseCatalog.objects.iterator():
```
iloveagent57: Maybe add in a `.all()` here just to make it explicit?

enterprise_catalog/apps/api/tasks.py
alex-sheehan-edx: oh yeah, that would solve the programs vs courses, wouldn't it

enterprise_catalog/apps/api/tasks.py
```diff
@@ -488,109 +554,81 @@ def index_content_keys_in_algolia(content_keys, algolia_client):  # pylint: disa
     catalog_uuids_by_key = defaultdict(set)
     customer_uuids_by_key = defaultdict(set)
     catalog_queries_by_key = defaultdict(set)
```
iloveagent57: I know this isn't something you changed, but do these actually need to be outside the main for-loop?

enterprise_catalog/apps/api_client/algolia.py
```diff
@@ -91,6 +91,22 @@ def index_exists(self):
...
+            self.algolia_index.partial_update_objects(algolia_objects, {
+                'createIfNotExists': True
+            })
+        except AlgoliaException as exc:
+            logger.exception(
+                f'Could not index {len(algolia_objects)} objects in the {self.ALGOLIA_INDEX_NAME} Algolia index due to '
+                f'an exception.',
+            )
+            raise exc
+
```
iloveagent57: we're not using this anymore, right?

alex-sheehan-edx: oh yes, artifact of old methods... will remove

enterprise_catalog/apps/api/tasks.py
alex-sheehan-edx: hmm, initially trying things with the dicts inside the for loop it seems like all tests pass

alex-sheehan-edx: thinking about the impact, we would be resetting the three dictionaries over each iteration of batched content keys instead of building up three dictionaries over all content keys. Regardless of where the dictionaries are initialized, we add the contents of the dictionaries to the algolia objects to index object at the end of the loop. I feel like it's ok to keep it inside the loop and therefore keep the dictionaries smaller?

iloveagent57 (approved):

PR #375, opened 2022-01-10
feat: Add relevant fields to the Explore Catalog CSV Download

johnnagro:
## Description

- follow-on to #373
- be more defensive with accessing algolia results (missing keys)
- extend tests to cover edge-cases seen in PROD

## References

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -458,7 +458,21 @@ class EnterpriseCatalogCsvDataViewTests(APITestMixin):
             'upgrade_deadline': 32503680000.0,
         },
         'objectID': 'course-3543aa4e-3c64-4d9a-a343-5d5eda1dacf8-catalog-query-uuids-0'
-    }]}
+    },
+        {
+        'aggregation_key': 'course:MITx+19',
+        'key': 'MITx+19',
+        'language': 'English',
+        'level_type': 'Intermediate',
+        'objectID': 'course-3543aa4e-3c64-4d9a-a343-5d5eda1dacf9-catalog-query-uuids-0'
+    },
+        {
+        'aggregation_key': 'course:MITx+20',
+        'language': 'English',
+        'level_type': 'Intermediate',
+        'objectID': 'course-3543aa4e-3c64-4d9a-a343-5d5eda1dacf7-catalog-query-uuids-0'
+    }
+    ]}
 
     mock_discovery_courses = [
         {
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -458,7 +458,21 @@ class EnterpriseCatalogCsvDataViewTests(APITestMixin):
             'upgrade_deadline': 32503680000.0,
         },
         'objectID': 'course-3543aa4e-3c64-4d9a-a343-5d5eda1dacf8-catalog-query-uuids-0'
-    }]}
+    },
+        {
+        'aggregation_key': 'course:MITx+19',
```
johnnagro, line 463: test cases with missing `content_type` and `key` fields, which is something unexpected we saw in production algolia data

kiram15 (approved):
LGTM!

binodpant:
looks like the commit message should have been a `fix: ...` instead? but yeah the change itself looks good to me. I wonder what those cases of missing key and/or content_type are... seems we are not good about enforcing schemas somewhere ( or maybe there are legit cases ?)

binodpant (approved):
looks good, just minor feedback/q

enterprise_catalog/apps/api/v1/tests/test_views.py
alex-sheehan-edx: missing `content_type`s in production seems highly alarming. What were the cases that this happened?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -380,25 +380,26 @@ def retrieve_indexed_data(self, facets, algoliaQuery):
             course_keys_chunk = []
             for hit in page.get('hits', []):
                 # ignore program data (for now)
-                if hit['content_type'] != 'course':
+                if hit.get('content_type') != 'course':
                     continue
-                course_keys_chunk.append(hit['key'])
+                if hit.get('key'):
```
alex-sheehan-edx, line 385: won't this mess with ordering of the data if the key doesn't exist?

alex-sheehan-edx (approved):
two questions but otherwise looks good

enterprise_catalog/apps/api/v1/views.py
alex-sheehan-edx: oh I see! Nevermind

johnnagro:
> missing content_types in production seems highly alarming. What were the cases that this happened?

agreed. since the button is "broken" given the current code and we're working on a wholesale replacement in parallel, I prioritized being defensive over chasing down exactly whats up in Algolia.

I will be looking a bit deeper for my parallel work.

PR #376, opened 2022-01-11
fix: Add relevant fields to the Explore Catalog CSV Download

johnnagro:
## Description

- follow-on to #375
- be more defensive with accessing algolia results (missing keys)
- ensure `key` and `content_type` are returned by algolia

## References

### DIFF
--- a/enterprise_catalog/apps/api/v1/views.py
+++ b/enterprise_catalog/apps/api/v1/views.py
@@ -210,6 +210,8 @@ class CatalogCsvDataView(GenericAPIView):
     ]
     algolia_attributes_to_retrieve = [
         'title',
+        'key',
+        'content_type',
         'partners',
         'advertised_course_run',
         'programs',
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api/v1/views.py
```diff
@@ -384,17 +386,18 @@ def retrieve_indexed_data(self, facets, algoliaQuery):
...
+            if len(course_keys_chunk) > 0:
+                query_params = {'keys': ','.join(course_keys_chunk)}
+                courses = discovery_client.get_courses(query_params=query_params)
+                for course in courses:
+                    course_by_key[course.get('key')] = course
 
             # combine discovery metadata with the algolia results
             # append the hit to the results
-            for hit in page['hits']:
+            for hit in page.get('hits', []):
```
binodpant, line 400: is there a test change that is needed to cover this? other than this, this is looking fine to me

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -210,6 +210,8 @@ class CatalogCsvDataView(GenericAPIView):
     ]
     algolia_attributes_to_retrieve = [
         'title',
+        'key',
```
johnnagro, line 213: these attributes were not consistently being returned, which i think is the reason why i was seeing such strange differences between local, stage, and prod

enterprise_catalog/apps/api/v1/views.py
alex-sheehan-edx: when do we expect no hits? When will page not have at least an empty hits attribute?

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -210,6 +210,8 @@ class CatalogCsvDataView(GenericAPIView):
...
+        'content_type',
```
alex-sheehan-edx, line 214: oof, I probably should have had content_type here from the beginning 😬

alex-sheehan-edx (approved):
looks good to me, just a couple context questions/comments

enterprise_catalog/apps/api/v1/views.py
johnnagro: no, just being extra defensive in what we get back from algolia

PR #382, opened 2022-01-18
feat: stream CSV export

johnnagro:
## Description

- new endpoint
- improved perception of performance (download starts fast)
- no cut-off before large exports can begin downloading
- DRY up export tools

TODO: transition frontned to use new endpoint

## References

- [ENT-5222](https://openedx.atlassian.net/browse/ENT-5222)


enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -0,0 +1,168 @@
...
+        csv_row.append(None)  # no upgrade deadline
+        pacing_type = None
+        key = None
+
+    csv_row.append(', '.join(hit.get('programs', [])))
+    csv_row.append(', '.join(hit.get('program_titles', [])))
+
+    csv_row.append(pacing_type)
+
+    csv_row.append(hit.get('level_type', 'No level_type'))
```
alex-sheehan-edx: if we're doing `None` and empty lists elsewhere for values not found, do we also want to do if for level type?

alex-sheehan-edx (approved):
one question but LGTM

PR #383, opened 2022-01-21
feat: new backend API for default course views

alex-sheehan-edx:
## Description

new api view to retrieve first four ranked courses under a catalog in algolia. 

## Ticket Link

Link to the associated ticket
[5241](https://openedx.atlassian.net/browse/ENT-5241)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -38,6 +38,103 @@ from enterprise_catalog.apps.catalog.utils import get_parent_content_key
 
 
 @ddt.ddt
+class EnterpriseCatalogDefaultCatalogResultsTests(APITestMixin):
+    """
+    Tests for the DefaultCatalogResultsView class
+    """
+    mock_algolia_hits = {'hits': [{
+        'aggregation_key': 'course:MITx+18.01.2x',
+        'key': 'MITx+18.01.2x',
+        'language': 'English',
+        'level_type': 'Intermediate',
+        'content_type': 'course',
+        'partners': [
+            {'name': 'Massachusetts Institute of Technology',
+             'logo_image_url': 'https://edx.org/image.png'}
+        ],
+        'programs': ['Professional Certificate'],
+        'program_titles': ['Totally Awesome Program'],
+        'short_description': 'description',
+        'subjects': ['Math'],
+        'skills': [{
+            'name': 'Probability And Statistics',
+            'description': 'description'
+        }, {
+            'name': 'Engineering Design Process',
+            'description': 'description'
+        }],
+        'title': 'Calculus 1B: Integration',
+        'marketing_url': 'edx.org/foo-bar',
+        'first_enrollable_paid_seat_price': 100,
+        'advertised_course_run': {
+            'key': 'MITx/18.01.2x/3T2015',
+            'pacing_type': 'instructor_paced',
+            'start': '2015-09-08T00:00:00Z',
+            'end': '2015-09-08T00:00:01Z',
+            'upgrade_deadline': 32503680000.0,
+        },
+        'objectID': 'course-3543aa4e-3c64-4d9a-a343-5d5eda1dacf8-catalog-query-uuids-0'
+    },
+        {
+        'aggregation_key': 'course:MITx+19',
+        'key': 'MITx+19',
+        'language': 'English',
+        'level_type': 'Intermediate',
+        'objectID': 'course-3543aa4e-3c64-4d9a-a343-5d5eda1dacf9-catalog-query-uuids-0'
+    },
+        {
+        'aggregation_key': 'course:MITx+20',
+        'language': 'English',
+        'level_type': 'Intermediate',
+        'objectID': 'course-3543aa4e-3c64-4d9a-a343-5d5eda1dacf7-catalog-query-uuids-0'
+    }
+    ]}
+
+    def setUp(self):
+        super().setUp()
+        self.set_up_staff_user()
+
+    def _get_contains_content_base_url(self):
+        """
+        Helper to construct the base url for the contains_content_items endpoint
+        """
+        return reverse('api:v1:default-course-set')
+
+    def test_facet_validation(self):
+        """
+        Tests that the view validates Algolia facets provided by query params
+        """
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -36,6 +36,91 @@
...
+    """
+    Tests for the DefaultCatalogResultsView class
+    """
+    mock_algolia_hits = {'hits': [{
+        'aggregation_key': 'course:MITx+18.01.2x',
+        'key': 'MITx+18.01.2x',
+        'language': 'English',
+        'level_type': 'Intermediate',
+        'content_type': 'course',
+        'partners': [
```
kiram15, line 51: Programs are a little different, authoring_orginization is more reliable way to show who the program is written by, because some programs in prod don't even have an associated partner.

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -174,6 +174,89 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids, *
...
+        'marketing_url',
+        'upcoming_course_runs',
+        'type',
+        'recent_enrollment_count',
+        'original_image_url',
+        'full_description',
+        'enterprise_catalog_query_uuids',
+        'enterprise_catalog_query_titles',
+        'card_image_url',
+        'availability',
```
kiram15: Maybe check with Binod that you have all the variables that will be eventually needed for program modal? (course_keys, price, program_type, etc)

kiram15 (changes requested):
My main thing is just to make it flexible enough that it will also be able to accommodate the future api calls for default programs as well

enterprise_catalog/apps/api/v1/views.py
alex-sheehan-edx: is price an option for programs currently? 🤔

binodpant: I indicated to. Alex that I am going to call discovery to get course details for program anyway so the price is already also in that response so I should have what I need AFAICT already (course keys, the what you will learn stuff (subtitle), and total program price, are in discovery response)

enterprise_catalog/apps/api/v1/views.py
```diff
@@ -174,6 +174,101 @@ def contains_content_items(self, request, uuid, course_run_ids, program_uuids, *
...
+            )
+
+        invalid_facets = self.validate_query_facets(facets)
+
+        content_type = facets.get("content_type", ['course'])[0]
+        if invalid_facets:
+            return Response(f'Error: invalid facet(s): {invalid_facets} provided.', status=HTTP_400_BAD_REQUEST)
+
+        catalog_filter = [
+            f'enterprise_catalog_query_titles:{facets.get("enterprise_catalog_query_titles")[0]}',
```
binodpant: do we only ever want to use the first of these query titles?

alex-sheehan-edx: yeah, the hope is that only one query title is provided

johnnagro (commented):
I think I'd prefer large refactors (moving view classes into dedicated files) to be separated into distinct PRs from from functionality changes. It makes it hard to pick out whats new functionality versus just new organization.

I also want to know if we've considered just re-using the front-end algolia client to make a call without a query/facets for the top 4 results, versus introducing a dedicated api? its a lot of extra surface area to maintain, as well as keeping the JSON response structures in some degree of parity, right?

enterprise_catalog/apps/api/v1/views/default_catalog_results.py
```diff
@@ -0,0 +1,103 @@
...
+        'full_description',
+        'enterprise_catalog_query_uuids',
+        'enterprise_catalog_query_titles',
+        'card_image_url',
+        'availability',
+        'program_type',
+        'course_keys',
+    ]
+
+    def validate_query_facets(self, facets):
```
johnnagro: i pulled these into `export_utils` to DRY up their use in the CSV exporter classes. You could re-use.

enterprise_catalog/apps/api/v1/views/default_catalog_results.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        Verify that provided query facet params are valid Algolia facets.
+        """
+        invalid_facets = []
+        for facet in facets.keys():
+            if facet not in self.valid_facets:
+                invalid_facets.append(facet)
+        return invalid_facets
+
+    def querydict_to_dict(self, query_dict):
```
johnnagro: i pulled these into `export_utils` to DRY up their use in the CSV exporter classes. You could re-use.

enterprise_catalog/apps/api/v1/views/default_catalog_results.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        GET entry point for the `CatalogCsvDataView`
+        """
+        facets = self.querydict_to_dict(request.query_params)
+
+        invalid_facets = self.validate_query_facets(facets)
+
+        content_type = facets.get("content_type", ['course'])[0]
+        if invalid_facets:
+            return Response(f'Error: invalid facet(s): {invalid_facets} provided.', status=HTTP_400_BAD_REQUEST)
```
johnnagro: should this be JSON formatted? ie machine readable so the front-end has a chance to do something with the message?

enterprise_catalog/apps/api/v1/views.py
johnnagro: price is available in algolia for courses, we should put it into programs documents in algolia too imo. (versus calling into discovery separately)

binodpant: yup so clarified in chat. The pull I am doing from discovery is going to be for only one program at a time, when the modal opens, it won't be for listing views like here. I agree though that for list/search use cases, we should leverage algolia when we can, we can discuss post this PR what options we can go with

binodpant: also I need to pull more than just price, there is course subtitles, course image etc. Adding all that to algolia may or may not be desirable, something we can also chat more about

kiram15 (approved):

PR #385, opened 2022-01-24
Cleanup of the enterprise catalog views

alex-sheehan-edx:
## Description

Breaking down the catalog api views into their own files for clarity.

## Post-review


kiram15 (approved):
Wow this looks so much better than a 717 line file .... woof

enterprise_catalog/apps/api/v1/views/enterprise_cusomter.py
```diff
@@ -0,0 +1,104 @@
+import logging
```
binodpant: typo in filename 
`enterprise_catalog/apps/api/v1/views/enterprise_cusomter.py`

binodpant (approved):
as long as the tests all pass and there is no actual url change this looks fine to me,

I commented on one typo you want to fix

please ensure the routes are all functional. I hope we have sufficient test cases to cover all the routes

enterprise_catalog/apps/api/v1/views/enterprise_cusomter.py
alex-sheehan-edx: yikes!

PR #392, opened 2022-02-01
feat: Improve algolia to enable better Formatting of CSV Download from Enterprise Catalog

johnnagro:
## Description

- add field to algolia to remove dep on discovery service
- removing that dependency will let us improve speed to the point that we can move back to excel

## References

- [ENT-5381](https://openedx.atlassian.net/browse/ENT-5381)

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -3,6 +3,7 @@ import datetime
 import logging
 import time
 
+from django.utils.html import strip_tags
 from django.utils.translation import ugettext as _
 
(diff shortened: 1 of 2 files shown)


kiram15 (commented):
It mostly looks good to me, but was this supposed to add any kind of cell formatting? All cell sizes seem small and consistent on my end

johnnagro:
> It mostly looks good to me, but was this supposed to add any kind of cell formatting? All cell sizes seem small and consistent on my end

sorry i wasnt more clear - this just sets up data in algolia, the actual excel stuff (with formatting) comes after this goes in

alex-sheehan-edx (approved):
LGTM

PR #396, opened 2022-02-05
feat: ENT-5364 Added pathway indexing for Algolia

irfanuddinahmad:
## Description

This PR adds Pathways type content metadata to the Algolia search index
image

## Ticket Link

[ENT-5364](https://openedx.atlassian.net/browse/ENT-5364)

PR #401, opened 2022-02-08
Added a new task to load missing programs metadata into the database.

saleem-latif:
## Description
Update command update_content_metadata from enterprise-catalog repo to save pathways content metadata as it is saving for programs and courses

## Ticket Link

Link to the associated ticket
[ENT-5362](https://openedx.atlassian.net/browse/ENT-5362)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -23,6 +23,8 @@ from enterprise_catalog.apps.catalog.algolia_utils import (
     create_algolia_objects,
     get_algolia_object_id,
     get_initialized_algolia_client,
+    get_pathway_course_keys,
+    get_pathway_program_uuids,
     partition_course_keys_for_indexing,
     partition_program_keys_for_indexing,
 )
@@ -30,7 +32,7 @@ from enterprise_catalog.apps.catalog.constants import (
     COURSE,
     DISCOVERY_COURSE_KEY_BATCH_SIZE,
     DISCOVERY_PROGRAM_KEY_BATCH_SIZE,
-    PATHWAY,
+    LEARNER_PATHWAY,
     PROGRAM,
     REINDEX_TASK_BATCH_SIZE,
     TASK_BATCH_SIZE,
@@ -463,8 +465,12 @@ def index_enterprise_catalog_in_algolia_task(self, force=False):  # pylint: disa
         indexable_program_keys, nonindexable_program_keys = partition_program_keys_for_indexing(
             programs_content_metadata,
         )
-        indexable_pathways_keys = ContentMetadata.objects.filter(content_type=PATHWAY).distinct().values_list(
-            'content_key', flat=True)
+        indexable_pathways_keys = ContentMetadata.objects.filter(
+            content_type=LEARNER_PATHWAY
+        ).distinct().values_list(
+            'content_key',
+            flat=True
+        )
         indexable_content_keys = indexable_course_keys + indexable_program_keys + list(indexable_pathways_keys)
         nonindexable_content_keys = nonindexable_course_keys + nonindexable_program_keys
         _reindex_algolia(
@@ -490,7 +496,9 @@ def get_pathways_by_associated_content():
     """ Prefetch course id and program id to pathway id mapping. """
     pathway_membership_by_course_key = defaultdict(set)
     pathway_membership_by_program_key = defaultdict(set)
-    pathways = ContentMetadata.objects.filter(content_type=PATHWAY).prefetch_related('associated_content_metadata')
+    pathways = ContentMetadata.objects.filter(content_type=LEARNER_PATHWAY).prefetch_related(
+        'associated_content_metadata'
+    )
     for pathway in pathways:
         for associated_content in pathway.associated_content_metadata.all():
             if associated_content.content_type == COURSE:
@@ -627,7 +635,7 @@ def index_content_keys_in_algolia(content_keys, algolia_client):
             | Q(content_metadata__parent_content_key__in=content_keys_batch)
             | Q(
                 content_metadata__associated_content_metadata__content_key__in=content_keys_batch,
-                content_metadata__content_type__in=[PROGRAM, PATHWAY]
+                content_metadata__content_type__in=[PROGRAM, LEARNER_PATHWAY]
             )
         )
         all_memberships = ContentMetadataToQueries.objects.select_related(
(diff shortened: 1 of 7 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -779,3 +780,75 @@ def fetch_missing_course_metadata_task(self):  # pylint: disable=unused-argument
...
+    """
+
+    learner_pathway_metadata_list = ContentMetadata.objects.filter(content_type=LEARNER_PATHWAY).values_list(
+        'json_metadata', flat=True,
+    )
+    program_uuids = set()
+    course_keys = set()
+    for learner_pathway_metadata in learner_pathway_metadata_list:
+        if learner_pathway_metadata is not None:
+            program_uuids.update([item.get('key') for item in learner_pathway_metadata.get('programs', set())])
```
HammadAhmadWaqas: Please use this utility defined here https://github.com/openedx/enterprise-catalog/pull/396/files#diff-f95d821947945f8e3094ac125b397892d0f51cf2f2169c01dec21f9d0c080633R359
`get_pathway_course_keys` `get_pathway_program_uuids`

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -9,6 +9,10 @@
 COURSE = 'course'
 COURSE_RUN = 'courserun'
 PROGRAM = 'program'
+
+# TODO: Update this after the correct value has been merged.
+LEARNER_PATHWAY = 'learner_pathway'
+
```
HammadAhmadWaqas, line 13: value should be `learnerpathway `
please add this in CONTENT_TYPE_CHOICES

HammadAhmadWaqas (changes requested):
please add the `include_learner_pathways` flag in search all query param as requested is the ticket.

enterprise_catalog/apps/api/tasks.py
saleem-latif: Okay 👍 , I will add this once the PR containing those methods is merged.

HammadAhmadWaqas: actually that PR is dependent on your PR. can you copy that utility in your code and when we will rebase, we will resolve conflicts.

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -10,13 +10,13 @@
...
 
-# TODO: Update this after the correct value has been merged.
-LEARNER_PATHWAY = 'learner_pathway'
+LEARNER_PATHWAY = 'learnerpathway'
 
 CONTENT_TYPE_CHOICES = [
     (COURSE, 'Course'),
     (COURSE_RUN, 'Course Run'),
     (PROGRAM, 'Program'),
+    (LEARNER_PATHWAY, 'Learner Pathway'),
```
HammadAhmadWaqas, line 18: Do we need migration for this change?

saleem-latif: Good catch, yes I will add a migration for this as well.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -832,3 +845,74 @@ def fetch_missing_course_metadata_task(self):  # pylint: disable=unused-argument
...
+    present_program_uuids = ContentMetadata.objects.filter(
+        content_type=PROGRAM, content_key__in=program_uuids
+    ).values_list(
+        'content_key', flat=True
+    )
+
+    missing_program_uuids = program_uuids.difference(present_program_uuids)
+    content_filter = {
+        'status': 'published',
+        'key': list(missing_program_uuids),
```
HammadAhmadWaqas: What if there is no missing key? I think we should not simply proceed if there is no missing key as it will create a dummy query daily.
Same should be true for `fetch_missing_course_metadata_task`

enterprise_catalog/apps/api/tasks.py
```diff
@@ -832,3 +845,74 @@ def fetch_missing_course_metadata_task(self):  # pylint: disable=unused-argument
...
+        'content_type': 'program',
+    }
+    catalog_query, _ = CatalogQuery.objects.get_or_create(
+        content_filter_hash=get_content_filter_hash(content_filter),
+        defaults={'content_filter': content_filter, 'title': None},
```
HammadAhmadWaqas: Can use create a Generic `title` here inserted of None, As it can help us trace queries that are created automatically.
example: `Auto-generated query to fetch_missing_pathway_metadata`

enterprise_catalog/apps/api/tasks.py
```diff
@@ -832,3 +845,74 @@ def fetch_missing_course_metadata_task(self):  # pylint: disable=unused-argument
...
+    missing_course_keys = course_keys.difference(present_course_keys)
+    content_filter = {
+        'status': 'published',
+        'key': list(missing_course_keys),
+        'content_type': 'course',
+    }
+
+    catalog_query, _ = CatalogQuery.objects.get_or_create(
+        content_filter_hash=get_content_filter_hash(content_filter),
+        defaults={'content_filter': content_filter, 'title': None},
```
HammadAhmadWaqas: Same as above.
Also should we consider make this change in old similar task?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1035,3 +1035,37 @@ def create_algolia_objects(products, algolia_fields):
...
+    Returns:
+       list: a list of program uuids associated with the pathway.
+    """
+    program_uuids = set()
+    steps = pathway.get('steps') or []
+    for step in steps:
+        programs = step.get('programs') or []
+        for program in programs:
+            program_uuids.add(program['uuid'])
+    return list(program_uuids)
```
HammadAhmadWaqas: Why we are creating these utilities again? Should not it be already in the feature branch?

HammadAhmadWaqas (changes requested):

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -9,12 +9,14 @@
 COURSE = 'course'
 COURSE_RUN = 'courserun'
 PROGRAM = 'program'
-PATHWAY = 'learnerpathway'
```
HammadAhmadWaqas, line 12: There is still one reference here enterprise_catalog/apps/catalog/algolia_utils.py with the old name

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -9,12 +9,14 @@
...
+
```
HammadAhmadWaqas: nit: can we remove this space?

HammadAhmadWaqas (changes requested):
Also, can we fix similar issues in fetch_missing_course_metadata_task as well?
What would you suggest?

HammadAhmadWaqas (approved):

PR #403, opened 2022-02-09
fix: Pre-requisites not appearing for courses in Explore Catalog

johnnagro:
## Description

- added the correct field for indexing
- added a test to check retrieved vs indexed fields

## References

- [ENT-5415](https://openedx.atlassian.net/browse/ENT-5415)

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_export_utils.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_export_utils.py
@@ -1,0 +1,17 @@
+from django.test import TestCase
+
+from enterprise_catalog.apps.api.v1 import export_utils
+from enterprise_catalog.apps.catalog import algolia_utils
+
+
+class ExportUtilsTests(TestCase):
+    """
+    Tests for the Enterprise Catalog API export utils
+    """
+
+    def test_retrieve_available_fields(self):
+        """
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -41,6 +41,7 @@
     'partner',
     'partners',
     'prerequisites',
+    'prerequisites_raw',
```
johnnagro, line 44: we were missing this field in the indexing

enterprise_catalog/apps/api/v1/tests/test_export_utils.py
```diff
@@ -0,0 +1,17 @@
+from django.test import TestCase
+
+from enterprise_catalog.apps.api.v1 import export_utils
+from enterprise_catalog.apps.catalog import algolia_utils
+
+
+class ExportUtilsTests(TestCase):
```
johnnagro, line 7: this will fail if we ever try to add a field for export which we're not indexing

alex-sheehan-edx: great test

alex-sheehan-edx (approved):
lgtm

PR #404, opened 2022-02-09
fix: Download is empty with prefix search terms

johnnagro:
## Description

- param manipulation was creating stricter search:

`['query param value']` -> `query param value`

- added error state when results are empty

## References

- tested locally

## References

- [ENT-5413](https://openedx.atlassian.net/browse/ENT-5413)

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -735,7 +735,7 @@ class EnterpriseCatalogWorkbookViewTests(APITestMixin):
     @mock.patch('enterprise_catalog.apps.api.v1.views.catalog_workbook.get_initialized_algolia_client')
     def test_success(self, mock_algolia_client):
         """
-        Tests when algolia returns no hits.
+        Tests basic, successful output.
         """
         mock_algolia_client.return_value.algolia_index.search.side_effect = [self.mock_algolia_hits, {'hits': []}]
         url = self._get_contains_content_base_url()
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/views/catalog_workbook.py
```diff
@@ -69,6 +75,10 @@ def get(self, request, **kwargs):
 
         # Algolia search will only retrieve all results if you query by empty string.
         page = algolia_client.algolia_index.search(algoliaQuery, search_options)
+
+        if len(page['hits']) == 0:
```
binodpant, line 81: would be nice to throw in a test for this? otherwise lgtm

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -654,6 +654,96 @@ def test_csv_row_construction_handles_missing_values(self, mock_algolia_client):
...
+    @mock.patch('enterprise_catalog.apps.api.v1.views.catalog_workbook.get_initialized_algolia_client')
+    def test_success(self, mock_algolia_client):
+        """
+        Tests basic, successful output.
+        """
+        mock_algolia_client.return_value.algolia_index.search.side_effect = [self.mock_algolia_hits, {'hits': []}]
+        url = self._get_contains_content_base_url()
+        facets = 'language=English'
+        response = self.client.get(f'{url}?{facets}')
+        assert response.status_code == 200
```
binodpant, line 744: maybe better to also test response here to some extent but otherwise lgtm!

binodpant (approved):

PR #406, opened 2022-02-10
ENT-5221 algolia course_details field in program object

binodpant:
## Description

course_details sourced from the disocery field of programs > courses, will also be indexed in algolia

### TESting notes

Tested against ny local aloglia index and verified it ended up with these fields 

Screen Shot 2022-02-10 at 5 10 29 PM

## Ticket Link

[ENT-5221](https://openedx.atlassian.net/browse/ENT-5221)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api_client/discovery.py
+++ b/enterprise_catalog/apps/api_client/discovery.py
@@ -158,6 +158,7 @@ class DiscoveryApiClient(BaseOAuthClient):
         request_params = {
             'ordering': 'key',
             'limit': DISCOVERY_OFFSET_SIZE,
+            'extended': 'True',
         }
         request_params.update(query_params or {})
 
--- a/enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
@@ -22,6 +22,7 @@ from enterprise_catalog.apps.catalog.algolia_utils import (
     get_initialized_algolia_client,
     get_program_availability,
     get_program_banner_image_url,
+    get_program_course_details,
     get_program_learning_items,
     get_program_level_type,
     get_program_partners,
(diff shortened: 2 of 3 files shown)


enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -637,6 +638,21 @@ def test_get_program_prices(self, program_metadata, expected_type):
         program_prices = get_program_prices(program_metadata)
         self.assertEqual(expected_type, program_prices)
 
+    @ddt.data(
+        (
+            {'courses': [{'key': 'akey', 'title': 'a_title', 'image': 'an_image', 'short_description': 'desc'}]},
```
kiram15, line 643: Q: Would there ever be a case where some but not all of these attributes are populated? Do we have a fallback?

kiram15: To be honest, I've been thinking about having a default for course/program image so I that change can be in another ticket

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -158,6 +158,7 @@ def get_programs(self, query_params=None):
         request_params = {
             'ordering': 'key',
             'limit': DISCOVERY_OFFSET_SIZE,
+            'extended': 'True',
```
binodpant, line 161: this is needed to pull additional fields per @alex-sheehan-edx 's recent work

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
binodpant: good point, I think it will work meaning we won't pick them up but better have a test here

binodpant: do we typically exclude the fields altogehter in those cases vs insert a None? 

```
(
            {'courses': [{'key': 'akey'}]},
            [{'key': 'akey', 'title': None, 'image': None, 'short_description': None}],
        ),
```

binodpant: ok we can populate default values separately then, if I can put a None here for now ...

alex-sheehan-edx: `None` should suffice for now

alex-sheehan-edx (approved):
lgtm

PR #416, opened 2022-02-28
fix: do not index archived courses in algolia

johnnagro:
## Description

- take into account availability when deciding to index
- expand test coverage

## References

- [ENT-5432](https://openedx.atlassian.net/browse/ENT-5432)


binodpant (approved):
nice and clean!

PR #417, opened 2022-03-01
fix: re-enable archived in index

johnnagro:
## Description

- had intended on stamping out Archived courses from the index
- learner portal assumes/requires archived courses
- restore this functionality until we can discuss the issue with those teams

## References

- [ENT-5432](https://openedx.atlassian.net/browse/ENT-5432)

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -142,9 +142,6 @@ def _should_index_course(course_metadata):
     if not is_course_run_active(advertised_course_run):
         return False
 
-    if is_course_archived(course_json_metadata):
-        return False
-
     owners = course_json_metadata.get('owners') or []
     return not advertised_course_run.get('hidden') and len(owners) > 0
 
@@ -312,7 +309,7 @@ def is_course_archived(course):
         boolean: "Archived" availability or not
     """
     availability_list = get_course_availability(course)
-    return len(availability_list) == 0 or availability_list[0] == 'Archived'
+    return len(availability_list) == 0 or 'Archived' in availability_list
 
 
 def get_course_partners(course):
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -312,7 +309,7 @@ def is_course_archived(course):
         boolean: "Archived" availability or not
     """
     availability_list = get_course_availability(course)
-    return len(availability_list) == 0 or availability_list[0] == 'Archived'
+    return len(availability_list) == 0 or 'Archived' in availability_list
```
johnnagro, line 312: this is more correct since as it turns out, courses can have more than one availability and "Archived" can appear in any position in that list

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -59,7 +59,7 @@ class AlgoliaUtilsTests(TestCase):
         {'expected_result': False, 'is_enrollable': False},
         {'expected_result': False, 'is_marketable': False},
         {'expected_result': True, },
-        {'expected_result': False, 'course_run_availability': None},
+        {'expected_result': True, 'course_run_availability': None},
```
johnnagro, line 62: a course run with an availability of `None` becomes `"Archived"` in this list of availabilities, and should be indexed (for now)

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -142,9 +142,6 @@ def _should_index_course(course_metadata):
     if not is_course_run_active(advertised_course_run):
         return False
 
-    if is_course_archived(course_json_metadata):
```
johnnagro, line 145: removing the skip on indexing archived

binodpant (approved):

PR #419, opened 2022-03-07
feat: Only index needed image field

binodpant:
In prior work, I had left this indexing the entire image by mistake, we really don't need it, we need just image.src field in algolia. The frontend epxlore catalog is already written to handle both cases so it's a backward compatible change

We set image: null in case image is not to be found which works fine from the frontend perspecitve too and is accurate wrt representing the absence of an image

Change: don't index entire image field, it's not needed for algolia usage by MFEs right now

ENT-5439

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -366,7 +366,7 @@ def get_program_course_details(program):
         mapped_course = {
             'key': course.get('key'),
             'title': course.get('title'),
-            'image': course.get('image'),
+            'image': course.get('image').get('src') if course.get('image') else None,
             'short_description': course.get('short_description'),
         }
         course_list.append(mapped_course)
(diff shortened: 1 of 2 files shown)


johnnagro (approved):

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -366,7 +366,7 @@ def get_program_course_details(program):
         mapped_course = {
             'key': course.get('key'),
             'title': course.get('title'),
-            'image': course.get('image'),
+            'image': course.get('image').get('src') if course.get('image') else None,
```
johnnagro: rather than an if, i sometimes like to use the default values. i _believe_ this accomplishes the same thing:
`course.get('image', {}).get('src')`

PR #422, opened 2022-03-16
bug: update content metadata task no longer reconstruct existing content metadata/query relations

alex-sheehan-edx:
## Description

The update content metadata task now, instead of clearing the entire content metadata/catalog query through table, only creates/removes relationships of new/old data. 

## Ticket Link

Link to the associated ticket
[ENT-5424](https://openedx.atlassian.net/browse/ENT-5424)

## Post-review

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -60,7 +60,7 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
-    command: bash -c 'cd /edx/app/enterprise_catalog/enterprise_catalog && celery -A enterprise_catalog worker -l DEBUG'
+    command: bash -c 'while true; do cd /edx/app/enterprise_catalog/enterprise_catalog && celery -A enterprise_catalog worker -l DEBUG; sleep 2; done'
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -60,7 +60,7 @@ services:
     build:
       context: .
       dockerfile: Dockerfile
-    command: bash -c 'cd /edx/app/enterprise_catalog/enterprise_catalog && celery -A enterprise_catalog worker -l DEBUG'
+    command: bash -c 'while true; do cd /edx/app/enterprise_catalog/enterprise_catalog && celery -A enterprise_catalog worker -l DEBUG; sleep 2; done'
```
johnnagro, line 63: note to other reviewers - this simply keeps the container running so you can shell in and fix requirements, etc.

johnnagro (approved):

docker-compose.yml
alex-sheehan-edx: * specifically the worker container, the app container already has this behavior

PR #427, opened 2022-03-24
fix: account for duplicate content metadata found when associating metadata with queries

alex-sheehan-edx:
## Description

Tasks that associated content metadata with queries were generating duplicate records of content, which was in turn causing a key error when trying to remove that piece of content from the set of content to remove.

## Ticket Link

Link to the associated ticket
[Enterprise-catalog Job is failing](https://openedx.atlassian.net/browse/ENT-5596)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -729,7 +729,12 @@ def generate_content_metadata_diff(metadata):
         existing_metadata_by_key = {metadata.content_key: metadata for metadata in existing_metadata}
 
         for key in existing_metadata_by_key.keys():
-            potential_items_to_delete.pop(key)
+            try:
+                del potential_items_to_delete[key]
+            except KeyError as e:
+                LOGGER.warning(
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -729,7 +729,13 @@ def generate_content_metadata_diff(metadata):
         existing_metadata_by_key = {metadata.content_key: metadata for metadata in existing_metadata}
 
         for key in existing_metadata_by_key.keys():
-            potential_items_to_delete.pop(key)
+            try:
+                del potential_items_to_delete[key]
+                LOGGER.info(f'Found content: {key}, removing it from list of items to delete')
```
johnnagro: not sure we need this info line as it will be pretty chatty

johnnagro (approved):

enterprise_catalog/apps/catalog/models.py
alex-sheehan-edx: that's why I made it an info, I was thinking it could be useful to know if/when we get duplicates because we would see the key come up once in this log and once in the other

PR #431, opened 2022-04-05
feat: adds a curation app, models, and API endpoints for highlights

iloveagent57:
## Description

Adds a `curation` app to hold models that support the Content Highlights feature.

## Ticket Link

https://openedx.atlassian.net/browse/ENT-5666

## Post-review


enterprise_catalog/apps/catalog/tests/test_rules.py
```diff
@@ -33,11 +33,14 @@ def setUp(self):
 
     @mock.patch('enterprise_catalog.apps.catalog.rules.crum.get_current_request')
     @ddt.data(
-        ('catalog.has_admin_access', SYSTEM_ENTERPRISE_ADMIN_ROLE, TEST_ENTERPRISE_UUID),
-        ('catalog.has_admin_access', SYSTEM_ENTERPRISE_ADMIN_ROLE, ALL_ACCESS_CONTEXT),
+        ('catalog.has_admin_access', SYSTEM_ENTERPRISE_LEARNER_ROLE, TEST_ENTERPRISE_UUID),
+        ('catalog.has_admin_access', SYSTEM_ENTERPRISE_LEARNER_ROLE, ALL_ACCESS_CONTEXT),
```
pwnage101, line 37: Let me know if this makes sense.  I removed `SYSTEM_ENTERPRISE_ADMIN_ROLE` from this test to reflect the newly escalated privileges added in this PR (added `ENTERPRISE_CATALOG_ADMIN_ROLE` to `SYSTEM_ENTERPRISE_ADMIN_ROLE` in the `SYSTEM_TO_FEATURE_ROLE_MAPPING`)

Also it seemed logical to keep this test to make sure `SYSTEM_ENTERPRISE_LEARNER_ROLE` does not have implicit admin access even if the provided context matches the object under test, since this specific case doesn't seem to be covered.

.annotation_safe_list.yml
```diff
@@ -18,11 +18,13 @@ catalog.CatalogUpdateCommandConfig:
...
   ".. no_pii": "This model has no PII"
 catalog.HistoricalContentMetadata:
-    ".. no_pii": "This model has no PII"
+  ".. no_pii": "This model has no PII"
 catalog.HistoricalEnterpriseCatalog:
   ".. no_pii": "This model has no PII"
 contenttypes.ContentType:
   ".. no_pii:": "This model has no PII"
+curation.HistoricalEnterpriseCurationConfig:
+  ".. no_pii:": "This model has no PII"
```
pwnage101, line 27: I don't understand why the pii checker step is harping on this model because I added the annotation here.  Perhaps I'm using the wrong identifier?

macdiesel: so strange... Double checked the package name/path?

Makefile
```diff
@@ -191,6 +191,9 @@ mysql-client:  # Opens mysql client in the mysql container shell
 %-restart: # Restart the specified service container
 	docker-compose restart $*
 
+app-restart-devserver: ## Kill the enterprise-catalog development server. Watcher should restart it.
```
macdiesel: I would suggest `%-restart-devserver` here.

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -2066,3 +2066,33 @@ def test_catalogs_different_uuids(self, use_different_query):
         else:
             assert response['num_distinct_query_ids'] == 1
         assert str(self.catalog_query_one.id) in response['catalog_uuids_by_catalog_query_id']
+
+
+class EnterpriseCurationConfigReadOnlyViewSetTests(APITestMixin):
+    """
+    TODO
```
macdiesel: Needs docs.

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -2066,3 +2066,33 @@ def test_catalogs_different_uuids(self, use_different_query):
...
+            catalog_query=self.catalog_query_one,
+        )
+
+    # @ddt.data(
+    #     False,
+    #     True,
+    # )
+    def test_unauthorized_list(self):
+        """
+        TODO
```
macdiesel: Needs docs.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -272,3 +277,83 @@ def _add_course_run_enrollment_urls(self, course_instance, serialized_course_run
 
         for serialized_run in serialized_course_runs:
             serialized_run['enrollment_url'] = urls_by_course_run_key.get(serialized_run['key'])
+
```
macdiesel, line 280: What I like about the following Serializers... there is no business logic in them.

enterprise_catalog/apps/api/v1/urls.py
```diff
@@ -43,6 +49,10 @@
 router.register(r'enterprise-catalogs', EnterpriseCatalogCRUDViewSet, basename='enterprise-catalog')
 router.register(r'enterprise-catalogs', EnterpriseCatalogContainsContentItems, basename='enterprise-catalog')
 router.register(r'enterprise-customer', EnterpriseCustomerViewSet, basename='enterprise-customer')
+router.register(r'enterprise-curations', EnterpriseCurationConfigReadOnlyViewSet, basename='enterprise-curations')
+router.register(r'enterprise-curations-admin', EnterpriseCurationConfigViewSet, basename='enterprise-curations-admin')
```
macdiesel, line 53: Does this line and the line 2 down from this effectively mean that admins hit one url endpoint to do the actions and the learners hit a different one?

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -0,0 +1,307 @@
...
+    permission_required = PERMISSION_HAS_ADMIN_ACCESS
+
+    # Fields required for controlling access in the `list()` action
+    allowed_roles = [ENTERPRISE_CATALOG_ADMIN_ROLE]
+
+    def create(self, request, *args, **kwargs):
+        """ Create a new EnterpriseCurationConfig """
+        if not self.requested_enterprise_uuid:
+            return Response(
+                f'An enterprise UUID was not specified.',
```
macdiesel, line 125: I appreciate the specificity of these response messages.

enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -73,7 +75,8 @@ def json_metadata(self):
         }
         if self.content_type == COURSE:
             owners = [{
-                'name': 'Partner Name',
+                'uuid': str(FAKE_CONTENT_AUTHOR_UUID),
```
macdiesel, line 78: FWIW I think this is fine but we could start using faker for things like this: https://faker.readthedocs.io/en/master/

enterprise_catalog/apps/curation/admin.py
```diff
@@ -0,0 +1,54 @@
...
+class HighlightSetAdmin(SimpleHistoryAdmin):
+    """ Admin config for HighlightSet model. """
+    list_display = (
+        'uuid',
+        'title',
+        'enterprise_curation',
+        'is_published',
+    )
+    search_fields = (
+        'uuid',
```
macdiesel, line 30: Please make title a searchable field.

enterprise_catalog/apps/curation/admin.py
```diff
@@ -0,0 +1,54 @@
...
+class EnterpriseCurationConfigAdmin(SimpleHistoryAdmin):
+    """ Admin configuration for the EnterpriseCurationConfig model. """
+    list_display = (
+        'uuid',
+        'enterprise_uuid',
+        'title',
+        'is_highlight_feature_active',
+    )
+    search_fields = (
+        'uuid',
```
macdiesel, line 16: Please make title a searchable field.

enterprise_catalog/apps/curation/admin.py
```diff
@@ -0,0 +1,54 @@
+from django.contrib import admin
```
macdiesel, line 1: for extra credit we've started using [djangoql](https://github.com/ivelum/djangoql) in license manager. If we have some capacity I would like to introduce it to catalog as well.

macdiesel (commented):
A few thoughts.

enterprise_catalog/apps/api/v1/urls.py
adamstankiewicz: @macdiesel Yes, this is a pattern we do for other enterprise endpoints. the learner endpoints are "read-only" and the admin endpoints are "read/write" for its permissions.

enterprise_catalog/apps/catalog/tests/test_rules.py
adamstankiewicz: @pwnage101 Would it make sense to keep both `SYSTEM_ENTERPRISE_LEARNER_ROLE` and `SYSTEM_ENTERPRISE_ADMIN_ROLE`?

enterprise_catalog/apps/curation/admin.py
adamstankiewicz: +1, I think it's worth pulling in proactively for newer admin screens.

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -0,0 +1,307 @@
...
+                status=HTTP_400_BAD_REQUEST
+            )
+        request.data['enterprise_curation'] = str(curation_config.uuid)
+
+        return super().create(request, *args, **kwargs)
+
+    @action(detail=True, methods=['post'], url_path='add-content')
+    def add_content(self, request, uuid, *args, **kwargs):
+        """ Add content to an existing HighlightSet """
+        content_keys = set(request.data.get('content_keys', []))
```
adamstankiewicz: nit: should/could we be using `get_content_keys_from_request_data` here?

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -0,0 +1,307 @@
...
+                'existing_content_keys': existing_content_keys,
+                'highlight_set': HighlightSetSerializer(highlight_set).data,
+            },
+            status=HTTP_201_CREATED,
+        )
+
+    @action(detail=True, methods=['post'], url_path='remove-content')
+    def remove_content(self, request, uuid, *args, **kwargs):
+        """ Remove existing content from an existing HighlightSet """
+        content_keys = set(request.data.get('content_keys', []))
```
adamstankiewicz: nit: should/could we be using `get_content_keys_from_request_data` here?

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -0,0 +1,307 @@
...
+        highlighted_content = []
+        ignored_content_keys = []
+        added_content_keys = []
+        existing_content_keys = []
+        for content_key in content_keys:
+            try:
+                content_metadata = ContentMetadata.objects.get(content_key=content_key)
+            except ContentMetadata.DoesNotExist:
+                # TODO: log error
```
adamstankiewicz: nit: I think it's worth this TODO in this PR?

enterprise_catalog/apps/curation/admin.py
```diff
@@ -0,0 +1,54 @@
...
+    """ Admin config for HighlightedContent model. """
+    list_display = (
+        'uuid',
+        'catalog_highlight_set',
+        'content_metadata',
+    )
+    search_fields = (
+        'uuid',
+        'catalog_highlight_set',
+        'content_metadata',
```
adamstankiewicz, line 46: [question] What are the fields being searched on for `catalog_highlight_set` and `content_metadata` here (e.g., title of the associated `content_metadata` objects)? similar comment to Brian, might be worth pulling in `djangoql` for these admin classes.

Makefile
pwnage101: Since `%-restart` already exists, I'd propose we just remove this line altogether.

As written, it only works for django services, not memcache/mysql/etc. My guess is that it just performs better than `%-restart` (since it does not take down the container), but serves the same function.  At least on hosted devstack, I've found that `app-restart` performs well enough.

enterprise_catalog/apps/catalog/tests/factories.py
pwnage101: The reason for not using faker here is because the UUID needs to be deterministic so that it can be recalled in the [unit test logic](https://github.com/openedx/enterprise-catalog/pull/431/files#diff-8d0c382d3777dad3851ca433546be57a60386bcbd35ad5ca1c9b96757ac0b946R38).  I'm open to using a faker if there's a way to also retreive the expected UUID in the unit test.

enterprise_catalog/apps/catalog/tests/test_rules.py
pwnage101: If I were to add `SYSTEM_ENTERPRISE_ADMIN_ROLE` back into this test, then it would begin to fail again; I fixed the test by removing that system role from the ddt.  I think it made sense for that test to fail, since the change in this PR to map `SYSTEM_ENTERPRISE_ADMIN_ROLE` to `ENTERPRISE_CATALOG_ADMIN_ROLE` would effectively grant implicit `catalog.has_admin_access` to that system role, conflicting with this test which asserts that it does NOT have implicit access.

Adding `SYSTEM_ENTERPRISE_LEARNER_ROLE` was actually unrelated to the failing test.  I just realized it was missing.

.annotation_safe_list.yml
pwnage101: Already triple checked, and dug a little into the code-annotations source code to no avail.  I need more eyes on this, I'll just ask around outside the team.

pwnage101: for posterity: Fixed this by renaming the `__init.py__` file to `__init__.py`

enterprise_catalog/apps/curation/admin.py
pwnage101: deferring to subsequent PR

enterprise_catalog/apps/catalog/tests/test_rules.py
adamstankiewicz: @pwnage101 Ah, yes. Thanks for the clarifications. Your understanding is correct and your corresponding change here makes sense; carry on :)

.annotation_safe_list.yml
adamstankiewicz: @pwnage101 Good find on the fix here!

Makefile
adamstankiewicz: I, too, usually just use `app-restart` (on standard devstack, not hosted). This is a remnant from Alex D. initial work on this branch; wasn't sure the original intent for it.

@pwnage101 I'd be in favor of removing it altogether; maybe just sanity check with Alex D. to verify he didn't add it for an explicit reason?

iloveagent57: Yup, it's much faster to restart a Django devserver locally than it is to restart the container.  Feel free to remove, I'll add it back with documented justification if I get frustrated with slowness in the future. :)

adamstankiewicz (approved):

PR #443, opened 2022-04-27
fix: add trailing slash to distinct-catalog-queries path

long74100:
## Description

Add trailing slash in path.

## Ticket Link

Link to the associated ticket
[Link title](https://openedx.atlassian.net/browse/ENT-5770)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/urls.py
+++ b/enterprise_catalog/apps/api/v1/urls.py
@@ -72,8 +72,8 @@ urlpatterns = [
         EnterpriseCatalogRefreshDataFromDiscovery.as_view({'post': 'post'}),
         name='update-enterprise-catalog'
     ),
-    path('distinct-catalog-queries', DistinctCatalogQueriesView.as_view(),
-         name='distinct-catalog-queries',
+    path('distinct-catalog-queries/', DistinctCatalogQueriesView.as_view(),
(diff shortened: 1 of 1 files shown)


macdiesel (approved):

enterprise_catalog/apps/api/v1/urls.py
```diff
@@ -72,9 +72,9 @@
         EnterpriseCatalogRefreshDataFromDiscovery.as_view({'post': 'post'}),
         name='update-enterprise-catalog'
     ),
-    path('distinct-catalog-queries', DistinctCatalogQueriesView.as_view(),
+    path('distinct-catalog-queries/', DistinctCatalogQueriesView.as_view(),
```
adamstankiewicz, line 75: [curious/clarification] Curious why trailing slash is needed here but not above with `catalog_workbook` or `generate_diff` or `refresh_metadata`, etc. or am I reading this wrong haha?

adamstankiewicz (approved):

enterprise_catalog/apps/api/v1/urls.py
long74100: They should also have a trailing slash, I added an item to our tech debt tracker so we can revisit this. We would also want to set APPEND_SLASH to true so that trailing slashes automatically get appended.

PR #448, opened 2022-05-12
fix: filter exec-ed/bootcamp data from enterprise catalog service

johnnagro:
## Description

New 2u content types like GetSmarter Exec. Ed and Bootcamps have been added to Discovery as `course` content types with new `course_type` values. Enterprise Catalogs haven't had to filter based on `course_type` and so many have started to erroneously include these new content types. As a result, exec ed courses are appearing in enterprise catalogs in the explore catalog, learner portal, and LMS integration synced content. Learners are discovering this content.

This change filters these new courses via an allow-list until we're ready to address them as an enterprise product.

## References

- [ENT-5823](https://2u-internal.atlassian.net/browse/ENT-5823)

### DIFF
--- a/enterprise_catalog/apps/catalog/constants.py
+++ b/enterprise_catalog/apps/catalog/constants.py
@@ -18,6 +18,28 @@ CONTENT_TYPE_CHOICES = [
     (LEARNER_PATHWAY, 'Learner Pathway'),
 ]
 
+# ContentMetadata allow/block lists
+
+# deliberate omissions:
+# 'executive-education'
+# 'executive-education-2u'
+# 'bootcamp-2u'
+# 'empty'
+
+CONTENT_COURSE_TYPE_ALLOW_LIST = [
+    'audit',
+    'professional',
+    'verified-audit',
+    'credit-verified-audit',
+    'masters',
+    'masters-verified-audit',
+    'verified',
+    'spoc-verified-audit',
+    'honor',
+    'verified-honor',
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/catalog/constants.py
```diff
@@ -18,6 +18,28 @@
...
 
+# ContentMetadata allow/block lists
+
+# deliberate omissions:
+# 'executive-education'
+# 'executive-education-2u'
+# 'bootcamp-2u'
+# 'empty'
+
+CONTENT_COURSE_TYPE_ALLOW_LIST = [
```
johnnagro, line 29: list derived from the production discovery database

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -18,6 +18,28 @@
     (LEARNER_PATHWAY, 'Learner Pathway'),
 ]
 
+# ContentMetadata allow/block lists
+
+# deliberate omissions:
```
johnnagro, line 23: these are some of the keys we've seen which we wish to exclude

enterprise_catalog/apps/catalog/constants.py
johnnagro: an allow-list (rather than a block-list) will prevent new course content types from accidentally slipping in again at the cost of us needing to maintain this list.

iloveagent57 (approved):

PR #452, opened 2022-05-17
feat: Filter recommendations based on specified catalogs or all enterprise catalogs

sameenfatima78:
## Description

This PR filters the course recommendations shown to a learner based on their enterprise catalogs

## Ticket Link

Link to the associated ticket
[ENT-5735](https://2u-internal.atlassian.net/browse/ENT-5735)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views/enterprise_customer.py
+++ b/enterprise_catalog/apps/api/v1/views/enterprise_customer.py
@@ -102,3 +102,27 @@ class EnterpriseCustomerViewSet(BaseViewSet):
         if (get_catalogs_containing_specified_content_ids or get_catalog_list):
             response_data['catalog_list'] = catalogs_that_contain_course
         return Response(response_data)
+
+    @action(detail=True, methods=['post'])
+    def filter_content_items(self, request, enterprise_uuid, **kwargs):
+        """
+        Filters the content items based on catalogs passed or all catalogs belonging to an enterprise.
+        """
+        customer_catalogs = request.data.get('catalog_uuids', [])
+        content_keys = set(request.data.get('content_keys', []))
+        if customer_catalogs:
+            customer_catalogs = EnterpriseCatalog.objects.filter(uuid__in=customer_catalogs)
+        else:
+            customer_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
+
+        filtered_content_keys = set()
+        for catalog in customer_catalogs:
+            items_included = catalog.filter_content_keys(content_keys)
+            if items_included:
+                filtered_content_keys = filtered_content_keys.union(items_included)
+
+        response_data = {
+            'filtered_content_keys': list(filtered_content_keys),
+        }
+
+        return Response(response_data)
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -309,6 +309,43 @@ class EnterpriseCatalog(TimeStampedModel):
         # if the filtered content metadata exists, the specified content_keys exist in the catalog
         return self.content_metadata.filter(query).exists()
 
+    def filter_content_keys(self, content_keys):
+        """
+        Determines whether content_keys are part of the catalog.
+
+        Arguments:
+            content_keys: (set) A set of string content keys to be filtered based on the catalog.
+
+        Returns:
+            items_included: (set) A filtered set of content keys contained in the catalog.
+
+        This method handles the following scenarios:
+          - associated metadata contains the specified content key.
+          - associated metadata contains the specified content key as a parent (when a catalog only contains
+           course runs but a course id is searched).
+        """
+        # cannot determine if specified content keys are part of catalog when catalog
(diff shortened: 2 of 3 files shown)


enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -102,3 +102,34 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
         if (get_catalogs_containing_specified_content_ids or get_catalog_list):
             response_data['catalog_list'] = catalogs_that_contain_course
         return Response(response_data)
+
+    @action(detail=True, methods=['post'])
+    def filter_content_items(self, request, enterprise_uuid, **kwargs):
+        """
+        Filters the content recommendations based on catalogs passed or all catalogs belonging to an enterprise.
+        """
+        customer_catalogs = request.data.pop('catalog_uuids', [])
```
saleem-latif: I think we should use `get` here instead of `pop`

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -102,3 +102,34 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
+        all_recommendations = request.data.pop('all_recommendations', [])
+        same_partner_recommendations = request.data.pop('same_partner_recommendations', [])
+        if customer_catalogs:
+            customer_catalogs = EnterpriseCatalog.objects.filter(uuid__in=customer_catalogs)
+        else:
+            customer_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
+
+        filtered_all_recommendations = []
```
saleem-latif: I think we should use `set` here as well if we want to return unique items in the end.
```
filtered_all_recommendations = set()
...
 filtered_all_recommendations.update(items_included_all)
```

enterprise_catalog/apps/catalog/models.py
```diff
@@ -309,6 +309,45 @@ def contains_content_keys(self, content_keys):
         # if the filtered content metadata exists, the specified content_keys exist in the catalog
         return self.content_metadata.filter(query).exists()
 
+    def filter_content_recommendations(self, all_recommendations, same_partner_recommendations):
+        """
+        Determines whether recommendations are part of the catalog.
```
saleem-latif: I think we should add `Arguments` and `Returns` sections in the doc string and explain the type of each and the purpose of arguments.

saleem-latif (approved):
Just a few comments, once those are addressed 👍🏼

enterprise_catalog/apps/catalog/models.py
```diff
@@ -309,6 +309,45 @@ def contains_content_keys(self, content_keys):
         # if the filtered content metadata exists, the specified content_keys exist in the catalog
         return self.content_metadata.filter(query).exists()
 
+    def filter_content_recommendations(self, all_recommendations, same_partner_recommendations):
```
irfanuddinahmad: We should call this function with a single argument twice ... rather than with these two similar arguments ...

enterprise_catalog/apps/catalog/models.py
```diff
@@ -309,6 +309,45 @@ def contains_content_keys(self, content_keys):
...
+        # where content_key and parent_content_key matches the specified content_keys to
+        # handle the following cases where the catalog:
+        #   - contains courses and the specified content_keys are course ids
+        #   - contains course runs and the specified content_keys are course ids
+        query = Q(content_key__in=content_keys) | Q(parent_content_key__in=content_keys)
+
+        items_included_all = []
+        items_included_same_partner = []
+        for content in self.content_metadata.filter(query).all():
+            if content.content_key in all_recommendations:
```
irfanuddinahmad: DRY issue here ...

irfanuddinahmad (changes requested):

saleem-latif (approved):

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -102,3 +102,27 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
+        if customer_catalogs:
+            customer_catalogs = EnterpriseCatalog.objects.filter(uuid__in=customer_catalogs)
+        else:
+            customer_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
+
+        filtered_content_keys = set()
+        for catalog in customer_catalogs:
+            items_included = catalog.filter_content_keys(content_keys)
+            if items_included:
+                filtered_content_keys = content_keys.intersection(items_included)
```
irfanuddinahmad: I think the intersection will always return items_included ... why do we need this operation?

irfanuddinahmad (approved):
LGTM!

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
sameenfatima78: Good catch... fixing it!

PR #464, opened 2022-06-07
fix: resolve formatting issues for django admin

bcitro:
## Description

Removes the filters from the enterprise catalogs view which were blocking the entire UI

## Ticket Link

https://2u-internal.atlassian.net/browse/ENT-5858

## Post-review


johnnagro (approved):
⛵

PR #469, opened 2022-06-16
build: add backup/restore make targets for mysql

iloveagent57:


Makefile
```diff
@@ -173,6 +173,12 @@ dev.destroy: dev.down # Kills containers and destroys volumes. If you get an err
 dev.stop: # Stops containers so they can be restarted
 	docker-compose stop
 
+dev.backup: dev.up
+	docker run --rm --volumes-from enterprise.catalog.mysql -v $$(pwd)/.dev/backups:/backup debian:jessie tar zcvf /backup/mysql57.tar.gz /var/lib/mysql
```
johnnagro: nit: i believe this project uses mysql 8 rather than 57, maybe adjust the name of the backup file. reference https://github.com/openedx/enterprise-catalog/blob/master/docker-compose.yml#L4

johnnagro (approved):
see my nit

PR #471, opened 2022-06-21
Llin/update enroll url

long74100:
## Description
Temporary band-aid to redirect learners to B2C if they have offers AND subs/codes.

(https://2u-internal.atlassian.net/browse/ENT-5968)

## Post-review


christopappas (approved):

adamstankiewicz (approved):
LGTM :shipit:

enterprise_catalog/apps/api_client/ecommerce.py
```diff
@@ -0,0 +1,38 @@
...
+                'page': 1,
+                'page_size': 50,
+                'filter': 'active'
+            }
+
+        url = COUPONS_OVERVIEW_ENDPOINT.format(enterprise_customer_uuid=enterprise_customer_uuid)
+        response = self.client.get(url, params=query_params)
+        response.raise_for_status()
+
+        results = response.json()['results']
```
macdiesel, line 42: Could use some exception handling here... Maybe even validate the that enterprise UUID is a valid UUID?

enterprise_catalog/apps/api_client/enterprise_cache.py
```diff
@@ -41,6 +51,14 @@ def slug(self):
         """
         return self.customer_data.get('slug', '')
 
+    @property
+    def active_catalogs(self):
+        """
+        Return catalogs associated with active subisides for the enterprise customer.
+        """
+        catalogs = self.customer_data.get('coupons_catalogs', []) + self.customer_data.get('subscriptions_catalogs', [])
```
macdiesel, line 60: Cache file but no cache on this method?

enterprise_catalog/apps/api_client/enterprise_cache.py
```diff
@@ -63,10 +81,32 @@ def _get_enterprise_customer_data(uuid):
     """
     cache_key = ENTERPRISE_CUSTOMER_CACHE_KEY_TPL.format(uuid=uuid)
     customer_data = cache.get(cache_key)
+
     if not isinstance(customer_data, dict):
-        client = EnterpriseApiClient()
-        customer_data = client.get_enterprise_customer(uuid)
+        enterprise_client = EnterpriseApiClient()
+        customer_data = enterprise_client.get_enterprise_customer(uuid)
```
macdiesel, line 88: This area and the method called are all light on the error handling.

enterprise_catalog/settings/base.py
```diff
@@ -363,6 +363,8 @@
 LMS_BASE_URL = os.environ.get('LMS_BASE_URL', '')
 DISCOVERY_SERVICE_API_URL = os.environ.get('DISCOVERY_SERVICE_API_URL', '')
 ENTERPRISE_LEARNER_PORTAL_BASE_URL = os.environ.get('ENTERPRISE_LEARNER_PORTAL_BASE_URL', '')
+ECOMMERCE_BASE_URL = os.environ.get('ECOMMERCE_BASE_URL', '')
+LICENSE_MANAGER_BASE_URL = os.environ.get('LICENSE_MANAGER_BASE_URL', '')
```
macdiesel, line 367: Safe to assume a forthcoming edx-internal PR for these variables for stage/prod?

macdiesel (commented):
I think it would be good to improve the error handling and logging so we can troubleshoot this a bit better should something go sideways.

enterprise_catalog/apps/api_client/enterprise_cache.py
long74100: These properties don't need to be cached since the instance doesn't persist. The cached data is loaded in to create an instance tho.

enterprise_catalog/apps/api_client/ecommerce.py
long74100: I don't think we do this type of argument validation for any of our clients, I could add some logging to these methods and reraise.

PR #485, opened 2022-08-03
feat: allowing catalog queries to allow for exec ed courses to be included

alex-sheehan-edx:
new query model attr that allows exec ed courses to be ingested and associated with catalogs

## Ticket Link

Link to the associated ticket
[Allow Exec Ed courses to be included for enterprise catalogs](https://2u-internal.atlassian.net/browse/ENT-6042)

## Post-review


enterprise_catalog/apps/catalog/models.py
```diff
@@ -697,7 +705,8 @@ def create_content_metadata(metadata):
         content_keys = []
         filtered_batched_metadata = []
         for entry in batched_metadata:
-            if _should_allow_metadata(entry):
+            # Exclude exec ed courses from being ingested unless the query specifies that they are allowed
```
kiram15, line 712: I'm not the authority on this, but it does feel weird to put kind of hardcoded check at this base level, though I understand if this is the best way to go about this. Will we always have this in there, and will we be excluding other things in the future in this way?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -697,7 +705,8 @@ def create_content_metadata(metadata):
...
+            if (catalog_query and catalog_query.include_exec_courses) or _should_allow_metadata(entry):
```
iloveagent57: To make the logic around the field match the intent of the field's name, I think you want to move this check into `_should_allow_metadata()` and add an if clause like
```
if course_type == 'executive-education-2u' and catalog_query.include_exec_courses
```
....something like that.  The way this is written, `catalog_query.include_exec_courses` being true really means "let any kind of content into the catalog".

enterprise_catalog/apps/catalog/models.py
```diff
@@ -84,6 +84,13 @@ class CatalogQuery(TimeStampedModel):
         max_length=100
     )
 
+    include_exec_courses = models.BooleanField(
```
iloveagent57: nit: maybe `include_exec_ed_2u_courses`....there's another type of course in discovery called plainly `executive education` that we might want to disambiguate from.

enterprise_catalog/apps/catalog/models.py
bcitro: Yeah I'm not sure this is the best place for the check - I could see us expanding the `_should_allow_metadata` function in the future with other checks besides just course type, at which point this logic wouldn't work as we intended (i.e., at that point we wouldn't want to skip calling `_should_allow_metadata` just because `include_exec_courses` was true).

I think this will also allow other 2U course types which we explicitly excluded (bootcamps, degrees, etc) to start sneaking into catalogs.

Maybe it makes sense to move the `include_exec_courses` check into `_should_allow_metadata` and update the logic there instead?

iloveagent57:
Quick idea for making this even more generalized.
1. Add something like this to Django settings.
```python
# maps field names on instances of the CatalogQuery model
# to course types that are allowed to be associated with
# that CatalogQuery instance.  The instance field must be set
# to true to allow the content with the course type to be
# associated with the query.
CATALOG_QUERY_COURSE_TYPE_ALLOW_LIST = {
    'allow_exec_ed_2u_courses': constants. EXEC_ED_2U_COURSE_TYPE,
}
```
2. Update `_should_allow_metadata()` line to be
```python
for field_name, course_type in settings.CATALOG_QUERY_COURSE_TYPE_ALLOW_LIST.items():
    if catalog_query and getattr(catalog_query, field_name, None) and entry_course_type == course_type:
        return True
```
Then it's a little more data-driven and we'll have a pattern to do this for future course types.  ~~And we can allow new course types without deployments.~~ I guess we'd still need to deploy a migration to add a field to the model.  We could generalize further to avoid a migration, but I'm not sure it's worth it.  Still think it's worth having the pattern/settings.

iloveagent57:
after offline discussion, we decided that my suggestion above leaned a little toward the YAGNI side of the spectrum.

iloveagent57 (approved):
Your codecov robot overlords also want you to write tests :P

PR #488, opened 2022-08-08
fix: refactor enrollment url creation

iloveagent57:
## Description

Compute the `enrollment_url` field from `ContentMetadata` instances, rather than serializations thereof.  So now we can do this: `enterprise_catalog.get_content_enrollment_url(content_metadata_record)` instead of passing in various field values from a ContentMetadata record.

Why do this?
In the near future, I want to be able to check if a related CatalogQuery allows `exec_ed_2u` types of courses to be present, as a safety mechanism before including an enrollment URL, e.g.
```python
if content_metadata.course_type == 'exec_ed_2u' and not self.catalog_query.allow_exec_ed_2u:
    return None
# ...otherwise compute the enrollment URL
```
Furthermore, it helps clean up the abstraction layer that previously was defined in the serializer's `to_representation` - that method had to know to pluck the resource type, content key, and parent content key out of the content metadata instance. With this change, we can take a catalog instance, pass a content_metadata record into it, and expect an enrollment URL to be returned - without having to know, as the caller, what the various components that determine the enrollment URL are.

FWIW, I verified (via script) that every program content metadata record that currently exists in our production environment has an `enrollment_url` set to null.  This is because program metadata records do not have a `key` field in their JSON metadata.  Therefore, I consider it "safe" to explicitly set to null in the lines: 
```python
elif content_type == PROGRAM:
    json_metadata['enrollment_url'] = None
```
(rather than calling `get_content_enrollment_url()` and expecting a null result, as we do today on the master branch).
## Ticket Link
https://2u-internal.atlassian.net/browse/ENT-6150

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/serializers.py
+++ b/enterprise_catalog/apps/api/v1/serializers.py
@@ -17,12 +17,10 @@ from enterprise_catalog.apps.catalog.constants import (
 )
 from enterprise_catalog.apps.catalog.models import (
     CatalogQuery,
+    ContentMetadata,
     EnterpriseCatalog,
 )
-from enterprise_catalog.apps.catalog.utils import (
-    get_content_filter_hash,
-    get_parent_content_key,
-)
+from enterprise_catalog.apps.catalog.utils import get_content_filter_hash
 
 
 logger = logging.getLogger(__name__)
@@ -190,7 +188,6 @@ class ContentMetadataSerializer(ImmutableStateSerializer):
         json_metadata = instance.json_metadata.copy()
         marketing_url = json_metadata.get('marketing_url')
         content_key = json_metadata.get('key')
-        parent_content_key = get_parent_content_key(json_metadata)
 
         # The enrollment URL field of content metadata is generated on request and is determined by the status of the
         # enterprise customer as well as the catalog. So, in order to detect when content metadata has last been
@@ -211,30 +208,32 @@ class ContentMetadataSerializer(ImmutableStateSerializer):
             json_metadata['marketing_url'] = marketing_url
 
         if content_type in (COURSE, COURSE_RUN):
-            json_metadata['enrollment_url'] = enterprise_catalog.get_content_enrollment_url(
-                content_resource=COURSE,
-                content_key=content_key,
-                parent_content_key=parent_content_key,
-            )
+            json_metadata['enrollment_url'] = enterprise_catalog.get_content_enrollment_url(instance)
             json_metadata['xapi_activity_id'] = enterprise_catalog.get_xapi_activity_id(
                 content_resource=content_type,
                 content_key=content_key,
             )
             if content_type == COURSE:
-                course_runs = json_metadata.get('course_runs', [])
-                json_metadata['active'] = is_any_course_run_active(course_runs)
-                for course_run in course_runs:
-                    course_run['enrollment_url'] = enterprise_catalog.get_content_enrollment_url(
-                        content_resource=COURSE,
-                        content_key=course_run.get('key'),
-                        parent_content_key=content_key,
-                    )
+                serialized_course_runs = json_metadata.get('course_runs', [])
+                json_metadata['active'] = is_any_course_run_active(serialized_course_runs)
+                self._add_course_run_enrollment_urls(instance, serialized_course_runs)
         elif content_type == PROGRAM:
-            # This URL will always be blank because json_metadata['key'] doesn't exist for programs
-            json_metadata['enrollment_url'] = enterprise_catalog.get_content_enrollment_url(
-                content_resource=PROGRAM,
-                content_key=content_key,
-                parent_content_key=parent_content_key,
-            )
+            # We want this to be null, because we have no notion
+            # of directly enrolling in a program.
+            json_metadata['enrollment_url'] = None
 
         return json_metadata
+
+    def _add_course_run_enrollment_urls(self, course_instance, serialized_course_runs):
+        """
+        For the given `course_instance`, computes the enrollment url for each
+        child course run and adds it to the serialized representation of the
+        course run record in `serialized_course_runs`.
+        """
+        urls_by_course_run_key = {}
+        for course_run in ContentMetadata.get_child_records(course_instance):
+            urls_by_course_run_key[course_run.content_key] = \
+                self.context['enterprise_catalog'].get_content_enrollment_url(course_run)
+
+        for serialized_run in serialized_course_runs:
+            serialized_run['enrollment_url'] = urls_by_course_run_key.get(serialized_run['key'])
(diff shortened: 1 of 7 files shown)


enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -211,30 +208,31 @@ def to_representation(self, instance):
...
         return json_metadata
+
+    def _add_course_run_enrollment_urls(self, course_instance, serialized_course_runs):
+        """
+        For the given `course_instance`, computes the enrollment url for each
+        child course run and adds it to the serialized representation of the
+        course run record in `serialized_course_runs`.
+        """
+        urls_by_course_run_key = {}
+        for course_run in ContentMetadata.get_child_records(course_instance):
```
iloveagent57, line 234: Introduces an additional database query, which is fine IMHO because:
* The primary endpoint that hits this code path, `get_content_metadata` is already very performant.
* The average time spent in the DB in these transactions is quite small.
* It's a very simple query (simple filter on the `parent_content_key` field, which is indexed).

enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -58,9 +58,9 @@ class ContentMetadataFactory(factory.django.DjangoModelFactory):
     class Meta:
         model = ContentMetadata
 
-    content_key = factory.Sequence(lambda n: f'metadata_item_{n}')
+    content_key = factory.Sequence(lambda n: f'{str(n).zfill(5)}_metadata_item')
```
iloveagent57, line 61: This is the change that allowed me to get rid of that big comment block in the tests above.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -211,30 +208,31 @@ def to_representation(self, instance):
...
-                    course_run['enrollment_url'] = enterprise_catalog.get_content_enrollment_url(
-                        content_resource=COURSE,
-                        content_key=course_run.get('key'),
-                        parent_content_key=content_key,
-                    )
+                serialized_course_runs = json_metadata.get('course_runs', [])
+                json_metadata['active'] = is_any_course_run_active(serialized_course_runs)
+                self._add_course_run_enrollment_urls(instance, serialized_course_runs)
         elif content_type == PROGRAM:
             # This URL will always be blank because json_metadata['key'] doesn't exist for programs
```
christopappas: is this comment still true? and if so, do we need to do the `get_content_enrollment_url ` computation here still?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -345,73 +345,82 @@ def filter_content_keys(self, content_keys):
...
-            content_key (str): The content key for the course to be displayed.
-            parent_content_key (str): The content key for the course that is parent of the given course run key.
-                                      This argument will be None if a course or program key is passed.
-
+            content_metadata (ContentMetadata): The record for which an enrollment URL is returned.
         Returns:
             (str): Enterprise landing page URL OR Enterprise Learner Portal course page URL.
         """
-        if not (content_key and content_resource):
+        resource_type = PROGRAM if content_metadata.content_type == PROGRAM else COURSE
```
christopappas: [question] content_type can be 4 things (COURSE_RUN, COURSE, PROGRAM, LEARNER_PATHWAY), is it confusing to set the var to COURSE if the content_type was LEARNER_PATHWAY? 🤷

enterprise_catalog/apps/catalog/tests/factories.py
christopappas: TIL zfill

christopappas (approved):
2 tiny questions. i like this refactor 👍

enterprise_catalog/apps/api/v1/serializers.py
iloveagent57: huh, yeah, great question.  lemme do some manual testing on this

iloveagent57: ah, yeah, the master branch has null enrollment URLs for programs - because the `json_metadata['key']` value is null, but my change started fetching the `content_metadata.content_key` instead of the json metadata 'key'.
🔑 all the way ⬇️ 
I'll not compute it here and write a test to enforce this.

enterprise_catalog/apps/catalog/models.py
iloveagent57: Yes! 👍 I made the condition more sane to read, i.e.
```python
if content_metadata.content_type not in (COURSE, COURSE_RUN):
    return None
```

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -211,30 +208,32 @@ def to_representation(self, instance):
...
         elif content_type == PROGRAM:
-            # This URL will always be blank because json_metadata['key'] doesn't exist for programs
-            json_metadata['enrollment_url'] = enterprise_catalog.get_content_enrollment_url(
-                content_resource=PROGRAM,
-                content_key=content_key,
-                parent_content_key=parent_content_key,
-            )
+            # We want this to be null, because we have no notion
+            # of directly enrolling in a program.
+            json_metadata['enrollment_url'] = None
```
adamstankiewicz, line 223: If we already having `null` `enrollment_url` for programs, maybe we treat the addition of enrollment urls as a separate smaller task.

that said, technically the "enrollment_url" today for a program would just be the program about page in the Enterprise Learner Portal, e.g. https://enterprise.edx.org/bnr-codes/program/e250b08e-be2a-4b48-a43c-c9ae47f6cf14, which is dependent on the program UUID. Maybe something we can clarify or get onto the roadmap for Markhors who implemented programs in the LP, if we intended to keep this `null` or not.

enterprise_catalog/apps/api_client/enterprise_cache.py
```diff
@@ -65,7 +65,9 @@ def last_modified_date(self):
         """
         Return Enterprise Customer last modified datetime or None if unavailable.
         """
-        return parser.parse(self.customer_data.get('modified', None))
+        if modified_value := self.customer_data.get('modified', None):
```
adamstankiewicz, line 68: :walrus:

adamstankiewicz (approved):

enterprise_catalog/apps/api/v1/serializers.py
iloveagent57: Yeah, agreed there's a "functional" enrollment URL for programs, as you said.  In the learner portal program search cards, that's computed in https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/702d4b13977ca1a8d23f8a008725926aab4ff5a9/src/components/search/SearchProgramCard.jsx#L55-L63
I did ping Markhors with a general question about this.

enterprise_catalog/apps/api_client/enterprise_cache.py
iloveagent57: Haha, yes!  I should mention that this change is included due to a local defect I've seen, where the `modified` field for enterprise customers somehow isn't making it onto these objects.  I don't believe we've encountered this in stage or prod environments.

PR #490, opened 2022-08-15
feat: create enrollment urls for exec ed 2u courses

iloveagent57:
https://2u-internal.atlassian.net/browse/ENT-6150

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/tests/factories.py
+++ b/enterprise_catalog/apps/catalog/tests/factories.py
@@ -32,7 +32,7 @@ class CatalogQueryFactory(factory.django.DjangoModelFactory):
         model = CatalogQuery
 
     content_filter = factory.Dict({'content_type': factory.Faker('words', nb=3)})
-    title = factory.Faker('word')
+    title = factory.Faker('text')
 
 
 class EnterpriseCatalogFactory(factory.django.DjangoModelFactory):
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/catalog/tests/test_models.py
```diff
@@ -243,9 +261,29 @@ def test_contentmetadata_update_from_discovery_ignore_exec_ed(self, mock_client)
...
+        with mock.patch(path + 'EnterpriseApiClient') as mock_enterprise_api_client, \
+             mock.patch(path + 'LicenseManagerApiClient') as mock_license_manager_client, \
+             mock.patch(path + 'EcommerceApiClient') as mock_ecommerce_client:
+            mock_enterprise_api_client.return_value.get_enterprise_customer.return_value = \
+                mock_enterprise_customer_return_value
+            mock_ecommerce_client.return_value.get_coupons_overview.return_value = \
+                mock_coupon_overview_return_value
+            mock_license_manager_client.return_value.get_customer_agreement.return_value = \
+                mock_customer_agreement_return_value
+            yield
```
iloveagent57, line 285: I thought I would need this in multiple tests, but turns out it's still only used in the original test from which it was factored.  But I like this better that how it was before, so I'm leaving it.

enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -43,9 +43,9 @@ class Meta:
         model = EnterpriseCatalog
 
     uuid = factory.LazyFunction(uuid4)
-    title = factory.Faker('word')
+    title = factory.Faker('text')
```
iloveagent57, line 46: Changed these to lessen probability of a UNIQUE constraint violation.

long74100 (approved):

PR #492, opened 2022-08-17
DRAFT: filtering bad course data out of the update contentmetadata from discovery task

alex-sheehan-edx:
I'm wondering if we should read the query and check if the course is valid based on the queries specific `availability` and `status` or if we should just have a base "there needs to be a minimum requirement for the content to be added to the catalog service" like I have now.

## Ticket Link

Link to the associated ticket
https://2u-internal.atlassian.net/browse/ENT-6041

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -116,6 +116,38 @@ class CatalogQuery(TimeStampedModel):
         except cls.DoesNotExist:
             return None
 
+    def is_course_run_content_valid(self, course_run_content):
+        """
+        Check if a piece of course run content (JSON blob) adheres to a query's filter for availability and status
+        """
+        if self.content_filter.get('availability'):
+            if not course_run_content.get('availability') or \
+            course_run_content.get('availability') not in self.content_filter.get('availability'):
+                return False
+
+        if self.content_filter.get('status'):
+            if not course_run_content.get('status') or \
+            course_run_content.get('status') != self.content_filter.get('status'):
+                return False
+
+        return True
+
+    def is_course_content_valid(self, course_content):
+        """
+        Check if a piece of course content (JSON blob) has course runs that adheres to a query's filter for availability
+        and status
+        """
+        course_runs = course_content.get('course_runs')
+        # A course can only be valid if it has course runs
+        if not course_runs:
+            return False
+
+        course_content_valid = False
+        for course_run in course_runs:
+            if self.is_course_run_content_valid(course_run):
+                course_content_valid = True
+        return course_content_valid
+
     def __str__(self):
         """
         Return human-readable string representation.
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -900,6 +900,33 @@ def update_contentmetadata_from_discovery(catalog_query):
...
+                        course_run.get('status') == "published"):
+                    course_valid = True
+
+            if not course_valid:
+                metadata.pop(index)
+                for key in content_item.get('course_run_keys'):
+                    invalid_course_runs.add(key)
+
+    if invalid_course_runs:
+        num_of_things_to_remove = len(invalid_course_runs)
```
alex-sheehan-edx: I wonder if I can be smarter here... maybe sort metadata...

enterprise_catalog/apps/catalog/models.py
```diff
@@ -900,6 +900,33 @@ def update_contentmetadata_from_discovery(catalog_query):
     # metadata will be an empty dict if unavailable from cache or API.
     metadata = CatalogQueryMetadata(catalog_query).metadata
 
+    invalid_course_runs = set()
+    for index, content_item in enumerate(metadata):
+        if get_content_type(content_item) == 'course':
+            course_runs = content_item.get('course_runs')
+            course_valid = False
+            for course_run in course_runs:
+                if ((course_run.get('availability') in ["Current", "Starting Soon", "Upcoming"]) and
```
johnnagro: lets make a validity function to call here

alex-sheehan-edx: I made some adjustments based on the team's thoughts. It's been pulled out to a method tied to the query

enterprise_catalog/apps/catalog/models.py
```diff
@@ -116,6 +116,38 @@ def get_by_uuid(cls, uuid):
...
             return None
 
+    def is_course_run_content_valid(self, course_run_content):
+        """
+        Check if a piece of course run content (JSON blob) adheres to a query's filter for availability and status
+        """
+        if self.content_filter.get('availability'):
+            if not course_run_content.get('availability') or \
+            course_run_content.get('availability') not in self.content_filter.get('availability'):
+                return False
```
iloveagent57, line 126: nit: might be easier to read as
```python
required_availability = self.content_filter.get('availability')
actual_availability = course_run_content.get('availability')
if required_availability and (actual_availability != required_availability):
    return False
```

iloveagent57 (commented):
I wonder if we could make this more organized in the following way:
1. Create a new file called `content_validators.py` or something that contains all our filter functions.  They'd generally take a single piece of JSON metadata as input and boolean output.
2. They should be as small as possible.  Even your one-line check in here insisting that there are course runs in a course should be a function.  Like `course_has_runs(content)`.
3. Declare a dict mapping content types to listS of validators to run the content through.  Something like `VALIDATORS_BY_CONTENT_TYPE`.
4. Then in `update_contentmetadata_from_discovery()` you can do something like: 
```python
validated_metadata = []
for record in metadata:
    validators = VALIDATORS_BY_CONTENT_TYPE[get_content_type(record)]
    if all([validator(record) for validator in validators]):
        validated_metadata.append(record)
```

This will have the effect of:
* making it obvious where new validation functions should live
* making it easy, at a glance, to tell how metadata straight from discovery might be filtered out before getting added to the enterprise-catalog models.
* make it easy to plug in new validators - just write your function and add it to that mapping, rather than trying to find the right place in the pipeline to insert any new business logic.

If we do something like this, we should do a search for other instances of content validation that should be explicitly represented in the same way.

johnnagro:
I really like the approach @iloveagent57 outlined in https://github.com/openedx/enterprise-catalog/pull/492#pullrequestreview-1092426801

alex-sheehan-edx:
This PR was made as part of a discovery investigation into filtering bad content. Now that we've reached a satisfying stopping point, I'm going to close this PR and let whoever takes the follow up ticket go off the work/comments here.

PR #507, opened 2022-09-27
fix: catalog_query is now unique on the content filter hash and inclu…

iloveagent57:
…de_exec_ed_2u_courses.

ENT-6360

## Description

We need this change to support cases where two catalogs (really CatalogQueries) want content from a common `content_filter`, except one wants to exclude exec ed content (by setting `include_exec_ed_2u_courses` to False) and the other wants to include such content (by setting the same field to True).

https://openedx.atlassian.net/browse/ENT-6360

## Post-review


docs/decisions/0004-catalog-query-exec-ed-field.rst
```diff
@@ -0,0 +1,41 @@
...
+------
+
+Accepted (circa September 2022)
+
+Context
+-------
+
+We currently do not allow course content with type ``executive-education-2u`` to be associated
+with any ``CatalogQuery`` record (and therefore, no such content is included in any Catalog).
+We'd like to selectively allow such content to be associated with queries/catalogs.
```
adamstankiewicz: It might be helpful to describe _why_ we don't allow course content with type `executive-education-2u` to be associated without opting in with the `CatalogQuery` flag.

docs/decisions/0004-catalog-query-exec-ed-field.rst
```diff
@@ -0,0 +1,41 @@
...
+
+Decision
+--------
+
+We created a new field, ``CatalogQuery.include_exec_ed_2u_courses``, which when ``True`` for
+a CatalogQuery record, indicates that course content with type ``executive-education-2u``
```
adamstankiewicz: nit: put backticks around `CatalogQuery` for consistency

adamstankiewicz (approved):
LGTM, really just 1 suggestion on the ADR to add slightly more context.

docs/decisions/0004-catalog-query-exec-ed-field.rst
```diff
@@ -0,0 +1,41 @@
...
+should not be excluded from being associated with that record.
```
johnnagro: double negative here, that can be confusing

adamstankiewicz: it is nice to be more explicit about the behavior, though, since the flag really isn't about inclusion per se (i.e., enabling this field doesn't mean you get Exec Ed courses on top of whatever your catalog query is). Rather, it is about exclusion based on the catalog query as written, as an additional filter.

Saying "... should include exec courses" is somewhat misleading as a result, depending on how you interpret "inclusion".

iloveagent57: yeah, I'm struggling to find a good way to both accurately describe the behavior and make word choices that are easy for a reader to grok.  I'll give it a few more minutes of thought.

PR #511, opened 2022-10-14
feat: allowing the default catalog view accept a course type facet

alex-sheehan-edx:
## Description

the default catalog view needs to be able to distinguish between edx course content and 2u course content. This PR allows for a course_type query param to the view that will filter the algolia search

## Ticket Link

Link to the associated ticket
[Prepare public catalog edx course content for exec ed being added to a la carte](https://2u-internal.atlassian.net/browse/ENT-6418)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -138,6 +138,21 @@ class EnterpriseCatalogDefaultCatalogResultsTests(APITestMixin):
             'You must provide at least one of the following query parameters: enterprise_catalog_query_titles.'
         ]
 
+    @mock.patch('enterprise_catalog.apps.api.v1.views.default_catalog_results.get_initialized_algolia_client')
+    def test_default_catalog_results_view_works_with_one_and_many_course_types(self, mock_algolia_client):
+        """
+        Test that the default catalog results view rejects requests where the query param course_type is not a list
+        """
+        mock_algolia_client.return_value.algolia_index.search.side_effect = [self.mock_algolia_hits, {'hits': []}]
+        url = self._get_contains_content_base_url()
+        facets = 'enterprise_catalog_query_titles=foo&course_type=course'
+        response = self.client.get(f'{url}?{facets}')
+        assert response.status_code == 200
+
+        facets = 'enterprise_catalog_query_titles=foo&course_type=course&course_type=notcourse'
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -138,6 +138,21 @@ def test_required_param_validation(self):
...
+        """
+        Test that the default catalog results view rejects requests where the query param course_type is not a list
+        """
+        mock_algolia_client.return_value.algolia_index.search.side_effect = [self.mock_algolia_hits, {'hits': []}]
+        url = self._get_contains_content_base_url()
+        facets = 'enterprise_catalog_query_titles=foo&course_type=course'
+        response = self.client.get(f'{url}?{facets}')
+        assert response.status_code == 200
+
+        facets = 'enterprise_catalog_query_titles=foo&course_type=course&course_type=notcourse'
```
marlonkeating, line 152: clarification: Would we expect to see urls with multiple course_type parameters?

enterprise_catalog/apps/api/v1/views/default_catalog_results.py
```diff
@@ -24,6 +25,7 @@ class DefaultCatalogResultsView(GenericAPIView):
     courses if a content type isn't provided.
     """
     permission_classes = []
+    queryset = CatalogQuery.objects.none()
```
marlonkeating: Clarification: Is queryset used anywhere?

enterprise_catalog/apps/api/v1/tests/test_views.py
alex-sheehan-edx: yes! The primary example would be edx course content has multiple types (verified, audit, etc). If we want to filter for just edx course content we would need to include multiple course content types to encompass all edx courses.

enterprise_catalog/apps/api/v1/views/default_catalog_results.py
alex-sheehan-edx: it's not- Django requires a specified query set for generic api views, however this view is just querying Algolia and not touching our models whatsoever. I don't love having an unused class attribute so I should maybe look into other work arounds. My thought is that query UUID is the only required param for the view so might as well specify catalog queries as the queryset (`queries` are referring to two different things here 🤦). Maybe there's a different base view to subclass from.

marlonkeating (approved):

PR #516, opened 2022-10-24
fix: add missing migration files

adamstankiewicz:
## Description

On `master`, running `./manage.py makemigrations` results in the following migration files that appear to be missing.

## Ticket Link

N/A

## Post-review


adamstankiewicz:
Closing this PR after seeing https://github.com/openedx/enterprise-catalog/pull/506/files.

We need to fix whatever is causing these migration(s) to show up when running `makemigrations`, otherwise we run the risk of getting these migrations merged in unintentionally.

PR #524, opened 2022-11-21
feat: deterministic content ordering and HighlightSet card_image_url

pwnage101:
This adds two closely-related features:
* Deterministic content ordering: Anywhere a list of highlighted content is serialized, the ordering of that list always mirrors the order in which they were added to that HighlightSet.
* HighlightSet `card_image_url`: Anywhere a HighlightSet is serialized or summarized, a `card_image_url` field is included and always borrowed from the first content added to that HighlightSet.

ENT-6495

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_curation_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_curation_views.py
@@ -1,13 +1,16 @@
 """
 Tests for curation-related views.
 """
+import itertools
 import uuid
 
 import ddt
+from faker import Faker
 from rest_framework import status
 from rest_framework.reverse import reverse
 
 from enterprise_catalog.apps.api.v1.tests.mixins import APITestMixin
+from enterprise_catalog.apps.catalog.constants import COURSE, PROGRAM
 from enterprise_catalog.apps.catalog.tests.factories import (
     ContentMetadataFactory,
 )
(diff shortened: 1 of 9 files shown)


enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -80,7 +82,9 @@ def get_permission_object(self):
             return str(self.requested_enterprise_uuid)
 
         try:
-            enterprise_curation_config = EnterpriseCurationConfig.objects.get(uuid=self.requested_enterprise_curation_config_uuid)
+            enterprise_curation_config = EnterpriseCurationConfig.objects.get(
+                uuid=self.requested_enterprise_curation_config_uuid
+            )
```
pwnage101, line 87: Many of the changes in this file are just quality/style fixes that pre-existed

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -287,13 +310,19 @@ def add_content(self, request, uuid, *args, **kwargs):
...
-        """ Remove existing content from an existing HighlightSet """
-        content_keys = set(utils.get_content_keys_from_request_data(request))
+        """
+        Remove existing content from an existing HighlightSet
+
+        Arguments:
+        - `uuid` (str): UUID of the HighlightSet to remove content from.
+        - `request.data["content_keys"]` (str): A list of content keys to remove.
+        """
+        content_keys = self.requested_content_keys
```
pwnage101, line 320: This line effectively changes content_keys from being a set to a (ordered) list, hence the following changes in this function.

enterprise_catalog/apps/api/v1/views/curation/utils.py
```diff
@@ -34,6 +34,16 @@ def get_enterprise_uuid_from_request_data(request):
...
+
+    Args:
+        - request (rest_framework.request.Request): The request object containing the request payload.
+
+    Returns:
+        list of str: content keys.
     """
-    content_keys = request.data.get('content_keys', [])
-    return content_keys
+    content_keys_raw = request.data.getlist('content_keys', [])
```
pwnage101: I have no idea why I had to do this now, but because `request.data` is a QueryDict, I wasn't able to get `request.data.get()` to actually fetch the entire list of content keys.  It was only giving me one single element instead of a list of all elements, so I had to switch to `request.data.getlist()`:

```
(Pdb) request.data
<QueryDict: {'content_keys': ['foo', '00001_metadata_item']}>
(Pdb) request.data.get('content_keys')
'00001_metadata_item'
(Pdb) request.data.getlist('content_keys')
['foo', '00001_metadata_item']
```

Why doesn't this bug appear in the currently-deployed code???  It must be something in this PR that is triggering this behavior...

enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -61,6 +63,10 @@ class ContentMetadataFactory(factory.django.DjangoModelFactory):
     """
     class Meta:
         model = ContentMetadata
+        exclude = ('card_image_url_prefix', 'card_image_url')
```
pwnage101, line 66: The `exclude` list allows me to declare certain input parameters to this factory without actually causing them to be passed-through to the ORM create.

enterprise_catalog/apps/api/v1/views/curation/utils.py
adamstankiewicz: @pwnage101 Hm, weird. I pulled down your branch, changed this line back to `request.data.get('content_keys', [])` and it works without error. When I leave it with `getlist`, I see `'dict' object has no attribute 'getlist'` 🤔

enterprise_catalog/apps/api/v1/tests/test_curation_views.py
```diff
@@ -297,3 +364,71 @@ def test_patch(self):
...
+            0 1 2 3 4
+            ^
+        After removing the second element, the selected URL still has an original index of 0:
+            0 2 3 4
+            ^
+        After removing the first element, the selected URL now has an original index of 2:
+            2 3 4
+            ^
+        After adding an element, the selected URL still has an original index of 2:
+            2 3 4 (new)
```
adamstankiewicz, line 382: Nice description of this test case 🙌

enterprise_catalog/apps/api/v1/views/curation/utils.py
pwnage101: Changes since last comment:
1. reverted `.getlist()` back to `.get()`.  It may be worth noting that this effectively drops support for requests containing `multipart/form-data`, but I don't think we care since we do everything in JSON.
2. added `'TEST_REQUEST_DEFAULT_FORMAT': 'json',`.
3. changed tests that previously POSTed `multipart/form-data` data to instead send `application/json` data.
4. removed all explicit formatting in test POSTs.

enterprise_catalog/apps/api/v1/tests/test_curation_views.py
macdiesel: +1 on this, excellent docs Troy.

enterprise_catalog/apps/curation/models.py
```diff
@@ -164,7 +195,7 @@ def card_image_url(self):
         * `COURSERUN` and `LEARNER_PATHWAY` content types are not supported, and result in `None`.
 
         Returns:
-            str: URL of the card image.
+            str: URL of the card image.  None if no card image is found.
```
macdiesel, line 198: 🙌

enterprise_catalog/apps/curation/tests/test_models.py
```diff
@@ -37,4 +36,5 @@ def test_authoring_organizations(self, content_type):
         authoring_organizations_under_test = highlighted_content.authoring_organizations
         self.assertEqual(authoring_organizations_under_test[0]['uuid'], str(FAKE_CONTENT_AUTHOR_UUID))
         self.assertEqual(authoring_organizations_under_test[0]['name'], FAKE_CONTENT_AUTHOR_NAME)
-        self.assertEqual(authoring_organizations_under_test[0]['logo_image_url'], FAKE_IMAGE_URL)
+        self.assertTrue(authoring_organizations_under_test[0]['logo_image_url'].startswith('https://'))
+        self.assertTrue(authoring_organizations_under_test[0]['logo_image_url'].endswith('.jpg'))
```
macdiesel, line 40: Curious about this, since it's just a test case I think the assert is fine. Do we allow other file formats than jpg? Not expecting a test case for them, more just curious. Do we support PNG?

pwnage101: I started using faker for this under the hood (instead of using a static module variable) which meant that the value is unpredictable now, except for the prefix/suffix.  The ".jpg" suffix is added in the factory logic, so this unit test does require some synchronization with the factory class, but I'm open to just chopping off that last assertion because the primary goal is to check that any URL is found (for which the "https://" check is sufficient)

adamstankiewicz (approved):
LGTM.

PR #525, opened 2022-11-23
feat: curation django admin functionality

pwnage101:
This is a bundle of changes all in pursuit of a complete django admin experience for curation models.

* Introduce DjangoQL into this project, and integrate with curation models to start.
* Make `title` for all curation objects searchable, as well as enable suggestions for them.
* Use a TextInput widget instead of ChoiceMenu for HighlightedContent.content_metadata to improve performance and UX.
* Make `enterprise_uuid` readonly for EnterpriseCurationConfig.
* Upgrade requirements.

ENT-5668

### DIFF
--- a/enterprise_catalog/apps/curation/admin.py
+++ b/enterprise_catalog/apps/curation/admin.py
@@ -1,46 +1,115 @@
+from django import forms
 from django.contrib import admin
+from djangoql.admin import DjangoQLSchema, DjangoQLSearchMixin
 from simple_history.admin import SimpleHistoryAdmin
 
 from enterprise_catalog.apps.curation import models
 
 
-class EnterpriseCurationConfigAdmin(SimpleHistoryAdmin):
-    """ Admin configuration for the EnterpriseCurationConfig model. """
+class EnterpriseCurationConfigQLSchema(DjangoQLSchema):
+    """
+    DjangoQL schema for EnterpriseCurationConfigAdmin, primarily for enabling suggestions for `title` fields.
+    """
+
+    suggest_options = {
+        models.EnterpriseCurationConfig: ['title'],
+    }
+
+
+class EnterpriseCurationConfigAdmin(DjangoQLSearchMixin, SimpleHistoryAdmin):
+    """
+    Admin configuration for the EnterpriseCurationConfig model.
+    """
+
+    djangoql_schema = EnterpriseCurationConfigQLSchema
+    # explicitly set the fields, only for the purpose of predictable ordering in the Admin UI.
+    fields = (
+        'title',
+        'enterprise_uuid',
+        'is_highlight_feature_active',
+    )
     list_display = (
         'uuid',
         'enterprise_uuid',
         'title',
         'is_highlight_feature_active',
     )
-    search_fields = (
-        'uuid',
-        'enterprise_uuid',
-    )
+
+    def get_readonly_fields(self, request, obj=None):
+        """
+        Force enterprise_uuid to readonly mode, if and only if the current request is to edit an existing object.
+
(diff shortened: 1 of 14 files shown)


enterprise_catalog/settings/base.py
```diff
@@ -54,7 +54,9 @@
     'corsheaders',
     'csrf.apps.CsrfAppConfig',  # Enables frontend apps to retrieve CSRF tokens
     'django_celery_results',  # Enables a Django model as the celery result backend
+    'djangoql',
     'rest_framework',
+    'simple_history',
```
pwnage101, line 59: This line fixed the template bug that appeared when using both DjangoQLSearchMixin and SimpleHistoryAdmin simultaneously.  Adding this line is the first step in the django-simple-history installation instructions, so I'm not going to comment about it in the code.

enterprise_catalog/settings/base.py
```diff
@@ -87,6 +89,10 @@
     'social_django.middleware.SocialAuthExceptionMiddleware',
     'crum.CurrentRequestUserMiddleware',
     'waffle.middleware.WaffleMiddleware',
+
+    # Lets simple history track which user made changes via API.  Manual changes via django admin already have the user
+    # tracked via SimpleHistoryAdmin.
+    'simple_history.middleware.HistoryRequestMiddleware',
```
pwnage101, line 95: I decided to just add this so that we can, for example, track exactly which enterprise customer user made a change to their curation objects.

macdiesel (commented):
Looks good, bump up the codecov and this should be all set.

enterprise_catalog/apps/curation/admin.py
```diff
@@ -1,50 +1,119 @@
...
+    list_display = (
         'uuid',
         'enterprise_uuid',
+        'title',
+        'is_highlight_feature_active',
     )
 
+    def get_readonly_fields(self, request, obj=None):
+        """
+        Force enterprise_uuid to readonly mode, iff the current request is to edit an existing object.
```
adamstankiewicz: nit: "iff" -> "if"

enterprise_catalog/apps/curation/admin.py
```diff
@@ -1,50 +1,119 @@
...
+
+        Rationale: I can't think of a valid reason to edit the enterprise_uuid.  Perhaps if you wanted to conveniently
```
adamstankiewicz, line 42: +1 to `enterprise_uuid` being read only for editing existing objects.

enterprise_catalog/apps/curation/admin.py
```diff
@@ -1,50 +1,119 @@
...
+    class Meta:
+        model = models.HighlightedContent
+        fields = ['catalog_highlight_set', 'content_metadata']
+        labels = {
+            'content_metadata': 'Content Metadata ID',
+        }
+        help_texts = {
+            'content_metadata': (
+                'ID of the Content Metadata object (different than the content key), which is discoverable via the '
+                '<a href="/admin/catalog/contentmetadata/">Content Metadata admin page</a>.'
```
adamstankiewicz, line 98: [kudos] Nice inclusion of a handy link!

adamstankiewicz (approved):
LGTM. Looks good running locally. I like the inclusion of the `suggest_options`!

enterprise_catalog/apps/curation/admin.py
pwnage101: Iff stands for "if and only if", meaning if the condition isn't met then the thing doesn't happen.  It's maybe obscure mathematics/logic vocabulary, but I've seen it used in edx code over the years.  Happy to be more verbose though.

adamstankiewicz: TIL haha. Wouldn't hurt to be a little more verbose, but I don't feel strongly :) Thanks for explaining!

PR #530, opened 2022-12-02
feat: Support for adding content while creating highlight set

pwnage101:
Also:
* [limits] Implement hard-coded limits to HighlightSet and HighlightedContent counts.
* [performance] Pull in RequestCache functionality from edx-django-util to cache the requested content_keys and avoid repeated calls to deduplication logic.
* [performance] In one code path, consolidate queries against ContentMetadata form N to 1 query.

ENT-6556

### DIFF
--- a/enterprise_catalog/apps/api/v1/views/curation/highlights.py
+++ b/enterprise_catalog/apps/api/v1/views/curation/highlights.py
@@ -1,18 +1,14 @@
 import logging
 from uuid import UUID
 
+from edx_django_utils.cache import RequestCache
 from edx_rbac.decorators import permission_required
 from edx_rbac.mixins import PermissionRequiredForListingMixin
-from rest_framework import viewsets
+from rest_framework import status, viewsets
 from rest_framework.decorators import action
 from rest_framework.exceptions import ParseError
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
-from rest_framework.status import (
-    HTTP_200_OK,
-    HTTP_201_CREATED,
-    HTTP_400_BAD_REQUEST,
-)
 from rest_framework_xml.renderers import XMLRenderer
 
 from enterprise_catalog.apps.api.v1.pagination import (
@@ -46,7 +42,18 @@ from enterprise_catalog.apps.curation.models import (
 )
 
 
+REQUEST_CACHE_NAMESPACE = 'CURATION_REQUEST_CACHE'
+CONTENT_PER_HIGHLIGHTSET_LIMIT = 12
+HIGHLIGHTSETS_PER_ENTERPRISE_LIMIT = 8
 logger = logging.getLogger(__name__)
+
+
+class LimitExceeded(Exception):
+    """
+    Use for any errors related to exceeding backend limits on object creation, such as maximum highlighted content per
+    highlight set.
+    """
+    pass
 
 
 class EnterpriseCurationConfigBaseViewSet(PermissionRequiredForListingMixin, BaseViewSet):
@@ -133,7 +140,7 @@ class EnterpriseCurationConfigViewSet(EnterpriseCurationConfigBaseViewSet, views
         if not self.requested_enterprise_uuid:
             return Response(
                 f'An enterprise UUID was not specified.',
-                status=HTTP_400_BAD_REQUEST
+                status=status.HTTP_400_BAD_REQUEST
             )
 
         try:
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -3,16 +3,11 @@
 
 from edx_rbac.decorators import permission_required
 from edx_rbac.mixins import PermissionRequiredForListingMixin
-from rest_framework import viewsets
+from rest_framework import status, viewsets
 from rest_framework.decorators import action
 from rest_framework.exceptions import ParseError
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
-from rest_framework.status import (
```
adamstankiewicz, line 11: nice QoL change :)

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -243,26 +238,82 @@ def _validate_existing_enterprise_curation_config(self):
...
+
     def create(self, request, *args, **kwargs):
         """
         Create a new HighlightSet
+
+        Arguments:
+        - `request.data["enterprise_customer"]` (str): UUID of enterprise customer for which to create a highlight set.
+        - `request.data["title"]` (str): Desired title of the highlight set.
+        - `request.data["is_published"]` (bool, optional): True if the highlight set should be published.
+        - `request.data["content_keys"]` (list of str, optional): A list of content keys to add.
```
macdiesel, line 348: Thank you again for the docs!

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -243,26 +238,82 @@ def _validate_existing_enterprise_curation_config(self):
...
+            set.
+        """
+        content_to_add = []
+        ignored_content_keys = []
+        added_content_keys = []
+        existing_content_keys = []
+
+        for content_key in self.requested_content_keys:
+            try:
+                content_metadata = ContentMetadata.objects.get(content_key=content_key)
```
macdiesel: Instead of running a query for each ContentMetadata item perhaps you could use [values()](https://docs.djangoproject.com/en/dev/ref/models/querysets/#values) or values_list() to fetch all the items in a single SQL call then do the filtering of the items in memory here? 

While no, I don't think this will have huge performance impacts for this particular use case, I do think it would be good habits for us to build and to avoid making many sequential database calls like this. This pattern can easily slip in to a n+1 type of issue down the road. Anytime we know the data set can be kept small, we can fetch it all in one SQL call and do the filtering in memory, we should do so.

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -243,26 +238,82 @@ def _validate_existing_enterprise_curation_config(self):
...
             return Response(
                 f'An EnterpriseCurationConfig must exist for enterprise UUID {self.requested_enterprise_uuid} '
                 'in order to create a HighlightSet',
-                status=HTTP_400_BAD_REQUEST
+                status=status.HTTP_400_BAD_REQUEST
             )
         request.data['enterprise_curation'] = str(curation_config.uuid)
 
-        return super().create(request, *args, **kwargs)
+        create_response = super().create(request, *args, **kwargs)
```
macdiesel: Nit: I found this variable name confusing reading the code.

macdiesel (commented):
A few nits and comments.

enterprise_catalog/apps/api/v1/views/curation/highlights.py
pwnage101: I have re-written this logic to consolidate it down to one query.

I also just overhauled this whole function also due to incorporating limit handling.

enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -237,32 +262,131 @@ def _validate_existing_enterprise_curation_config(self):
...
 
+    def _add_requested_content(self, highlight_set):
+        """
+        Helper function to add requested content to the given highlight set.
+
+        Arguments:
+            highlight_set (HighlightSet model instance): The highlight set to attempt to add content to.
+
+        Returns:
+            3-tuple of lists of str: Each tuple element is represents "added", "ignored", and "existing" content_keys.
```
adamstankiewicz: nit: "...element ~is~ represents..."?

macdiesel (approved):

adamstankiewicz (approved):

PR #531, opened 2022-12-02
feat: new learning_type algolia facet filter

alex-sheehan-edx:
## Description

Adding learning type to Algolia

## Post-review

### DIFF
--- a/docs/decisions/0005-creating-learning-type-facet.rst
+++ b/docs/decisions/0005-creating-learning-type-facet.rst
@@ -1,0 +1,35 @@
+Grouping and indexing of content and course type
+================================================
+
+Status
+------
+
+Pending (December 2022)
+
+Context
+-------
+
+There exists a disconnect between what product and ux want for enterprise content categorization, and what exists
+today in enterprise's source of truth (the Discovery Service). Currently, Executive Education is grouped with other edX
+course content under the same content type (course), while programs and pathways hold seperate values. This contradicts
+the mocks, ACs and visions for the enterprise product suite, where customers have access to all distinct learning types.
+We would like to maintain our current system of categorization while also allowing Executive Education content to be a
+faceted value of content in Algolia, without breaking any existing dependencies on the content_type facet.
+
+Decision
+--------
+
+We create a new indexed and faceted field in algolia, ``learning_type``, which will act very similar to Discovery's
+`content_type`. However `learning_type` will allow for the enterprise team to account for situations like this; where there
+in a sub-grouping of content that needs to be surfaced to users along side the parent `content_type` groupings.
+
+Consequences
+------------
+
+Further bloating Algolia data. Rule of developing universal standards (https://xkcd.com/927/).
+
+Alternatives considered
+-----------------------
+
+We considered instead of creating a new faceted field, to override the `content_type` Algolia field for Executive Education courses
+but chose to forego that option as to not run the risk of changing requirements/behaviors for consumers of the existing field.
--- a/enterprise_catalog/apps/api/v1/views/default_catalog_results.py
+++ b/enterprise_catalog/apps/api/v1/views/default_catalog_results.py
@@ -81,7 +81,7 @@ class DefaultCatalogResultsView(GenericAPIView):
 
         if content_type == 'course':
             if course_types := facets.get("course_type"):
-                catalog_filter.append([f'course_type:{type}' in course_types])
+                catalog_filter.append([f'course_type:{type}' for type in course_types])
 
         search_options = {
             'facetFilters': catalog_filter,
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -10,6 +10,7 @@ from enterprise_catalog.apps.api.v1.utils import is_course_run_active
 from enterprise_catalog.apps.api_client.algolia import AlgoliaSearchClient
 from enterprise_catalog.apps.catalog.constants import (
     COURSE,
+    EXEC_ED_2U_COURSE_TYPE,
     LEARNER_PATHWAY,
     PROGRAM,
     PROGRAM_TYPES_MAP,
@@ -72,6 +73,7 @@ ALGOLIA_FIELDS = [
     'course_type',
     'course_length',
     'entitlements',
+    'learning_type',
 ]
 
 # default configuration for the index
@@ -111,6 +113,7 @@ ALGOLIA_INDEX_SETTINGS = {
         'course_type',
         'course_length',
         'aggregation_key',
+        'learning_type',
     ],
     'unretrievableAttributes': [
         'enterprise_catalog_uuids',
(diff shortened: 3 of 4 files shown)


docs/decisions/0005-creating-learning-type-facet.rst
```diff
@@ -0,0 +1,35 @@
+Allowing inclusion of exec ed content in Catalog Queries
```
iloveagent57: nit: copypasta

docs/decisions/0005-creating-learning-type-facet.rst
```diff
@@ -0,0 +1,35 @@
...
+Pending (December 2022)
+
+Context
+-------
+
+There exists a disconnect between what product and ux want for enterprise content categorization, and what exists
+today in enterprise's source of truth (the Discovery Service). Currently, Executive Education is grouped with other edX 
+course content under the same content type (course), while programs and pathways hold seperate values. This contradicts 
+the mocks, ACs and visions for the enterprise product suite, where customers have access to all distinct learning types.  
+We would like to keep the existing classification and categorization and support the grouping of content across the
```
iloveagent57: nit, might rephrase as "We would like to maintain our current system of categorization while also allowing _Exec Ed_ content to be a faceted value of content in Algolia, without breaking any existing dependencies on the `content_type` facet."

docs/decisions/0005-creating-learning-type-facet.rst
```diff
@@ -0,0 +1,35 @@
...
+today in enterprise's source of truth (the Discovery Service). Currently, Executive Education is grouped with other edX 
+course content under the same content type (course), while programs and pathways hold seperate values. This contradicts 
+the mocks, ACs and visions for the enterprise product suite, where customers have access to all distinct learning types.  
+We would like to keep the existing classification and categorization and support the grouping of content across the 
+enterprise product. 
+
+Decision
+--------
+
+We create a new indexed and faceted field in algolia, ``learning_type``, which when will act very similar to Discovery's
```
iloveagent57: nit: extraneous "when" in "which when will act..."

docs/decisions/0005-creating-learning-type-facet.rst
```diff
@@ -0,0 +1,35 @@
...
+--------
+
+We create a new indexed and faceted field in algolia, ``learning_type``, which when will act very similar to Discovery's 
+`content_type`. However `learning_type` will allow for the enterprise team to account for situations like this; where there
+in a sub-grouping of content that needs to be surfaced to users along side the parent `content_type` groupings.   
+
+Consequences
+------------
+
+Further bloating Algolia data. Technically storing redundant data. Rule of developing universal standards (https://xkcd.com/927/).
```
iloveagent57: I'm not sure that "technically storing redundant data" is any different than "bloating Algolia data" - the only extra data we're storing in this scheme is in the new Algolia facet/field.

Also, part of the problem is that we _don't have_ any universal standard about this - that's why you had to write an ADR. :)

enterprise_catalog/apps/api/v1/views/default_catalog_results.py
```diff
@@ -81,7 +81,7 @@ def get(self, request, **kwargs):
 
         if content_type == 'course':
             if course_types := facets.get("course_type"):
-                catalog_filter.append([f'course_type:{type}' in course_types])
+                catalog_filter.append([f'course_type:{type}' for type in course_types])
```
iloveagent57, line 84: holy...what?!

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1051,6 +1054,50 @@ def get_course_first_paid_enrollable_seat_price(course):
...
+    Arguments:
+        course (dict): a dictionary representing a course
+
+    Returns:
+        str: the learning type (ADR: docs/decisions/0005-creating-learning-type-facet)
+        of the course.
+    """
+    if course.get('course_type') == EXEC_ED_2U_COURSE_TYPE:
+        return EXEC_ED_2U_COURSE_TYPE
+    return course.get('content_type')
```
iloveagent57: I think this function actually covers the other two functions.  Like if you had this:
```python
def get_learning_type(content):
    if content.get('course_type') == EXEC_ED_2U_COURSE_TYPE:
        return EXEC_ED_2U_COURSE_TYPE
    return content.get('content_type')
```
It would work for courses, programs, and pathways, because programs and pathways will have `content.get('course_type')` evaluate to null.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -4,45 +4,9 @@
...
-    get_program_prices,
-    get_program_skill_names,
-    get_program_subjects,
-    get_program_title,
-    get_program_type,
-    get_upcoming_course_runs,
-    is_course_archived,
-)
-from enterprise_catalog.apps.catalog.constants import COURSE, PROGRAM
+from enterprise_catalog.apps.catalog import algolia_utils as utils
```
iloveagent57, line 7: nice :)

iloveagent57 (commented):
looks good, a few comments/suggestions

docs/decisions/0005-creating-learning-type-facet.rst
```diff
@@ -0,0 +1,35 @@
...
+Consequences
+------------
+
+Further bloating Algolia data. Technically storing redundant data. Rule of developing universal standards (https://xkcd.com/927/). 
+
+Alternatives considered
+-----------------------
+
+We considered instead of creating a new faceted field, to override the `content_type` Algolia field for Executive Education courses
+but chose to forego that option as to not run the risk of changing requirements/behaviors for consumers of the existing field.
```
adamstankiewicz, line 35: [clarification] If we don't override the existing `content_type` Algolia field, will we have to update all the places we currently render the facet options for `content_type` to replace its data source with `learning_type` (e.g., learner portal, admin portal, explore catalog)? What is the risk of adding one more option to the existing `content_type` Algolia field more specifically?

docs/decisions/0005-creating-learning-type-facet.rst
alex-sheehan-edx: 🤦

enterprise_catalog/apps/api/v1/views/default_catalog_results.py
alex-sheehan-edx: yeah.... I actually talked a bit with @adamstankiewicz and we may be removing this endpoint all together but I wanted to fix the bug while it's still in use as this is currently affecting prod 🙃

enterprise_catalog/apps/catalog/algolia_utils.py
alex-sheehan-edx: I supposed you're right!

enterprise_catalog/apps/api/v1/views/default_catalog_results.py
adamstankiewicz: [clarification/context] Regarding "we may be removing this endpoint all together", I was just musing yesterday why enterprise-catalog makes GET-type requests to Algolia since all of its data already lives in the enterprise-catalog DB through `ContentMetadata`; making these GET network requests to Algolia in enterprise-catalog could possibly be avoided in the future (tech debt).

I suppose a caveat to this is whether enterprise-catalog relies on the custom search ranking provided by Algolia, in which case querying Algolia could still make sense.

johnnagro (approved):
i like adding a new `learning_type` instead of transforming the existing `content_type` - my opinion is that it gives us more options/flexibility without changing the semantics of the underlying data from discovery

PR #534, opened 2022-12-05
feat: adding backoff/retry and logging to DiscoveryAPIClient

johnnagro:
## Description

Update catalog job often generates a bunch of json decode errors which lead us to believe we are overloading the service a bit, leading to non-json error messages (and then parse exceptions). This change adds basic backoff/retry and extra logging to the prime complainer method. It should make the error more recoverable, less likely to happen, and more obvious when it happens.

Cribbed off [prior art in edx-enterprise](https://github.com/edx/edx-enterprise/blob/master/integrated_channels/degreed2/client.py#L436). 
 
## Ticket Link

- [ENT-6546](https://2u-internal.atlassian.net/browse/ENT-6546)

### DIFF
--- a/enterprise_catalog/apps/api_client/discovery.py
+++ b/enterprise_catalog/apps/api_client/discovery.py
@@ -24,12 +24,14 @@ class DiscoveryApiClient(BaseOAuthClient):
     Object builds an API client to make calls to the Discovery Service.
     """
 
+    # the maximum number of retries to attempt a call
     MAX_RETRIES = getattr(settings, "ENTERPRISE_DISCOVERY_CLIENT_MAX_RETRIES", 4)
+    # the number of seconds to sleep beteween tries, which is doubled every attempt 
     BACKOFF_FACTOR = getattr(settings, "ENTERPRISE_DISCOVERY_CLIENT_BACKOFF_FACTOR", 2)
 
     def _calculate_backoff(self, attempt_count):
         """
-        Calcualte the seconds to sleep based on attempt_count
+        Calculate the seconds to sleep based on attempt_count
         """
         return (self.BACKOFF_FACTOR * (2 ** (attempt_count - 1)))
 
--- a/enterprise_catalog/apps/api_client/discovery.py
+++ b/enterprise_catalog/apps/api_client/discovery.py
@@ -2,8 +2,10 @@
 Discovery service api client code.
 """
 import logging
+import time
 
 from celery.exceptions import SoftTimeLimitExceeded
+from django.conf import settings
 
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -22,18 +24,40 @@ class DiscoveryApiClient(BaseOAuthClient):
     Object builds an API client to make calls to the Discovery Service.
     """
 
+    MAX_RETRIES = getattr(settings, "ENTERPRISE_DISCOVERY_CLIENT_MAX_RETRIES", 4)
+    BACKOFF_FACTOR = getattr(settings, "ENTERPRISE_DISCOVERY_CLIENT_BACKOFF_FACTOR", 2)
+
+    def _calculate_backoff(self, attempt_count):
+        """
+        Calcualte the seconds to sleep based on attempt_count
```
marlonkeating: typo
```suggestion
        Calculate the seconds to sleep based on attempt_count
```

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -22,18 +24,40 @@ class DiscoveryApiClient(BaseOAuthClient):
     Object builds an API client to make calls to the Discovery Service.
     """
 
+    MAX_RETRIES = getattr(settings, "ENTERPRISE_DISCOVERY_CLIENT_MAX_RETRIES", 4)
+    BACKOFF_FACTOR = getattr(settings, "ENTERPRISE_DISCOVERY_CLIENT_BACKOFF_FACTOR", 2)
```
marlonkeating, line 30: We should document BACKOFF_FACTOR, as it's not automatically clear that it represents a number of seconds that will be doubled on every retry.

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -22,18 +24,40 @@ class DiscoveryApiClient(BaseOAuthClient):
...
         """
         LOGGER.info('Retrieving results from course-discovery for page %s...', page)
-        response = self.client.post(
-            DISCOVERY_SEARCH_ALL_ENDPOINT,
-            json=content_filter,
-            params=request_params,
-        ).json()
-        return response
+        attempts = 0
+        while True:
```
marlonkeating, line 45: We may be only using it in one place in this code base for right now, but given that backoff/retry is a pretty common operation, should we put it in a helper method (that can take functions as arguments for the bits of functionality that will differ between cases)?

marlonkeating (changes requested):

marlonkeating (approved):

PR #540, opened 2022-12-10
feat: add segment events for curation app

pwnage101:
```
commit 8de2b0da5533efcbb0ffaf082fbd7dee62723485
Author: Troy Sankey <tsankey@edx.org>
Date:   Mon Dec 12 12:44:08 2022 -0500

    chore: install analytics-python (segment) and make upgrade
    
    ENT-5672

commit cd076ba3c4ba52152b377fe29fd00376604e283f
Author: Troy Sankey <tsankey@edx.org>
Date:   Thu Dec 8 14:37:57 2022 -0500

    feat: add segment events for curation app
    
    Also:
    * Update highlight set create validator to prevent creation of empty
      highlight set if the requested content count exceeds the limit.
    * Limit exceeded tests should actually check the JSON response for an
      appropriate error description.
    * flesh out deletion API tests
```


enterprise_catalog/apps/track/apps.py
```diff
@@ -0,0 +1,27 @@
...
+
+import analytics
+from django.apps import AppConfig
+from django.conf import settings
+
+
+logger = logging.getLogger(__name__)
+
+
+class TrackConfig(AppConfig):
```
iloveagent57, line 14: Do we use this anywhere?  Does the `track` app need to be added to the `INSTALLED_APPS` list in the base Django settings?

iloveagent57 (approved):
This looks great, just one question about settings (maybe?).

enterprise_catalog/apps/track/apps.py
pwnage101: I think I was supposed to add it to `PROJECT_APPS`.  This "app" was an attempt to just create a space to add shared code (see `segment.py`); I didn't want to just stuff the generic tracking logic into one of the existing apps.

pwnage101: to be clear, this whole pattern is borrowed from enterprise-access.  e.g.: https://github.com/edx/enterprise-access/blob/b64875f64e2bf3166144acbcc826623bee066017/enterprise_access/settings/base.py#L59

iloveagent57: Ah, right, throw it into `PROJECT_APPS` - I think it has to be included in the project's `INSTALLED_APPS` for the segment key setting to be available at Django project start...or something like that.

PR #543, opened 2022-12-14
feat: adding exec ed to csv

kiram15:
## Description

Adding Executive Education information to the csv download on explore catalog. 

## Ticket Link

[Associated Ticket
](https://2u-internal.atlassian.net/browse/ENT-6421)

## Post-review


enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -61,28 +61,51 @@
...
+    'What You’ll Learn',
+    'Full Description',
+]
+
 ALGOLIA_ATTRIBUTES_TO_RETRIEVE = [
-    'title',
-    'key',
-    'content_type',
-    'partners',
     'advertised_course_run',
```
kiram15, line 85: It was hard to look through all of these so I just alphabetized them!

johnnagro (approved):

enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -119,6 +142,7 @@ def course_hit_to_row(hit):
     """
     Helper function to construct a CSV row according to a single Algolia result course hit.
     """
+    exec_course = hit.get('course_type') == 'executive-education-2u' if True else False
```
marlonkeating: Would make the variable name more clear that it's a boolean, and also remove the extraneous logic at the end.
```suggestion
    is_exec_course = hit.get('course_type') == 'executive-education-2u'
```

marlonkeating (changes requested):
question: For course_hit_to_row(), given the number of exceptions for executive courses, might the logic be easier to follow if we structured the function like
`if is_exec_course:
    # Parse everything for executive course
  else:
    # Parse everything for non-executive course
  
  `

enterprise_catalog/apps/api/v1/export_utils.py
kiram15: Yeah this is what we were talking about, basically its the question of do we want a lot of copied code to be cleaner and more readable, or do we want to reduce copied code. I agreed with you, but John and Alex were in the nay boat. Tbh its just preference 
https://twou.slack.com/archives/C049WNMFFRP/p1669755717967509

marlonkeating (approved):

enterprise_catalog/apps/api/v1/export_utils.py
alex-sheehan-edx: I have no objections to change the variable name!

alex-sheehan-edx (approved):
lgtm!

PR #544, opened 2022-12-14
feat: HighlightSet.card_image_url supports all content types

pwnage101:
Support for card_image_url on all content types, as well as unit tests and docs for all other properties.

Epic: ENT-5666

### DIFF
--- a/enterprise_catalog/apps/curation/models.py
+++ b/enterprise_catalog/apps/curation/models.py
@@ -5,7 +5,12 @@ from django.utils.translation import gettext as _
 from model_utils.models import TimeStampedModel
 from simple_history.models import HistoricalRecords
 
-from enterprise_catalog.apps.catalog.constants import COURSE, PROGRAM
+from enterprise_catalog.apps.catalog.constants import (
+    COURSE,
+    COURSE_RUN,
+    LEARNER_PATHWAY,
+    PROGRAM,
+)
 from enterprise_catalog.apps.catalog.models import ContentMetadata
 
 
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/curation/models.py
```diff
@@ -197,22 +200,29 @@ def card_image_url(self):
...
         if content_type == COURSE:
             return self.content_metadata.json_metadata.get('image_url')  # pylint: disable=no-member
+        if content_type == COURSE_RUN:
+            return self.content_metadata.json_metadata.get('image_url')  # pylint: disable=no-member
         elif content_type == PROGRAM:
             return self.content_metadata.json_metadata.get('card_image_url')  # pylint: disable=no-member
+        elif content_type == LEARNER_PATHWAY:
+            try:
+                # pylint: disable=invalid-sequence-index
+                return self.content_metadata.json_metadata['card_image']['card']['url']
```
adamstankiewicz, line 221: [confirmation] If a pathway has no image associated with it, how is that represented in the `json_metadata`? Is the `KeyError` exception handling here specifically catching that case?

adamstankiewicz (approved):
LGTM, just one clarifying/confirmation question.

enterprise_catalog/apps/curation/models.py
pwnage101: Catching `KeyError` is just defensive programming, just like using `get()` instead of `[]`.  Truth is I don't know what the json metadata would look like otherwise because I looked at all the learner pathways on prod and they all had a URL at that path.

adamstankiewicz: @pwnage101 On the `LearnerPathway` model in course-discovery, the `card_image` field is optional. If not provided, `card_image` looks to be `null`.

Looks like if you run `self.content_metadata.json_metadata['card_image']['card']['url']` when `"card_image": null`, it'll throw a `TypeError`:

image

Do you think this is worth handling?

pwnage101: Oh, you're right I forgot about that case.

adamstankiewicz (approved):

PR #545, opened 2022-12-15
fix: csv price and exec ed fix

kiram15:
## Description

Quick fix to the exec ed csv output, and correction of a minor bug that included cents on the price (want to remain consistent with course prices) 

## Post-review


alex-sheehan-edx (approved):
do you think a test is necessary here?

PR #548, opened 2022-12-22
fix: add aggregation_key to HighlightedContentSerializer

adamstankiewicz:
## Description

Adds `aggregation_key` to `HighlightedContentSerializer`.

## Ticket Link

N/A

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/serializers.py
+++ b/enterprise_catalog/apps/api/v1/serializers.py
@@ -301,7 +301,7 @@ class HighlightedContentSerializer(serializers.ModelSerializer):
         """
         Returns the aggregation key for the associated ContentMetadata.
         """
-        return obj.content_metadata.aggregation_key
+        return obj.aggregation_key
 
 
 class HighlightSetSerializer(serializers.ModelSerializer):
--- a/enterprise_catalog/apps/api/v1/serializers.py
+++ b/enterprise_catalog/apps/api/v1/serializers.py
@@ -327,7 +327,7 @@ class HighlightSetSerializer(serializers.ModelSerializer):
         """
         Returns the data for the associated content included in this HighlightSet object.
         """
-        qs = obj.highlighted_content.order_by('created')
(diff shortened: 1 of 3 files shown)


iloveagent57 (commented):
Could you do me a favor and add a `select_related('content_metadata')` on this line https://github.com/openedx/enterprise-catalog/blob/53a5a57d04078217009a24e76c4baf775acedfec/enterprise_catalog/apps/api/v1/serializers.py#L321 so that we don't end up with one SELECT per highlighted content item?

adamstankiewicz:
@iloveagent57 Good idea, added.

enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -283,17 +283,26 @@ class HighlightedContentSerializer(serializers.ModelSerializer):
...
             'title',
             'card_image_url',
             'authoring_organizations',
         ]
 
+    def get_aggregation_key(self, obj):
+        """
+        Returns the aggregation key for the associated ContentMetadata.
+        """
+        return obj.content_metadata.aggregation_key
```
iloveagent57: I think you meant to do `obj.aggregation_key` here, since you added that nice property on the `HighlightedContent` model.

iloveagent57 (commented):
One more thing I missed!

enterprise_catalog/apps/api/v1/serializers.py
adamstankiewicz: D'oh, yep! Fixed.

iloveagent57 (approved):
Looks good, you might need a test update or two to appease the codecov gods.

adamstankiewicz:
> Looks good, you might need a test update or two to appease the codecov gods.

@iloveagent57 The only thing Codecov is complaining about now is this conditional:

image

Considering the `codecov/project` is above the 92% threshold and only the `codecov/patch` is failing, I'm going to go ahead and merge.

PR #552, opened 2023-01-05
feat: adding registration date to exec ed csv download

alex-sheehan-edx:
## Description

adding registration deadline to the exec ed csv download endpoint

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/export_utils.py
+++ b/enterprise_catalog/apps/api/v1/export_utils.py
@@ -74,6 +74,7 @@ CSV_EXEC_ED_COURSE_HEADERS = [
     'Skills',
     'Min Effort',
     'Max Effort',
+    'Registration Deadline',
     'Length',
     'What You’ll Learn',
     'Full Description',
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -228,6 +229,7 @@ def exec_ed_course_to_row(hit):
             end_date = parser.parse(hit['additional_metadata']['end_date']).strftime(DATE_FORMAT)
         csv_row.append(end_date)
         key = hit.get('advertised_course_run', {}).get('key')
+        enroll_by_date = hit.get('additional_metadata', {}).get('registration_deadline')
```
marlonkeating: question: Is there a reason we use hit.get('key') vs. hit['key']?

alex-sheehan-edx: no real reason, I like using the getter when the existence isn't guaranteed but it is in this case so I can switch it!

enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -227,11 +228,13 @@ def exec_ed_course_to_row(hit):
         if hit['additional_metadata'].get('end_date'):
             end_date = parser.parse(hit['additional_metadata']['end_date']).strftime(DATE_FORMAT)
         csv_row.append(end_date)
-        key = hit.get('advertised_course_run', {}).get('key')
+        enroll_by_date = hit['additional_metadata'].get('registration_deadline')
```
marlonkeating: nit: Given how much hit['additional_metadata'] is being used, might it make sense to just break it out into a variable at the top of the if block?

marlonkeating (approved):

PR #555, opened 2023-01-11
feat: added boolean field for catalog visibility

brobro10000:
Ticket: https://2u-internal.atlassian.net/browse/ENT-6579

Added boolean field to enterprise curation model `can_only_view_highlight_sets` as a flag for the learners to display either highlights only, or all content in their catalogs

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review


enterprise_catalog/apps/catalog/migrations/0035_alter_contentmetadata_associated_content_metadata.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.16 on 2023-01-11 21:32
```
adamstankiewicz: This migration and `enterprise_catalog/apps/core/migrations/0003_alter_user_first_name.py` below are (unfortunately) incorrectly/unintentionally generated. AFAIK we haven't identified why, though. That said, when merged, they have caused performance issues in the past (see https://github.com/openedx/enterprise-catalog/pull/506).

That said, they can removed :)

Before removing the files, though, I might recommend unapplying those migrations, if already applied:

```python
 # make app-shell

# 1. show list of all migrations and whether they are applied or not.
./manage.py showmigrations

# 2. replace <migration name> with the previous migration you want to revert back to.
./manage.py migrate <migration name>
```

enterprise_catalog/apps/curation/admin.py
```diff
@@ -27,12 +27,14 @@ class EnterpriseCurationConfigAdmin(DjangoQLSearchMixin, SimpleHistoryAdmin):
         'title',
         'enterprise_uuid',
         'is_highlight_feature_active',
+        'view_only_highlight_sets',
```
adamstankiewicz: totally the tiniest nit haha, but technically this property could be interpreted as "view only" highlight sets (as in, read-only?). Perhaps `can_only_view_highlight_sets` or similar (something with a verb that implies a boolean)?

enterprise_catalog/apps/curation/migrations/0002_auto_20230111_2132.py
```diff
@@ -0,0 +1,23 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('curation', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='enterprisecurationconfig',
+            name='view_only_highlight_sets',
```
adamstankiewicz: [tiniest nit continued from above] If you do end up renaming this, you'll want to unapply and reapply the migrations with the same method described above.

enterprise_catalog/apps/curation/migrations/0002_auto_20230111_2132.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.2.16 on 2023-01-11 21:32
```
adamstankiewicz: nit: it's generally a good practice to name the migration file in some way, e.g., `0002_add_can_only_view_highlight_sets` or similar.

adamstankiewicz (commented):
Looks great! A couple really small nits, and we'll want to remove those automatically generated migrations unrelated to your changes.

Do we need to modify the tests to include any checks for that new field?

adamstankiewicz (approved):

enterprise_catalog/apps/curation/migrations/0002_auto_20230111_2132.py
brobro10000: Leaving open for learning purposes

enterprise_catalog/apps/catalog/migrations/0035_alter_contentmetadata_associated_content_metadata.py
brobro10000: Leaving open for future reference

PR #566, opened 2023-01-25
bug: properly validating request params for contains content items

alex-sheehan-edx:
## Description

Validating customer uuid in the `https://enterprise-catalog.edx.org/api/v1/enterprise-customer/<CUSTOMER UUID>/contains_content_items/` endpoint

## Ticket Link

https://opsg.in/a/i/2u-internal/64283c70-eb01-407d-b5ba-c33989aad0e8-1674529020487

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views/enterprise_customer.py
+++ b/enterprise_catalog/apps/api/v1/views/enterprise_customer.py
@@ -1,10 +1,12 @@
 import logging
+import uuid
 
 from django.utils.decorators import method_decorator
 from edx_rbac.utils import get_decoded_jwt
 from rest_framework.decorators import action
 from rest_framework.exceptions import PermissionDenied
 from rest_framework.response import Response
+from rest_framework.status import HTTP_400_BAD_REQUEST
 
 from enterprise_catalog.apps.api.v1.decorators import (
     require_at_least_one_query_parameter,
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -84,7 +85,18 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
-        customer_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
+        try:
+            customer_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
+        except ValueError as exc:
+            logger.warning(
+                f"Could not parse catalogs from provided enterprise uuid: {enterprise_uuid}. "
+                f"Query failed with exception: {exc}"
+            )
+            return Response(
+                f'Error: invalid enterprice customer uuid: {enterprise_uuid} provided.',
```
johnnagro: maybe put quotes inside the string around `{enterprise_uuid}` to capture/make-obvious whitespace, etc.

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -84,7 +85,18 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
+                status=HTTP_400_BAD_REQUEST
```
johnnagro, line 98: what are the callers getting now for a response? a 200 with an empty list? will throwing a 400 start causing weird upstream issues? (do we need to let anyone know, or offer some inbetween to ease the transition)

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -84,7 +85,18 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
         get_catalog_list = request.GET.get('get_catalog_list', False)
         course_run_ids = unquote_course_keys(course_run_ids)
 
-        customer_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
+        try:
+            customer_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
```
johnnagro: perhaps we should validate that its a properly formed UUID before we try to use it on the DB?

johnnagro (approved):
looks good - i have some suggestions to perhaps make it better

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
alex-sheehan-edx: the callers are getting a `500` AFAIK this is breaking on both prod and stage, so in my opinion any change here is a good change.

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
alex-sheehan-edx: sure I think that's fine but doesn't this accomplish essential the same thing?

PR #576, opened 2023-02-27
fix: preserving request authentication by passing drf request object 

alex-sheehan-edx:
## Description

the catalog crud view auth validation helper method was using crum to retrieve the request object which lacked the necessary attrs for the `edx_rest_framework_extensions` library to verify the request's authentication. I've opted instead to pass the request object directly to the helper method from the django view.

## Ticket Link

Link to the associated ticket
https://2u-internal.atlassian.net/browse/ENT-3995
## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views/curation/highlights.py
+++ b/enterprise_catalog/apps/api/v1/views/curation/highlights.py
@@ -35,10 +35,6 @@ from enterprise_catalog.apps.catalog.models import (
     ContentMetadata,
     EnterpriseCatalogRoleAssignment,
 )
-from enterprise_catalog.apps.catalog.rules import (
-    enterprises_with_admin_access,
-    has_access_to_all_enterprises,
-)
 from enterprise_catalog.apps.curation.models import (
     EnterpriseCurationConfig,
     HighlightedContent,
--- a/enterprise_catalog/apps/catalog/rules.py
+++ b/enterprise_catalog/apps/catalog/rules.py
@@ -109,16 +109,19 @@ def has_access_to_all_enterprises(enterprise_ids):
     return ACCESS_TO_ALL_ENTERPRISES_TOKEN in enterprise_ids
 
 
-def enterprises_with_admin_access(user):
+def enterprises_with_admin_access(request):
     """
     Returns a set of enterprise ids to which a user has been granted admin access.
 
     Note that this may include the "*" wildcard identifier, which means that the user
     is allowed access to all enterprises.
     """
-    if user.is_superuser:
+    if hasattr(request, 'user') and request.user.is_superuser:
         return {ACCESS_TO_ALL_ENTERPRISES_TOKEN}
-    return set(_enterprises_with_jwt_admin_access() + _enterprises_with_database_admin_access(user))
+    eligible_enterprises_jwt_admin_access = _enterprises_with_jwt_admin_access(request)
+    if hasattr(request, 'user'):
+        eligible_enterprises_jwt_admin_access += _enterprises_with_database_admin_access(request.user)
+    return set(eligible_enterprises_jwt_admin_access)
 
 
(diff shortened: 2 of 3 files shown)


enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -35,10 +35,10 @@
     ContentMetadata,
     EnterpriseCatalogRoleAssignment,
 )
-from enterprise_catalog.apps.catalog.rules import (
-    enterprises_with_admin_access,
-    has_access_to_all_enterprises,
-)
+# from enterprise_catalog.apps.catalog.rules import (
```
adamstankiewicz: nit: remove the uncommented lines if no longer necessary

enterprise_catalog/apps/catalog/rules.py
```diff
@@ -109,16 +109,19 @@ def has_access_to_all_enterprises(enterprise_ids):
...
     Returns a set of enterprise ids to which a user has been granted admin access.
 
     Note that this may include the "*" wildcard identifier, which means that the user
     is allowed access to all enterprises.
     """
-    if user.is_superuser:
+    if hasattr(request, 'user') and request.user.is_superuser:
         return {ACCESS_TO_ALL_ENTERPRISES_TOKEN}
-    return set(_enterprises_with_jwt_admin_access() + _enterprises_with_database_admin_access(user))
+    enterprises_w_admin_access = set(_enterprises_with_jwt_admin_access(request))
```
adamstankiewicz: tiny nit: `eligible_enterprises_jwt_admin_access` or similar (not too keen on the `_w_` 😄)

enterprise_catalog/apps/api/v1/views/curation/highlights.py
alex-sheehan-edx: I was checking if we used the methods being changed elsewhere in the repo. This is the only other occurrence I could find and it's an empty import 🙃 will remove entirely before merging the PR

enterprise_catalog/apps/catalog/rules.py
```diff
@@ -134,10 +137,12 @@ def _enterprises_with_database_admin_access(user):
...
 
-def _enterprises_with_jwt_admin_access():
+def _enterprises_with_jwt_admin_access(request=None):
     """
     Returns a list of enterprise ids to which a user has been granted admin access via JWT roles.
     """
-    request = crum.get_current_request()
+    if not request:
+        # Note crum returns a WSGIResquest objsect and __not__ a django rest framework request object.
+        request = crum.get_current_request()
```
adamstankiewicz: [curious] Is `_enterprises_with_jwt_admin_access` called/used anywhere else? Do we need to support both `request` types here or could we cutover to using the `request` in the kwargs?

enterprise_catalog/apps/catalog/rules.py
alex-sheehan-edx: lol fair, fair

enterprise_catalog/apps/api/v1/views/curation/highlights.py
adamstankiewicz: Addressed via https://github.com/openedx/enterprise-catalog/pull/576#discussion_r1119078002

enterprise_catalog/apps/catalog/rules.py
alex-sheehan-edx: I believe this is the only usage. I was just trying to tread with extreme care 😄

alex-sheehan-edx: yeah it's 1000% not used anywhere else

adamstankiewicz (approved):
LGTM. Though, are there any tests we should be adding/updating to account for these changes?

PR #578, opened 2023-03-02
fix: blocking emeritus product source

kiram15:
## Description

Adding in the allowed values for the new `product_source` field, purposefully excluding Emeritus courses. 

## Ticket Link

Link to the associated ticket
[ENT-6785](https://2u-internal.atlassian.net/browse/ENT-6785)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -721,6 +721,14 @@ def _create_new_content_metadata(nonexisting_metadata_defaults):
     return metadata_list
 
 
+def _fetch_product_source(metadata_entry):
+    product_source = metadata_entry.get('product_source')
+    if isinstance(product_source, dict):
+        return product_source.get('slug')
+    else:
+        return product_source
+
+
 def _should_allow_metadata(metadata_entry, catalog_query=None):
     """
     Determines if an object from Discovery meets our criteria for indexing
@@ -731,11 +739,11 @@ def _should_allow_metadata(metadata_entry, catalog_query=None):
     Returns:
         bool: If we should save the metadata as a ContentMetaData object
     """
-    # make sure to exclude exec ed course runs
-    content_type = get_content_type(metadata_entry)
-    entry_product_source = metadata_entry.get('product_source')
+    entry_product_source = _fetch_product_source(metadata_entry)
     if entry_product_source not in CONTENT_PRODUCT_SOURCE_ALLOW_LIST and entry_product_source is not None:
         return False
+    # make sure to exclude exec ed course runs
+    content_type = get_content_type(metadata_entry)
     if not catalog_query or not catalog_query.include_exec_ed_2u_courses:
         if content_type == 'courserun':
             if seat_types := metadata_entry.get('seat_types'):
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -732,13 +733,15 @@ def _should_allow_metadata(metadata_entry, catalog_query=None):
     """
     # make sure to exclude exec ed course runs
     content_type = get_content_type(metadata_entry)
+    entry_product_source = metadata_entry.get('product_source')
```
bcitro: I think we'll want to write a helper function (something like `get_product_source(metadata_entry)`) to extract the product source from the metadata, since it may be in different formats depending on the discovery endpoint used. The `search/all` endpoint will be a string (e.g. `"product_source": "2u"`), but `courses` endpoint will be a dictionary, e.g.:

`"product_source": {
        "name": "2u",
        "slug": "2u",
        "description": "2U, Trilogy, Getsmarter -- external source for 2u courses and programs"
    }`

enterprise_catalog/apps/catalog/tests/test_models.py
```diff
@@ -85,6 +85,46 @@ def test_2U_exec_ed_content_inclusion_logic(self, mock_client):
...
+            ('aggregation_key', 'course:edX+edxSourceX'),
+            ('key', 'edX+edxSourceX'),
+            ('title', 'test course'),
+            ('product_source', 'edX'),
+        ])
+        twou_course_metadata = OrderedDict([
+            ('aggregation_key', 'course:edX+2uSourceX'),
+            ('key', 'edX+2uSourceX'),
+            ('title', 'test course 2'),
+            ('product_source', OrderedDict(
```
johnnagro, line 105: great that you have both the string and dict style metadata in your test 👍

enterprise_catalog/apps/catalog/models.py
```diff
@@ -730,6 +739,9 @@ def _should_allow_metadata(metadata_entry, catalog_query=None):
     Returns:
         bool: If we should save the metadata as a ContentMetaData object
     """
+    entry_product_source = _fetch_product_source(metadata_entry)
+    if entry_product_source not in CONTENT_PRODUCT_SOURCE_ALLOW_LIST and entry_product_source is not None:
```
johnnagro, line 743: we may find that we need this to be case insensitive in the future.

johnnagro (approved):
⛵

enterprise_catalog/apps/catalog/tests/test_models.py
```diff
@@ -85,6 +85,46 @@ def test_2U_exec_ed_content_inclusion_logic(self, mock_client):
         update_contentmetadata_from_discovery(catalog.catalog_query)
         self.assertEqual(ContentMetadata.objects.count(), 3)
 
+    @override_settings(DISCOVERY_CATALOG_QUERY_CACHE_TIMEOUT=0)
+    @mock.patch('enterprise_catalog.apps.api_client.discovery_cache.DiscoveryApiClient')
+    def test_product_source_content_inclusion_logic(self, mock_client):
```
bcitro, line 90: can we add a test case for null product_source?

enterprise_catalog/apps/catalog/models.py
alex-sheehan-edx, line 743: tiny nit- I believe it's possible to shorten this to `if entry_product_source and entry_product_source not in CONTENT_PRODUCT_SOURCE_ALLOW_LIST:`

alex-sheehan-edx (approved):
tiny nit but not requiring a change 👍

PR #580, opened 2023-03-06
feat: added content metadata get endpoints

alex-sheehan-edx:
DRAFT

## Description

New get and list endpoints for content metadata items

## Post-review


enterprise_catalog/apps/catalog/migrations/0035_auto_20230306_1517.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 3.2.17 on 2023-03-06 15:17
```
adamstankiewicz: nit: helpful to rename migration files to something human-readable / more discoverable

enterprise_catalog/apps/catalog/migrations/0036_auto_20230306_1550.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 3.2.17 on 2023-03-06 15:50
```
adamstankiewicz: nit: helpful to rename migration files to something human-readable / more discoverable

enterprise_catalog/apps/catalog/models.py
```diff
@@ -476,6 +477,18 @@ class ContentMetadata(TimeStampedModel):
...
 
+    content_uuid = models.UUIDField(
+        null=True,
+        blank=True,
+        max_length=32,
+        unique=False,
+        verbose_name='Content UUID',
+        help_text=_(
+            "The UUID that represents a piece of content. This value is usually a secondary identifier to content_key "
+            "the enterprise environment."
```
adamstankiewicz: nit: "... _in_ the enterprise environment"?

enterprise_catalog/apps/catalog/utils.py
```diff
@@ -26,6 +26,13 @@ def get_content_filter_hash(content_filter):
     return content_filter_hash
 
 
+def get_content_uuid(metadata):
+    """
+    Returns the content uuid for a piece of metadata. Returns None for course runs.
```
adamstankiewicz, line 31: [curious/suggestion] Given that not all content types have a `uuid` field in course-discovery / the `ContentMetadata`'s `json_metadata` fields, I wonder if we could use `aggregation_key` which is a standard format for all content types (e.g., `courserun:State-Bank-of-India+SBICB002x`, `course:DelftX+MMFEL1X`, `program:b17cc627-16b5-4e6c-bba1-112a7aed93ae`, `learnerpathway:e383d0aa-5869-4f57-a440-44c0ce8dbb99`)?

This tangentially came up during the implementation of the "Highlights" feature, where we ended up needing to rely on `aggregation_key` (and exposed it in Algolia) because it is the only identifier common to all content types in enterprise-catalog's `ContentMetadata`.

enterprise_catalog/apps/api/v1/views/content_metadata.py
```diff
@@ -0,0 +1,79 @@
...
+from enterprise_catalog.apps.api.v1.views.base import BaseViewSet
+from enterprise_catalog.apps.catalog.models import ContentMetadata
+from enterprise_catalog.apps.catalog.rules import enterprises_with_admin_access
+
+
+class ContentMetadataCRUDViewSet(BaseViewSet, viewsets.ModelViewSet):
+    """ Viewset for CRUD operations on Enterprise Catalogs """
+    serializer_class = ContentMetadataSerializer
+    renderer_classes = [JSONRenderer, XMLRenderer]
+    permission_required = 'catalog.has_admin_access'
```
adamstankiewicz: [consideration] I believe there are use cases where enterprise learners could be hitting at least the `GET` endpoint for `ContentMetadataCRUDViewSet`. Currently, the enterprise learner portal's course/program pages get their metadata about the content from discovery (e.g., `/api/v1/courses`). If we are collectively deciding to use enterprise-catalog as the source of truth for serving up enterprise related content metadata, we may migrate to using this new `GET` endpoint in the course page, where users only have the `enterprise_learner` role.

In the past, we have taken an approach of defining a [base viewset](https://github.com/openedx/enterprise-catalog/blob/f90850063fda4eb903d045813f92b6a24ee5d730/enterprise_catalog/apps/api/v1/views/curation/highlights.py#L59) that encapsulates the majority of the logic and then [individual learner-specific](https://github.com/openedx/enterprise-catalog/blob/f90850063fda4eb903d045813f92b6a24ee5d730/enterprise_catalog/apps/api/v1/views/curation/highlights.py#L117) and [admin-specific viewsets](https://github.com/openedx/enterprise-catalog/blob/f90850063fda4eb903d045813f92b6a24ee5d730/enterprise_catalog/apps/api/v1/views/curation/highlights.py#L127) where only the permissions and available endpoints might change (i.e., learner-specific viewset is read-only, just exposing the `GET` endpoint).

johnnagro (approved):
⛵  - we may wanna tweak the serializer, the permissions, etc.

enterprise_catalog/apps/api/v1/views/content_metadata.py
alex-sheehan-edx: @johnnagro and I were just chatting offline about the permissions needed for this endpoint! Do you think we even really need the enterprise_learner role instead of just simply "authed"?

alex-sheehan-edx: what if we stripped the roles required entirely?

enterprise_catalog/apps/catalog/migrations/0035_auto_20230306_1517.py
alex-sheehan-edx: Something I definitely don't do often enough 😄

enterprise_catalog/apps/catalog/models.py
alex-sheehan-edx: oops.. something something... multiline python strings.. something something

enterprise_catalog/apps/catalog/migrations/0036_auto_20230306_1550.py
```diff
@@ -0,0 +1,22 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('catalog', '0035_auto_20230306_1517'),
+    ]
+
+    operations = [
+        migrations.RunPython(backfill_content_uuids, reverse_code=migrations.RunPython.noop),
+    ]
```
iloveagent57, line 22: Just double-checking: it's fine to add this operation (as a discrete migration 👏 ) in the same PR/deployment, and then a subsequent deployment will set `content_uuid` to be a unique field - correct?

alex-sheehan-edx: there shouldn't be any reason for this to be rejected, right?

alex-sheehan-edx: meaning there's no restraints on the column so django has nothing to be angry about

alex-sheehan-edx: it's also relevant that these records both don't get a default uuid, nor are they manually created. We pull the uuid out at the time of content ingestion from discovery.

enterprise_catalog/apps/api/v1/views/content_metadata.py
adamstankiewicz: Following up from Slack discussions/meetings, we've decided to account for `enterprise_learner` role, by moving the new API endpoint as an extension to the existing `/enterprise-customer/<enterprise_uuid>/` API viewset/route.

docs/decisions/0006-pivoting-to-content-uuids.rst
```diff
@@ -0,0 +1,47 @@
...
+
+
+Decision
+--------
+The Enterprise approach to identifying content will shift towards-
+
+1) Aggregation key will become deprecated; it’s a field computed by course-discovery for search purposes, but it’s not a
+persisted field on any content metadata model/table.
+
+2) Content UUID is added to the Enterprise Catalog service's Content Metadata table as a queriable field and is
```
adamstankiewicz: nit: "queryable"

docs/decisions/0006-pivoting-to-content-uuids.rst
```diff
@@ -0,0 +1,47 @@
...
+extracted from Course Discovery's response payload at time of content ingestion in the Catalog Service.
+
+3) Content UUID is the recommended primary content identifier for content types across the Enterprise product.
+
+4) The Enterprise Catalog api serves as a central point of conversion across content identification types used in
+the enterprise product.
```
iloveagent57: I'd add something to (3) along the lines of "but we'll continue to support ``content_key`` (from the ``key`` field of content records in course-discovery) as an equally-valid primary identifier of content, since every type of content record has a ``key``."

And for (4), it's a central point of conversion between ``key`` and ``uuid`` in the long term, particularly through this new endpoint you're adding.  And for a while, that same endpoint will also allow for the discovery of an aggregation key by either ``key`` or ``uuid``.  The endpoint will _not_ support querying by ``aggregation_key``.

docs/decisions/0006-pivoting-to-content-uuids.rst
```diff
@@ -0,0 +1,47 @@
...
+
+
+Consequences
+------------
+
+Content key was previously considered to be Enterprise content's primary identifier. This pivot reverses past
+decisions and means that work to implement `content_key` as a primary identifier will be halted. Importantly though,
+this does not require a change to previous implementations, and provides an easily accessable conversion tool from
+older identifiers to UUID.
```
iloveagent57: I'm not sure this is really true.  It's kind of like ``content_uuid`` and ``content_key`` are now a 1a, 1b situation.  ``content_uuid`` is preferred, but ``content_key`` is supported where it's already in use, and furthermore, we think it's valid for certain record types to reference _both_, if needed/helpful (and not too risky).

docs/decisions/0006-pivoting-to-content-uuids.rst
```diff
@@ -0,0 +1,47 @@
...
+
+Alternatives considered
+-----------------------
+
+We considered pivoting to other existing content identifiers like aggregation key, as well as keeping the existing
+pattern of using `content_key`. However, the universality of UUID across all content types, as well as the
+uniformity of the UUID format in said content types means UUID is the most appropriate value for the job.
```
iloveagent57: I might change this a little to again declare that the uuid is _preferred_ for the reasons you describe, but the key is still supported and will be going forward (basically because course-discovery births the content records with two primary keys, so we'll support both.  They're both useful for different reasons, we like UUID a little bit better a little more of the time).

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -140,3 +146,67 @@ def filter_content_items(self, request, enterprise_uuid, **kwargs):
...
+        return ContentMetadataSerializer(content_metadata_queryset, many=many)
+
+    @action(detail=True, methods=['get', 'list'])
+    def content_metadata(self, customer_uuid, content_identifier=None, **kwargs):  # pylint: disable=unused-argument
+        """
+        Get/List endpoint for `/api/v1/enterprise-customer/{customer uuid}/content-metadata/{content identifier}`.
+        Accepts both content uuids and content keys for the specific content metadata record requested
+        """
+        serializer = self.get_metadata_serializer()
+        return Response(serializer.data)
```
iloveagent57: We're already using a viewset - would it be simpler to break this apart into `list()` and `retrieve()` methods?  That might simplify that serializer logic you're doing above.

docs/decisions/0006-pivoting-to-content-uuids.rst
```diff
@@ -0,0 +1,47 @@
+Pivoting to Content Uuids as preferred primary identifiers
+==========================================================
+
+Status
+------
+
+Draft
```
adamstankiewicz: nit: accepted at this point? OK to keep as draft is we expect to keep adding to / editing this ADR

iloveagent57 (commented):
a few comments for now, I think you're still writing code as I'm reading it, though :)

enterprise_catalog/apps/catalog/utils.py
adamstankiewicz: Following up from discussions offline, we are deprecating aggregation_key in favor of content UUID and content key. This endpoint will only support content uuid/key. Comment resolved.

docs/decisions/0006-pivoting-to-content-uuids.rst
adamstankiewicz: > since every type of content record has a `key`."

@iloveagent57 Is this true for programs (i.e., `key` isn't included in course-discovery's `/api/v1/programs/` API response anyways)? Same for pathways. We use the `uuid` value for programs/pathways in the `content_key` model field on `ContentMetadata` objects.

adamstankiewicz (approved):

docs/decisions/0006-pivoting-to-content-uuids.rst
iloveagent57: Oh wow, yeah, good point.  So maybe what @alex-sheehan-edx had initially is the right thing to say (i.e. uuid is the recommended identifier), with the caveat of like "we support content_key because it's an equally-valid unique identifier of courses and course runs."

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -140,3 +145,41 @@ def filter_content_items(self, request, enterprise_uuid, **kwargs):
         }
 
         return Response(response_data)
+
+    def get_metadata_item_serializer(self, **kwargs):
```
iloveagent57, line 149: nit: does this need `**kwargs`?  You're relying on `self.kwargs` in the method body.

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -140,3 +145,41 @@ def filter_content_items(self, request, enterprise_uuid, **kwargs):
...
+        Fetches the queryset of content metadata under a specific customer context for a request
+        """
+        # Grab the enterprise from the url params and fetch the catalogs belonging to the customer
+        enterprise_uuid = self.kwargs.get('enterprise_uuid')
+        enterprise_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
+
+        # Set the context IFF there exists only one customer catalog
+        serializer_context = {}
+        if len(enterprise_catalogs) == 1:
+            serializer_context['enterprise_catalog'] = enterprise_catalogs.first()
```
iloveagent57, line 160: You're not adding anything more to this query, so maybe list() it first to make it easier to reason about? i.e.
```python
enterprise_catalogs = list(EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid))
# Set the context IFF there exists only one customer catalog
serializer_context = {}
if len(enterprise_catalogs) == 1:
    serializer_context['enterprise_catalog'] = enterprise_catalogs[0]
```

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -140,3 +145,41 @@ def filter_content_items(self, request, enterprise_uuid, **kwargs):
...
+
+        # Collect a queryset of all content under all of the customer's catalogs
+        content_metadata_queryset = EnterpriseCatalog.objects.none()
+        for catalog in enterprise_catalogs:
+            content_metadata_queryset = content_metadata_queryset | catalog.content_metadata
```
iloveagent57, line 165: It's confusing to me that we're building a thing called `content_metadata_queryset` but initializing it with an `EnterpriseCatalog` queryset....but maybe this is the cleanest way to select through this m2m relationship?  Have you taken a look at the SQL queries that this spits out?

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -2136,3 +2137,106 @@ def test_catalogs_different_uuids(self, use_different_query):
         else:
             assert response['num_distinct_query_ids'] == 1
         assert str(self.catalog_query_one.id) in response['catalog_uuids_by_catalog_query_id']
+
+
+@ddt.ddt
+class EnterpriseCustomerContentMetadataViewSetTests(APITestMixin):
+    """
+    Tests for the Enterprise Customer Content Metadata related endpoints.
+    """
```
iloveagent57, line 2146: nit: maybe we want a test for the case where one content item lives in multiple catalogs for the same customer?

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
iloveagent57: could you perhaps do something like
```python
content_metadata_queryset = ContentMetadata.prefetch_related(
    'catalog_queries',
    'catalog_queries__enterprise_catalog',
).filter(
    'catalog_queries__enterprise_catalog__uuid__in': [catalog.uuid for catalog in enterprise_catalogs],
)
```
I'm not sure that's any easier to read, but it at least makes it a little clearer about the type of data we're trying to find?

iloveagent57: Or use https://docs.djangoproject.com/en/3.2/ref/models/querysets/#django.db.models.Prefetch to cut down the number of queries:
```python
if not content_identifier := self.kwargs.get('content_identifier'):
    # early return here

try:
    uuid.UUID(content_identifier)
    content_filter = {'content_uuid': content_identifier}
except ValueError:
    content_filter = {'content_key': content_identifier}
catalog_query_queryset = CatalogQuery.objects.select_related('enterprise_catalogs').filter(
    'enterprise_catalog__uuid__in': [catalog.uuid for catalog in enterprise_catalogs],
)
content_metadata_queryset = ContentMetadata.objects.filter(**content_filter).prefetch_related(
    Prefetch('catalog_queries', queryset=catalog_query_queryset)
)
```

Sorry for the long comment string...I'm actually gonna leave it as-is so you can glean my thought process.

PR #584, opened 2023-03-15
fix: make product_source check case insensitive

bcitro:
## Description

Making the product_source check to include/exclude courses case-insensitive, as the data in discovery appears to vary between mixed case and lowercase.

## Ticket Link

https://2u-internal.atlassian.net/browse/ENT-6918

## Post-review


johnnagro (approved):
⛵

PR #588, opened 2023-03-30
fix: filter for content metadata contined in any customer's catalog better

iloveagent57:
# Handle cases where we search for course runs in a customer, but the customer's catalog(s) contain only `course` metadata proper.

## Post-review


johnnagro (approved):
⛵

pwnage101 (approved):
Code LGTM.  If I'm understanding correctly, this causes no behavior change when a content_uuid is passed to get_metadata_item_serializer, but when a content_key is passed, search results are potentially expanded to include parent content?

PR #597, opened 2023-05-03
fix: Added proxy login with executive education course enrollment url

zamanafzal:
## Description

 Added proxy login with executive education course enrollment URL

## Ticket Link

Link to the associated ticket
[ENT-7071](https://2u-internal.atlassian.net/browse/ENT-7071)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -40,6 +40,7 @@ from enterprise_catalog.apps.catalog.constants import (
 )
 from enterprise_catalog.apps.catalog.utils import (
     batch,
+    enterprise_proxy_login_url,
     get_content_filter_hash,
     get_content_key,
     get_content_type,
@@ -403,6 +404,12 @@ class EnterpriseCatalog(TimeStampedModel):
                 LOGGER.warning(warning, content_metadata.content_key, self.uuid)
                 return None
             params['sku'] = entitlement_sku
+            exec_ed_enrollment_url = update_query_parameters(url, params)
+            enterprise_proxy_exec_ed_enrollment_url = enterprise_proxy_login_url(
+                self.enterprise_customer.slug,
+                next_url=exec_ed_enrollment_url
+            )
+            return enterprise_proxy_exec_ed_enrollment_url
         elif self._can_enroll_via_learner_portal(content_key):
             course_key = content_key
             if parent_content_key:
--- a/enterprise_catalog/apps/catalog/utils.py
+++ b/enterprise_catalog/apps/catalog/utils.py
@@ -5,7 +5,9 @@ import hashlib
 import json
 from datetime import datetime
 from logging import getLogger
+from urllib.parse import urljoin
 
+from django.conf import settings
 from edx_rbac.utils import feature_roles_from_jwt
 from edx_rest_framework_extensions.auth.jwt.authentication import (
     get_decoded_jwt_from_auth,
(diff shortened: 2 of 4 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -434,8 +435,9 @@ def _get_exec_ed_2u_enrollment_url(self, content_metadata):
         for entitlement in content_metadata.json_metadata.get('entitlements', []):
             if entitlement['mode'] == EXEC_ED_2U_ENTITLEMENT_MODE:
                 entitlement_sku = entitlement.get('sku')
+        url = enterprise_proxy_login_url(self.enterprise_customer.slug)
         return (
-            '{}/executive-education-2u/checkout'.format(settings.ECOMMERCE_BASE_URL),
+            '{}&next={}/executive-education-2u/checkout'.format(url, settings.ECOMMERCE_BASE_URL),
             entitlement_sku,
```
adamstankiewicz: [suggestion] It might be worthwhile to add a `next` kwarg to the `enterprise_proxy_login_url` method so that it can be responsible for generating the full URL, including the `next` query parameter (and any other additional query parameters in the future). Perhaps the function signature looks something like:

```python
def enterprise_proxy_login_url(enterprise_slug, next_url=None):
    # ...
```

```python
next_url = f"{settings.ECOMMERCE_BASE_URL}/executive-education-2u/checkout"
url = enterprise_proxy_login_url(self.enterprise_customer.slug, next_url=next_url)
```

enterprise_catalog/apps/catalog/utils.py
```diff
@@ -105,3 +106,12 @@ def batch(iterable, batch_size=1):
 def localized_utcnow():
     """Helper function to return localized utcnow()."""
     return UTC.localize(datetime.utcnow())  # pylint: disable=no-value-for-parameter
+
+
+def enterprise_proxy_login_url(slug, executive_ed_enrollment_url):
```
adamstankiewicz: @zamanafzal nit: I think the method argument should be generic here as `next_url`, not Exec Ed specific, given in theory it could also be used for non-exec ed use cases too.

enterprise_catalog/apps/catalog/models.py
zamanafzal: Thank you for the suggestion. I've moved the next_url inside the function.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -403,6 +404,12 @@ def get_content_enrollment_url(self, content_metadata):
                 LOGGER.warning(warning, content_metadata.content_key, self.uuid)
                 return None
             params['sku'] = entitlement_sku
+            exec_ed_enrollment_url = update_query_parameters(url, params)
```
zamanafzal, line 407: `update_query_parameters` should be called first now as we have query params already in the URL and due to this it adds `&` except `?` in the `next` enrollment URL which will not fail the URL.

enterprise_catalog/apps/catalog/utils.py
```diff
@@ -105,3 +106,10 @@ def batch(iterable, batch_size=1):
 def localized_utcnow():
     """Helper function to return localized utcnow()."""
     return UTC.localize(datetime.utcnow())  # pylint: disable=no-value-for-parameter
+
+
+def enterprise_proxy_login_url(slug, next_url=None):
+    url = f"{settings.LMS_BASE_URL}/enterprise/proxy-login/?enterprise_slug={slug}"
+    if next_url:
+        url += f"&next={next_url}"
```
irfanuddinahmad: use urljoin

adamstankiewicz (approved):

irfanuddinahmad (approved):
LGTM!

PR #604, opened 2023-05-23
feat: Add new catalog type column

katrinan029:
## Description

**Problem**: The explore catalog tool displays on the screen displays very well which courses are included in the catalog categories: A la carte, business subscription, and education. However, the CSV download file does not present that information.

**Solution**: Add a column with the header "Associated Catalogs" in the tabs - Programs, Courses, and Course Runs to include the catalog categories. 

Screenshot 2023-05-23 at 3 32 51 PM

## Ticket Link

Link to the associated ticket
[Catalog Explore Download results include Associated catalogs](https://2u-internal.atlassian.net/browse/ENT-6963)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/export_utils.py
+++ b/enterprise_catalog/apps/api/v1/export_utils.py
@@ -34,6 +34,7 @@ CSV_COURSE_HEADERS = [
     'Length',
     'What You’ll Learn',
     'Pre-requisites',
+    'Associated Catalogs',
 ]
 
 CSV_PROGRAM_HEADERS = [
@@ -42,6 +43,7 @@ CSV_PROGRAM_HEADERS = [
     'Partner',
     'Short Description',
     'Number of courses',
+    'Associated Catalogs',
 ]
 
 CSV_COURSE_RUN_HEADERS = [
@@ -61,6 +63,7 @@ CSV_COURSE_RUN_HEADERS = [
     'Skills',
     'Subjects',
     'Language',
+    'Associated Catalogs'
 ]
 
 CSV_EXEC_ED_COURSE_HEADERS = [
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -124,6 +128,19 @@ def write_headers_to_sheet(worksheet, headers, cell_format):
         worksheet.write(0, col_num, cell_data, cell_format)
 
 
+def fetch_catalog_types(hit):
+    """
+    Helper function to extract only the three needed catalog types.
+    """
+    CATALOG_TYPES = [
```
kiram15, line 135: Did you talk with Emilio about him only wanting these catalogs?

alex-sheehan-edx: There are a lot of junk catalogs within the system. FWIKT the idea is that if this is a CSV download on the public catalog page,  and the public catalog only deals with the three default catalogs. But I also agree that Emilio should probably sign off before this is merged

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -620,6 +620,7 @@ class EnterpriseCatalogCsvDataViewTests(APITestMixin):
         'language': 'English',
         'level_type': 'Intermediate',
         'content_type': 'course',
+        'enterprise_catalog_query_titles': ['A la carte', 'Business', 'DemoX'],
```
alex-sheehan-edx, line 623: Love the inclusion of a catalog not included in the column! 👍

alex-sheehan-edx (approved):
As long as Emilio is a 👍 then I'm a 👍! :shipit:

alex-sheehan-edx (approved):
As long as Emilio is a 👍 then I'm a 👍! :shipit:

enterprise_catalog/apps/api/v1/export_utils.py
katrinan029: Emilio confirmed that the download should just include the three default catalog types.

PR #614, opened 2023-06-07
fix: minimize error rate

katrinan029:
## Description

**Problem**: At the endpoint of https://explore-catalog.edx.org/ the page automatically gets redirected to
[https://explore-catalog.edx.org/?enterprise_catalog_query_titles=A la carte&availability=Available Now&availability=Starting Soon&availability=Upcoming](https://explore-catalog.edx.org/?enterprise_catalog_query_titles=A%20la%20carte&availability=Available%20Now&availability=Starting%20Soon&availability=Upcoming) 
by default. On-call reported seeing the error https://onenr.io/0qwLXOdKzw5 that is caused by the URL https://explore-catalog.edx.org/ not having query parameters.

**Solution**: On the client, there are existing checks in place to verify that the query parameters are loaded before making the API request. The fix is implemented on the API by removing the method decorator that validates the query parameters because that same validation is already done on the client.

_Note:_ This error could not be reproduced locally. The changes will be closely monitored after being deployed. 

## Ticket Link

Link to the associated ticket
[[enterprise-catalog] You must provide at least one of the following query parameters: enterprise_catalog_query_titles](https://2u-internal.atlassian.net/browse/ENT-7003)

## Post-review

* [x] Ensure that once the changes have been [deployed to stage](https://gocd.tools.edx.org/go/pipeline/activity/stage-enterprise_catalog), prod is manually deployed

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_decorators.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_decorators.py
@@ -1,0 +1,29 @@
+from django.test import TestCase
+from django.http import HttpResponse
+from django.test import RequestFactory
+
+from rest_framework.exceptions import ValidationError
+from enterprise_catalog.apps.api.v1.decorators import require_at_least_one_query_parameter
+
+
+class DecoratorTests(TestCase):
+    """
+    Tests for the existence of at least one of the specified query in the decorator
+    """
+
+    def test_require_at_least_one_query_parameter(self):
+        """
+        Tests that at least one of the specified query parameters are included in the request
+        """
+        query_parameter_names = 'enterprise_catalog_query_titles'
+
+        @require_at_least_one_query_parameter(query_parameter_names)
+        def my_view(request):
+            return HttpResponse('a response for request {}'.format(request))
+        request = RequestFactory().get('/')
+        request.query_params = request.GET
+
+        with self.assertRaisesMessage(ValidationError,
+                                      'You must provide at least one of the following '
+                                      'query parameters: enterprise_catalog_query_titles.'):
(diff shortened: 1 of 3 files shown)


kiram15 (approved):
LGTM! 👍

Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM ubuntu:bionic as app
-LABEL maintainer="devops@edx.org"
+LABEL maintainer="sre@edx.org"
```
alex-sheehan-edx: Could I get some context on the docker file change?

kiram15: https://github.com/openedx/enterprise-catalog/pull/615 But also a bit confused why its showing up in this PR when it's already merged in

johnnagro: yeah i dunno how to "git properly" my bad. and the email address is the one used in more recent cookiecutter projects so i think its the correct one.

johnnagro: the dockerfile was no longer working. we needed at least `python3-pip`. i chose to also reformat stuff (to make it easier to see when we add new deps) as well as update the base ubuntu image.

johnnagro (approved):

Dockerfile
alex-sheehan-edx: Thanks @johnnagro ! Looks like your commit was included in this PR, we should probably exclude it before merging?

alex-sheehan-edx: pretty sure we can just exclude it with an interactive rebase?

katrinan029: Thanks for the suggestion, Alex. I rebased so this change won't show up in this PR

PR #621, opened 2023-06-21
feat: reindex-algolia constant memory consumption

pwnage101:
Before this change, the `index_enterprise_catalog_in_algolia_task` task (a.k.a. reindex-algolia) had a maximum memory consumption propotional to the total number of content_keys and number of related objects such as enterprises, catalogs, catalog queries, etc.  After this change, the maximum memory consumption will no longer be sensitive to the number of content keys (but still sensitive to the number of related objects).  This should significantly reduce the memory footprint of the task.

Estimate: In prod recently, I have observed 95,117 algolia products being indexed in total.  The default algolia library batch size is 1,000 products, so this should result in a ~95x factor decrease in memory consumption.

This is achieved by leveraging the fact that Algolia's `replace_all_objects` library method accepts any iterable, including generators, and does not accumulate more than 1 batch in-memory before shipping it over the network.  Before this change, we were accumulating all results into a list before even calling `replace_all_objects`; after this change, we instead construct a lazy generator and pass that along.

ENT-7311

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -514,13 +514,30 @@ def get_pathways_by_associated_content():
 
 def add_metadata_to_algolia_objects(
     metadata,
-    content_key,
-    catalog_uuids_by_key,
     algolia_products_by_object_id,
-    customer_uuids_by_key,
-    catalog_queries_by_key,
+    catalog_uuids,
+    customer_uuids,
+    catalog_queries,
 ):
-    """ Helper method to take individual content objects and add them to a combined `algolia products` object"""
+    """
+    Convert ContentMetadata objects into Algolia products and accumulate results into `algolia_products_by_object_id`.
+
+    At minimum, there are 3 duplicate Algolia products generated per course, one for each of [catalog uuids, customer
+    uuids, catalog queries], and possibly more if any one of those exceeds ALGOLIA_UUID_BATCH_SIZE.  In the case of the
+    batch size being exceeded, create further duplicate algolia product records, batching the uuids to reduce the
+    payload size of the Algolia product objects.
+
+    Args:
+        metadata (ContentMetadata): The course, program or learner pathway for which to generate aloglia products.
+        algolia_products_by_object_id (dict):
+            Object to append the resulting algolia products to.  Keys are objectIDs, and values are algolia products to
+            actually index.
+        catalog_uuids (list of str): Associated catalog UUIDs.
+        customer_uuids (list of str): Associated customer UUIDs.
+        catalog_queries (list of tuple(str, str)): Associated catalog queries, as a list of (UUID, title) tuples.
+    """
+
+    content_key = metadata.content_key
     # add enterprise-related uuids to json_metadata
     json_metadata = copy.deepcopy(metadata.json_metadata)
     json_metadata.update({
@@ -528,7 +545,7 @@ def add_metadata_to_algolia_objects(
     })
 
     # enterprise catalog uuids
-    catalog_uuids = sorted(list(catalog_uuids_by_key[content_key]))
+    catalog_uuids = sorted(list(catalog_uuids))
     batched_metadata = _batched_metadata(
         json_metadata,
         catalog_uuids,
@@ -538,7 +555,7 @@ def add_metadata_to_algolia_objects(
     _add_in_algolia_products_by_object_id(algolia_products_by_object_id, batched_metadata)
 
     # enterprise customer uuids
-    customer_uuids = sorted(list(customer_uuids_by_key[content_key]))
+    customer_uuids = sorted(list(customer_uuids))
     batched_metadata = _batched_metadata(
         json_metadata,
         customer_uuids,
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -555,11 +555,147 @@ def add_metadata_to_algolia_objects(
     _add_in_algolia_products_by_object_id(algolia_products_by_object_id, batched_metadata)
 
 
-def index_content_keys_in_algolia(content_keys, algolia_client):  # pylint: disable=too-many-statements
+# pylint: disable=too-many-statements
+def _get_algolia_products_for_batch(
```
pwnage101, line 575: The vast majority of this function is just copy+paste existing code.  Only changes are minor variable name changes.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -692,9 +736,9 @@ def index_content_keys_in_algolia(content_keys, algolia_client):  # pylint: disa
         f'[ENTERPRISE_CATALOG_ALGOLIA_REINDEX] {len(algolia_products_by_object_id.keys())} products found.'
     )
 
-    # extract out only the fields we care about and send to Algolia index
-    algolia_objects = create_algolia_objects(algolia_products_by_object_id.values(), ALGOLIA_FIELDS)
-    algolia_client.replace_all_objects(algolia_objects)
+    # Feed the un-evaluated flat iterable of algolia products into the 3rd party library function.  As of this writing,
+    # this library function will chunk the interable again using a default batch size of 100.
+    algolia_client.replace_all_objects(algolia_products_generator)
```
pwnage101: This is basically where the magic happens.  Instead of passing a huge 16GB+ python object into this function, we pass a 0GB generator.  `replace_all_objects` happens to already be written in a way that it will not load the entire list into memory.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -555,147 +555,188 @@ def add_metadata_to_algolia_objects(
...
-        # content_key or parent_content_key or course associated programs. returns courses, programs and course runs.
-        query = (
-            Q(content_key__in=content_keys_batch)
-            | Q(parent_content_key__in=content_keys_batch)
-            | Q(
-                associated_content_metadata__content_key__in=content_keys_batch,
-                content_type__in=[PROGRAM, LEARNER_PATHWAY]
-            )
+    # retrieve ContentMetadata records that match the specified content_keys in the
+    # content_key or parent_content_key or course associated programs. returns courses, programs and course runs.
```
adamstankiewicz: nit: this also includes learner pathways as well now (line 587 below). wondering if this comment should include pathways?

pwnage101:
I think I just realized there's no unit test that exercises the content_key batching, which wouldn't effectively test the generator logic beyond one batch.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -555,147 +555,188 @@ def add_metadata_to_algolia_objects(
...
-            | Q(content_type=LEARNER_PATHWAY)
-        )
-        # iterate over courses, programs and pathways and add their metadata to the list of objects to be indexed
-        for metadata in filtered_content_metadata:
+    # iterate through ContentMetadata records, retrieving the enterprise_catalog_uuids
+    # and enterprise_customer_uuids associated with each ContentMetadata record (either
+    # a course or a course run or program), storing them in a dictionary with the related
+    # content_key as a key for later retrieval. the content_key is determined by
+    # the content_key field if the metadata is a `COURSE` or `PROGRAM` or by the parent_content_key
+    # field if the metadata is a `COURSE_RUN`.
```
adamstankiewicz: nit: unrelated to your changes, also retrieves the associated `enterprise_catalog_queries` as well now

enterprise_catalog/apps/api/tasks.py
```diff
@@ -555,147 +555,188 @@ def add_metadata_to_algolia_objects(
...
+                customer_uuids_by_key[pathway.content_key].update(enterprise_customer_uuids)
+
+        if metadata.content_type == PROGRAM:
+            for pathway in program_to_pathway_mapping[content_key]:
+                catalog_queries_by_key[pathway.content_key].update(enterprise_catalog_queries)
+                catalog_uuids_by_key[pathway.content_key].update(enterprise_catalog_uuids)
+                customer_uuids_by_key[pathway.content_key].update(enterprise_customer_uuids)
+
+    # iterate through the courses and programs, retrieving the enterprise-related uuids from the
+    # dictionary created above. there is at least 2 duplicate course records per course,
```
adamstankiewicz: nit: "at least **3** duplicate course records per course", since it has the `catalog_queries_by_key` (i.e., catalog query uuids), too, now, though unrelated to your immediate changes.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -555,147 +555,188 @@ def add_metadata_to_algolia_objects(
...
+    filtered_content_metadata = content_metadata.filter(
+        Q(content_type=COURSE)
+        | Q(content_type=PROGRAM)
+        | Q(content_type=LEARNER_PATHWAY)
+    )
+    # iterate over courses, programs and pathways and add their metadata to the list of objects to be indexed
+    for metadata in filtered_content_metadata:
+        content_key = metadata.content_key
+        # Check if we've indexed the course recently
+        # (programs/pathways are indexed every time regardless of last indexing)
```
adamstankiewicz, line 670: [unrelated to your change, general musing] I wonder why this is? 🤔

enterprise_catalog/apps/api/tasks.py
```diff
@@ -555,147 +555,188 @@ def add_metadata_to_algolia_objects(
...
-        f'[ENTERPRISE_CATALOG_ALGOLIA_REINDEX] max number of uuids held in memory during reindexing: {max_uuid_count}'
+        f'[ENTERPRISE_CATALOG_ALGOLIA_REINDEX] There are {len(content_keys)} total content keys to include in the'
+        f' Algolia index.'
     )
+    course_to_pathway_mapping, program_to_pathway_mapping = get_pathways_by_associated_content()
+    metrics_accumulator = {
+        'total_products_count': 0,
+    }
+    # Produce a generator of batches of algolia products to index.  Each batch has an unpredictable, variable length.
+    # Not immediately evaluated, so no memory is consumed yet.
```
adamstankiewicz, line 717: 🎉

enterprise_catalog/apps/api/tasks.py
```diff
@@ -555,147 +555,188 @@ def add_metadata_to_algolia_objects(
...
+    # Still not evaluated, so no memory is consumed yet.
+    algolia_products_generator = (
+        algolia_product
+        for batch in algolia_products_batch_generator
+        for algolia_product in batch
+    )
+
+    # Feed the un-evaluated flat iterable of algolia products into the 3rd party library function.  As of this writing,
+    # this library function will chunk the interable again using a default batch size of 1000.
+    algolia_client.replace_all_objects(algolia_products_generator)
```
adamstankiewicz, line 743: nit: I wonder whether leaving a link to the Algolia docs for this `replace_all_objects` method as a comment would be helpful for future us?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -555,147 +555,188 @@ def add_metadata_to_algolia_objects(
...
-            Prefetch('catalog_queries', queryset=catalog_queries),
-            Prefetch('associated_content_metadata', queryset=associate_programs_query, to_attr='associate_programs'),
-        )
-        logger.info(
-            f'[ENTERPRISE_CATALOG_ALGOLIA_REINDEX] batch#{batch_num}: {content_metadata.count()} metadata found.'
-        )
+    catalog_queries = CatalogQuery.objects.prefetch_related(
+        'enterprise_catalogs',
+    )
+    associate_programs_query = ContentMetadata.objects.filter(content_type__in=[PROGRAM, LEARNER_PATHWAY])
```
adamstankiewicz: nit: unrelated to your immediate changes, but i believe this variable should be `associated_programs_query` (with the `d`)?

adamstankiewicz (approved):
these changes look good to me! all my comments are nits, mostly about prior code / namings.

pwnage101:
Refactored the function signatures a bit to be more sane/readable, overhauled the comments and moved them closer to the code they were describing, added a test for the new generator logic.

enterprise_catalog/apps/api/tasks.py
pwnage101: The explanation is here:

https://github.com/openedx/enterprise-catalog/blob/f5c761158f0622cc4f96d1427268485189326bb7/enterprise_catalog/apps/api/tasks.py#L788-L798

but it feels a bit sus to me because it seems to be describing functionality that doesn't need to leverage a caching framework.  However, according to git blame Alex D wrote it, so it's probably legit.  I'm not putting too much thought into it though since I'm treating it as out of scope for this PR.

iloveagent57: You're playing a dangerous game by assuming anything I write is legit :p 

I think the idea here is that the cache is serving as a distributed lock - I think this is meant to cover situations where two different algolia task instances run close together in time, due to manual updates of enterprise catalog/query records.

iloveagent57: But on the other hand - we probably _do_ want to update the record in that case, because it likely has new catalog or query uuids associated with it.  So yeah, I think we probably don't want this `_was_recently_indexed()` call at all.  But don't worry about it for this PR.

iloveagent57 (approved):

PR #624, opened 2023-06-23
fix: retry on POST verb too

johnnagro:
## Description

- by default `Retry` only retries certain HTTP verbs: `{'DELETE', 'GET', 'HEAD', 'OPTIONS', 'PUT', 'TRACE'}`
- we need `POST` to be retried
- using `None` reties all verbs
- https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/retry.py#L125-L132

### DIFF
--- a/enterprise_catalog/apps/api_client/base_oauth_with_retry.py
+++ b/enterprise_catalog/apps/api_client/base_oauth_with_retry.py
@@ -22,6 +22,7 @@ class BaseOAuthClientWithRetry(BaseOAuthClient):
             total=max_retries,
             backoff_factor=backoff_factor,
             backoff_jitter=backoff_jitter,
+            allowed_methods=None,
         )
         adapter = HTTPAdapter(max_retries=retry)
         self.client.mount('http://', adapter)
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api_client/base_oauth_with_retry.py
```diff
@@ -22,6 +22,7 @@ def __init__(
             total=max_retries,
             backoff_factor=backoff_factor,
             backoff_jitter=backoff_jitter,
+            allowed_methods=None,
```
adamstankiewicz: [suggestion] I might recommend not changing the default behavior here, but make it possible for the caller of `BaseOAuthClientWithRetry` to override the default with `None` (on a per-call basis, if possible). That way, we don't allow retries on all current and future `POST` calls, which feels fairly risky.

adamstankiewicz (approved):
LGTM, given there's only 1 `POST` call being made in `discovery.py`. that said, _IF_ there were a 2nd POST in discovery.py, it too would get retries, right? I still wonder if its possible to enable on per-API call basis. But I defer to you if its worth tackling that now given these changes are only impacting `discovery.py` with a single `POST` as is.

PR #626, opened 2023-06-26
fix: remove len() calls on a generator

pwnage101:
ENT-7311

### DIFF
--- a/enterprise_catalog/apps/api_client/algolia.py
+++ b/enterprise_catalog/apps/api_client/algolia.py
@@ -91,7 +91,7 @@ class AlgoliaSearchClient:
 
         return exists
 
-    def replace_all_objects(self, algolia_objects):
+    def replace_all_objects(self, algolia_objects):  # pragma: no cover
         """
         Clears all objects from the index and replaces them with a new set of objects. The records are
         replaced in the index without any downtime due to an atomic reindex.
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api_client/algolia.py
```diff
@@ -109,15 +109,10 @@ def replace_all_objects(self, algolia_objects):
             self.algolia_index.replace_all_objects(algolia_objects, {
                 'safe': True,  # wait for asynchronous indexing operations to complete
             })
-            logger.info(
-                'The %s Algolia index was successfully indexed with %s objects.',
-                self.ALGOLIA_INDEX_NAME,
-                len(algolia_objects),
-            )
+            logger.info('The %s Algolia index was successfully indexed.', self.ALGOLIA_INDEX_NAME)
```
pwnage101, line 112: I assert we don't need the count here because it's already captured here: https://github.com/openedx/enterprise-catalog/blob/17fc84fae2dacd00951b74f7b0b78cfcda28e0a2/enterprise_catalog/apps/api/tasks.py#L748-L750

iloveagent57 (approved):
I'm fine with throwing a `pragma: no cover` on the added line, so that the codecov diff check passes.

macdiesel (approved):

PR #628, opened 2023-06-26
feat: adding addition exceptions to retry

johnnagro:
## Description

- python requests has its own exception stack so urllib3 Retry doesn't know out of the box what to do


alex-sheehan-edx (approved):
lgtm

iloveagent57 (approved):

PR #633, opened 2023-06-27
fix: improve retry visibilty

johnnagro:
## Description

I am not getting the behavior i expect. This makes several changes: 

- It modifies a different error handler. I had modified the read error handler, where perhaps i should be doing the connection error handler because they treat the error state differently. The read error one also looks at the verb and status code and the exceptions i am interested in do not have a status.
- It adds a log line when retries are exhausted
- It explicitly sets the read and connect retry counts. The docs say the total takes precedence but I cannot find where the read and connect counts are set if not in the init.

https://github.com/urllib3/urllib3/blob/main/src/urllib3/util/retry.py#L460-L472

### DIFF
--- a/enterprise_catalog/apps/api_client/base_oauth_with_retry.py
+++ b/enterprise_catalog/apps/api_client/base_oauth_with_retry.py
@@ -14,11 +14,11 @@ class EnterpriseRetry(Retry):
     A class to tweak which Exceptions are considered retryable by Retry
     """
 
-    def _is_read_error(self, err: Exception) -> bool:
+    def _is_connection_error(self, err: Exception) -> bool:
         """
         Extending this method to account for requests.exceptions.RequestException
         """
-        super_result = super()._is_read_error(err)
+        super_result = super()._is_connection_error(err)
         # this is triggered by a ProtocolError but isnt in the stack
         local_result = isinstance(err, RequestException)
         return super_result or local_result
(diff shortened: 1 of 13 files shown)


enterprise_catalog/apps/api_client/base_oauth_with_retry.py
```diff
@@ -35,9 +35,18 @@ def increment(
...
+        logger.info(f'EnterpriseRetry retrying {method} to {url}, error={error}...')
         return super().increment(method, url, response, error, _pool, _stacktrace)
 
+    def is_exhausted(self) -> bool:
+        """
+        Are we out of retries? Adding logging.
+        """
+        super_result = super().is_exhausted()
+        if super_result:
+            logger.info('EnterpriseRetry retries exhausted.')
```
marlonkeating, line 47: Does this give enough logging info for us to know the context?

enterprise_catalog/apps/api_client/base_oauth_with_retry.py
```diff
@@ -56,6 +65,8 @@ def __init__(
         super().__init__(**kwargs)
         retry = EnterpriseRetry(
             total=max_retries,
+            read=max_retries,
+            connect=max_retries,
```
marlonkeating, line 69: Where is max_retries defined?

marlonkeating (changes requested):

enterprise_catalog/apps/api_client/base_oauth_with_retry.py
iloveagent57: it's a parameter to `__init__()`

enterprise_catalog/apps/api_client/base_oauth_with_retry.py
iloveagent57: I _think_ we'd get the task id in the logs along with this line, which is probably enough to go on?

iloveagent57 (approved):

marlonkeating (approved):

PR #634, opened 2023-06-27
feat: fix reindex-algolia task dict bug and add dry_run

pwnage101:
Sadly, the last attempt to add logging resulted in a dict iteration bug:
```
...
File "/edx/app/enterprise_catalog/enterprise_catalog/enterprise_catalog/apps/api/tasks.py", line 689, in _get_algolia_products_for_batch
    for algolia_object_id in algolia_products_by_object_id.keys():
RuntimeError: dictionary changed size during iteration
```

This PR fixes this bug, adds a unit test to cover it, and also implements --dry-run for the associated management command.

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -5,6 +5,7 @@ import logging
 import time
 from collections import defaultdict
 from datetime import timedelta
+from operator import itemgetter
 
 from celery import shared_task, states
 from celery.exceptions import Ignore
@@ -455,7 +456,7 @@ def _batched_metadata_with_queries(json_metadata, sorted_queries):
 
 @shared_task(base=LoggedTaskWithRetry, bind=True, default_retry_delay=UNREADY_TASK_RETRY_COUNTDOWN_SECONDS)
 @expiring_task_semaphore()
-def index_enterprise_catalog_in_algolia_task(self, force=False):  # pylint: disable=unused-argument
+def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):  # pylint: disable=unused-argument
     """
     Index course and program data in Algolia with enterprise-related fields.
 
@@ -466,6 +467,7 @@ def index_enterprise_catalog_in_algolia_task(self, force=False):  # pylint: disa
 
     Args:
         force (bool): If true, forces execution of task and ignores time since last run.
+        dry_run (bool): If true, does everything except call Algolia APIs.
     """
     try:
         if unready_tasks(update_full_content_metadata_task, ONE_HOUR).exists():
@@ -491,6 +493,7 @@ def index_enterprise_catalog_in_algolia_task(self, force=False):  # pylint: disa
         _reindex_algolia(
             indexable_content_keys=indexable_content_keys,
             nonindexable_content_keys=nonindexable_content_keys,
+            dry_run=dry_run,
         )
     except Exception as exep:
         logger.exception(f'[ENTERPRISE_CATALOG_ALGOLIA_REINDEX] reindex_algolia failed. Error: {exep}')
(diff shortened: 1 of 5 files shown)


enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -75,6 +75,7 @@ class Meta:
 
     # model fields
     content_key = factory.Sequence(lambda n: f'{str(n).zfill(5)}_metadata_item')
+    content_uuid = factory.LazyFunction(uuid4)
```
pwnage101, line 78: The set of changes to this ContentMetadataFactory are to make it possible to feed in a content_uuid from the caller down into the json_metadata.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -5,6 +5,7 @@
 import time
 from collections import defaultdict
 from datetime import timedelta
+from operator import itemgetter
```
iloveagent57, line 8: this is one of my favorite standard lib functions

enterprise_catalog/apps/api/tasks.py
```diff
@@ -573,6 +576,7 @@ def add_metadata_to_algolia_objects(
     _add_in_algolia_products_by_object_id(algolia_products_by_object_id, batched_metadata)
 
 
+# pylint: disable=too-many-statements
```
iloveagent57: 😭 someday, we will fix all of this

enterprise_catalog/apps/api/tasks.py
```diff
@@ -685,10 +689,14 @@ def _get_algolia_products_for_batch(
     # algolia objectID collision), do not send more than one.  Note that selection of duplicate content is
     # non-deterministic because we do not use order_by() on the queryset.
     context_accumulator.setdefault('generated_algolia_object_ids', set())
+    context_accumulator.setdefault('discarded_algolia_object_ids', {})
```
iloveagent57: nit: instead of `{}`, could you make this a `defaultdict(int)`?  could simplify the for-loop body below

enterprise_catalog/apps/api/tasks.py
```diff
@@ -722,7 +730,7 @@ def _index_content_keys_in_algolia(content_keys, algolia_client):
 
     Arguments:
         content_keys (list): List of indexable content_key strings.
-        algolia_client: Instance of an Algolia API client
+        algolia_client: Instance of an Algolia API client, or None if dry_run is enabled.
```
iloveagent57, line 730: nit/personal preference: don't use one parameter to sorta represent two different things, just explicitly add `dry_run=False` as a parameter, here, too.

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -975,3 +975,65 @@ def mock_replace_all_objects(products_iterable):
         mock_search_client().replace_all_objects.assert_called_once()
 
         mock_was_recently_indexed.assert_called_once_with(self.course_metadata_published.content_key)
+
+    # pylint: disable=unused-argument
+    @mock.patch('enterprise_catalog.apps.api.tasks._was_recently_indexed', return_value=False)
+    @mock.patch('enterprise_catalog.apps.api.tasks.get_initialized_algolia_client', return_value=mock.MagicMock())
```
iloveagent57, line 981: nit: this might be a good mock to add `autospec=True` to.

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
```diff
@@ -17,11 +17,18 @@ class Command(BaseCommand):
...
             default=False,
             action='store_true',
+            help='Force execution of celery task, ignoring time since last execution.',
+        )
+        parser.add_argument(
+            '--dry-run',
+            dest='dry_run',
+            action='store_true',
+            default=False,
+            help='Generate algolia products to index, but do not actually send them to algolia for indexing.',
```
iloveagent57, line 31: how are we gonna run this with `--dry-run` set from jenkins?

iloveagent57 (approved):
👍 handful of non-blocking suggestions

enterprise_catalog/apps/catalog/management/commands/reindex_algolia.py
pwnage101: i'd have to create a new temp job with the extra argument

enterprise_catalog/apps/api/tasks.py
pwnage101: Indeed, it was a side-effect of splitting things up that I was able to remove disable=too-many-statements, but logging forced me to re-add it.

enterprise_catalog/apps/api/tests/test_tasks.py
pwnage101: something I'd love to add later once I'm more familiar with autospec and not in a rush to unblock the pipeline.

enterprise_catalog/apps/api/tasks.py
pwnage101: will defer until a subsequent PR.

PR #637, opened 2023-06-29
fix: going back to the old retry

johnnagro:
## Description

- pulling out the `Retry` experiment
- going back to the older approach
- adding exception handling to the logic
- more opportunity in the future to generalize this or use other libraries


iloveagent57 (approved):
:sadpanda:

PR #644, opened 2023-07-11
fix: attempt to pass --dry-run correctly to task

pwnage101:
ENT-7311

## Description

Description of changes made

## Ticket Link

Link to the associated ticket

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -208,7 +208,7 @@ def expiring_task_semaphore(time_delta=None):
                     },
                 )
                 raise TaskRecentlyRunError(message)
-            return task(self, *args, *kwargs)
+            return task(self, *args, **kwargs)
         return wrapped_task
     return decorator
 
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -208,7 +208,7 @@ def wrapped_task(self, *args, **kwargs):
                     },
                 )
                 raise TaskRecentlyRunError(message)
-            return task(self, *args, *kwargs)
+            return task(self, *args, **kwargs)
```
pwnage101, line 211: My theory is that since dictionary keys are almost always truthy, dry_run was being forced to a True value.  Experiment:

```py
>>> def foobar(force=False, dry_run=False):
...     print(bool(force))
...     print(bool(dry_run))
... 
>>> kwargs = {'force': True, 'dry_run': False}
>>> args = []
>>> foobar(*args, *kwargs)
True
True
```

The fix is to pass kwargs as actual keyword arguments:

```py
>>> foobar(*args, **kwargs)
True
False
```

iloveagent57: crap, that's terrible 🤦

iloveagent57 (approved):

PR #645, opened 2023-07-11
fix: don't index courses with a verified upgrade deadline in the past

iloveagent57:
## Description
Update our enterprise-catalog algolia indexing to not include course runs if the verified upgrade has passed.  This will make it so that content, with a non-null verified course mode, that is not currently enrollable in the verified track, is not discoverable by learners.

## Ticket Link
https://openedx.atlassian.net/browse/ENT-7394

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -137,7 +137,9 @@ def _should_index_course(course_metadata):
     Replicates the B2C index check of whether a certain course should be indexed for search.
 
     A course should only be indexed for algolia search if it has a non-indexed advertiseable course run, at least
-    one owner, and a marketing url slug.
+    one owner, a marketing url slug, and the advertisable course run has a verified upgrade deadline in the future
+    (in the case where there is *no* verified seat, or the upgrade deadline is null, we consider the deadline
+    to be some arbitrarily distant date in the future).
     The course-discovery check that the course's partner is edX is included by default as the discovery API filters
     to the request's site's partner.
     The discovery check that the course has an availability level was decided to be a duplicate check that the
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -137,7 +137,9 @@ def _should_index_course(course_metadata):
     Replicates the B2C index check of whether a certain course should be indexed for search.
 
     A course should only be indexed for algolia search if it has a non-indexed advertiseable course run, at least
-    one owner, and a marketing url slug.
+    one owner, a marketing url slug, and the advertisable course run has a verified upgrade deadline in the future
+    (in the case where there is *no* verified seat, or the upgrade deadline is null, we consider the deadline
+    to be some arbitrariliy distant date in the future).
```
adamstankiewicz: nit: typo (`arbitrarily`)

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -164,6 +166,10 @@ def _should_index_course(course_metadata):
     if not is_course_run_active(advertised_course_run):
         return False
 
+    upgrade_deadline = _get_verified_upgrade_deadline(advertised_course_run)
```
adamstankiewicz: nit/suggestion: Might suggest naming this variable `enroll_by_deadline` as a generic name that accounts for both OCM course's "upgrade deadline" and Exec Ed's "registration deadline", which seems to match how we're starting communicate about these dates in the UX (e.g., "Enroll by March 3, 2023").
image

adamstankiewicz: For Exec Ed courses, the registration deadline doesn't come from the course run, but rather the `additional_metadata.registration_deadline` on the course object.

adamstankiewicz: [curious] I'm wondering if we should we be looking at all **paid** course run modes, not just the `verified` track here to remain "mode-agnostic" similar to some of the enterprise enrollment [logic](https://github.com/openedx/edx-enterprise/blob/d14b10e92af70f658d00ef9152f25be033ce8e70/enterprise/utils.py#L2242-L2250) [uses](https://github.com/openedx/edx-enterprise/blob/d14b10e92af70f658d00ef9152f25be033ce8e70/enterprise/views.py#L659-L685)?

```python
def get_best_mode_from_course_key(course_key):
    course_modes = [mode.slug for mode in CourseMode.objects.filter(course_id=course_key)]
    if best_mode := [mode for mode in BEST_MODE_ORDER if mode in course_modes]:
        return best_mode[0]
    return CourseModes.AUDIT
```

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -38,6 +45,24 @@ class AlgoliaUtilsTests(TestCase):
...
+            'seats': [
+                {'type': 'something-else', 'upgrade_deadline': _fake_upgrade_deadline(-100)}
+            ],
+        },
+        {
+            'expected_result': False,
+            'seats': [
+                {'type': 'verified', 'upgrade_deadline': _fake_upgrade_deadline(-1)}
+            ],
+        },
```
adamstankiewicz, line 65: Similar to another comment, may also be worth having a test case for Exec Ed courses given the data is stored in the `additional_metadata` for the associated course.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -164,10 +166,35 @@ def _should_index_course(course_metadata):
...
+        return False
+
     owners = course_json_metadata.get('owners') or []
     return not advertised_course_run.get('hidden') and len(owners) > 0
 
 
+def _has_enroll_by_deadline_passed(course_json_metadata, advertised_course_run):
+    """
+    Helper to determine if the enrollment deadline has passed for the given course
+    and advertised course run.  For COURSE types of courses, this is based on the
```
adamstankiewicz: nit: "For COURSE types of courses" 🤔 Is this supposed to be "For **OCM** types of courses"?

adamstankiewicz (approved):
LGTM, just 1 nit. Likely unrelated to this PR, but our codecov/project is failing in CI due to being 0.05% below our target project coverage of 92%. Not worth fixing in this PR, but wanted to flag anyways 🤷‍♂️ 

Related to [this comment](https://github.com/openedx/enterprise-catalog/pull/645#discussion_r1260294393), totally agree it makes sense to defer on it for now, but I do still think it's worth some consideration in the future.

PR #653, opened 2023-07-25
fix: logging/refactoring for ENT-7458

iloveagent57:
## Description
https://2u-internal.atlassian.net/browse/ENT-7458
* Refactors `_should_index_course()` for easier introspection.
* Adds logging to help understand a specific problem in ticket above.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -662,6 +662,11 @@ def index_content_keys_in_algolia(content_keys, algolia_client):  # pylint: disa
         # iterate over courses, programs and pathways and add their metadata to the list of objects to be indexed
         for metadata in filtered_content_metadata:
             content_key = metadata.content_key
+            # TODO: remove when https://2u-internal.atlassian.net/browse/ENT-7458 is resolved
+            if 'GTx+MGT6203x' in content_key:
+                logger.info(
+                    f'[ENT-7458] {content_key} will be added to Algolia index'
+                )
             # Check if we've indexed the course recently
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -662,6 +662,11 @@ def index_content_keys_in_algolia(content_keys, algolia_client):  # pylint: disa
         # iterate over courses, programs and pathways and add their metadata to the list of objects to be indexed
         for metadata in filtered_content_metadata:
             content_key = metadata.content_key
+            # TODO: remove when https://2u-internal.atlassian.net/browse/ENT-7458 is resolved
+            if 'GTx+MGT6203x' in content_key:
+                logger.info(
+                    f'[ENT-7458] {content_key} will be added to Algolia index'
+                )
```
iloveagent57, line 669: This is the "interesting part" - if we say we're going to index this course when `_should_index_course()` says we should *not* index it, we have a problem.

pwnage101: FYI: You're trying to find out if we hit this code path so it'll do the trick, but technically there's still a chance that a content_key that ends up getting printed still does not get indexed.  This is rare, and would only happen if it was part of no catalog query.

pwnage101 (approved):

enterprise_catalog/apps/api/tasks.py
iloveagent57: Hmm, good to know!  I'll come back to this if it becomes an issue.

PR #656, opened 2023-07-31
re-add constant-memory code and add "feat: do not index objects that are not indexable"

pwnage101:
Highly recommend reviewing this by looking at just the commit after the revert.  The revert essentially represents all attempts I've made since the beginning of this whole project to improve the memory consumption of the task.
```
commit 6fa565de1c9489edd75deda699fc4c0ca3311040
Author: Troy Sankey <tsankey@edx.org>
Date:   Fri Jul 28 07:49:01 2023 -0400

    feat: do not index objects that are not indexable
    
    This commit represents a change of behavior from the original
    implementation.  Previously, programs would be indexed just by touching
    an indexable object, regardless of whether the program passed the
    `_should_index_program()` test.  After this change, ONLY indexable
    programs are actually indexed.  Also add a ton of tests for specific
    scenarios to validate that claim.
    
    ENT-7311

commit 181d2ddc853297111d3948982cd906391ab269b9
Author: Troy Sankey <tsankey@edx.org>
Date:   Wed Jul 26 20:40:09 2023 -0400

    Revert "revert: Un-do all the reindex-algolia changes to improve memory consumption"
    
    This reverts commit 067ac7156d682a4a03916565be7453b77a055515.
```


enterprise_catalog/apps/api/tasks.py
```diff
@@ -641,26 +652,26 @@ def _get_algolia_products_for_batch(
...
             content_type__in=[COURSE_RUN, COURSE, PROGRAM],
             associated_content_metadata__content_type__in=[PROGRAM, LEARNER_PATHWAY],
             associated_content_metadata__content_key__in=content_keys_batch,
         )
-        # All programs, pathways to which any requested course belongs.
-        | Q(
-            content_type__in=[PROGRAM, LEARNER_PATHWAY],
-            associated_content_metadata__content_type__in=[COURSE, PROGRAM],
-            associated_content_metadata__content_key__in=content_keys_batch,
-        )
```
pwnage101: No need to climb up the tree any more.  If an indexable course is contained in a hidden program, don't index the program just because the two objects are associated!  The course can simply stand alone, not part of the program from algolia's perspective.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -641,26 +652,26 @@ def _get_algolia_products_for_batch(
...
     ).prefetch_related(
         Prefetch('catalog_queries', queryset=all_catalog_queries),
     )
+    # Perform filtering of non-indexable objects in-memory because the list may be too long to shove into a SQL query.
+    content_metadata_no_courseruns = [
+        cm for cm in content_metadata_no_courseruns
+        if cm.content_key in all_indexable_content_keys
```
iloveagent57, line 686: nit: we should try to make `all_indexable_content_keys` a set, if possible.

iloveagent57 (approved):
One nit that's worth addressing, rest looks good!

enterprise_catalog/apps/api/tasks.py
pwnage101: ah, good point.  that probably makes the lookup O(log(n)) instead of O(n), right?

iloveagent57: It should make the lookup of each `content_key` O(1)

PR #665, opened 2023-08-15
docs: add clarifying comments for extra targets in Dockerfile

pwnage101:
The newrelic and legacy_devapp targets are not currently in use, and migrating the service to Kubernetes cements the fact that they will never be used again.

ENT-5540

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -69,11 +69,19 @@ CMD ["gunicorn", "--workers=2", "--name", "enterprise_catalog", "-c", "/edx/app/
 # bust the image cache
 COPY . /edx/app/enterprise_catalog/enterprise_catalog
 
+###############################################################
+# Create newrelic image used by the experimental docker shim. #
+###############################################################
+# TODO: remove this after we migrate to k8s since it will serve no more purpose.
 FROM app as newrelic
 RUN pip install newrelic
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -69,11 +69,19 @@ CMD ["gunicorn", "--workers=2", "--name", "enterprise_catalog", "-c", "/edx/app/
 # bust the image cache
 COPY . /edx/app/enterprise_catalog/enterprise_catalog
 
+###############################################################
+# Create newrelic image used by the experimental docker shim. #
```
adzuci, line 73: I vaguely remember the docker shim experiment but thought it was long over, do we know where this logic might be in use currently?

adzuci (approved):
These changes are mostly TODOs so I see no reason not to merge them, I am surprised that additional changes are not needed, but think it's worth testing as is before hunting for missing items.

Dockerfile
pwnage101: It's turned off, controlled by a boolean ansible variable.  So in that sense, it's already not serving a purpose currently.  However, I'd still advocate we keep this around to minimize changes until after the cutover.

https://github.com/edx/configuration/blob/a791a9f7857ed0c9515a5399f4b7d0d239080e17/playbooks/roles/enterprise_catalog/defaults/main.yml#L47

PR #666, opened 2023-08-18
feat: normalize start/end/enroll_by dates across content types

pwnage101:
ENT-7545

## Ticket Link

[[enterprise-catalog] Expose net-new fields on the existing `content-metadata` API](https://2u-internal.atlassian.net/browse/ENT-7545)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/constants.py
+++ b/enterprise_catalog/apps/api/constants.py
@@ -1,0 +1,23 @@
+"""
+Constants for api app.
+"""
+
+
+class CourseMode():
+    """
+    Content metadata course mode keys.
+
+    Copied from https://github.com/edx/edx-platform/blob/831a8bcc/common/djangoapps/course_modes/models.py#L164
+    """
+    HONOR = "honor"
+    PROFESSIONAL = "professional"
+    VERIFIED = "verified"
+    AUDIT = "audit"
+    NO_ID_PROFESSIONAL_MODE = "no-id-professional"
+    CREDIT_MODE = "credit"
+    MASTERS = "masters"
+    EXECUTIVE_EDUCATION = "executive-education"
+    PAID_EXECUTIVE_EDUCATION = "paid-executive-education"
+    UNPAID_EXECUTIVE_EDUCATION = "unpaid-executive-education"
+    PAID_BOOTCAMP = "paid-bootcamp"
+    UNPAID_BOOTCAMP = "unpaid-bootcamp"
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -16,6 +16,7 @@ from django.db.utils import OperationalError
 from django_celery_results.models import TaskResult
 from requests.exceptions import ConnectionError as RequestsConnectionError
 
+from enterprise_catalog.apps.api.constants import CourseMode
 from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.catalog.algolia_utils import (
     ALGOLIA_FIELDS,
@@ -61,6 +62,16 @@ UNREADY_TASK_RETRY_COUNTDOWN_SECONDS = 60 * 5
 
 # ENT-4980 every batch "shard" record in Algolia should have all of these that pertain to the course
 EXPLORE_CATALOG_TITLES = ['A la carte', 'Business', 'Education']
+
+# The closer a mode is to the beginning of this list, the more likely a seat with that mode will be used to find the
+# upgrade deadline for the course (and course run).
+BEST_MODE_ORDER = [
+    CourseMode.VERIFIED,
+    CourseMode.PROFESSIONAL,
+    CourseMode.NO_ID_PROFESSIONAL_MODE,
+    CourseMode.UNPAID_EXECUTIVE_EDUCATION,
+    CourseMode.AUDIT,
+]
 
 
 def _fetch_courses_by_keys(course_keys):
@@ -253,6 +264,68 @@ def update_full_content_metadata_task(self, force=False):  # pylint: disable=unu
     _update_full_content_metadata_program(content_keys)
 
 
+def _find_best_mode_seat(seats):
+    """
+    Find the seat with the "best" course mode.  See BEST_MODE_ORDER to find which modes are best.
+    """
+    sort_key_for_mode = {mode: index for (index, mode) in enumerate(BEST_MODE_ORDER)}
+
+    def sort_key(seat):
+        """
+        Get a sort key (int) for a seat dictionary based on the position of its mode in the BEST_MODE_ORDER list.
+
+        Modes not found in the BEST_MODE_ORDER list get sorted to the end of the list.
+        """
+        mode = seat['type']
+        return sort_key_for_mode.get(mode, len(sort_key_for_mode))
+
+    sorted_seats = sorted(seats, key=sort_key)
+    if sorted_seats:
+        return sorted_seats[0]
+    return None
+
+
+def _normalize_course_metadata(course_metadata_record):
+    """
+    Add normalized metadata keys with values calculated by normalizing existing keys. This will be helpful for
+    downstream consumers which no longer will need to do their own independent normalization.
+
+    At the time of writing, output normalized metadata keys incldue:
+
+    * normalized_metadata.start_date: When the course starts
+    * normalized_metadata.end_date: When the course ends
+    * normalized_metadata.enroll_by_date: The deadline for enrollment
+
+    Note that course-type-specific definitions of each of these keys may be more nuanced.
+    """
+    json_meta = course_metadata_record.json_metadata
+    normalized_metadata = {}
+
+    # For each content type, find the values that correspond to the desired output key.
+    if course_metadata_record.is_exec_ed_2u_course:
+        # First case covers Exec Ed courses.
+        additional_metdata = json_meta.get('additional_metadata', {})
+        normalized_metadata['start_date'] = additional_metdata.get('start_date')
+        normalized_metadata['end_date'] = additional_metdata.get('end_date')
+        normalized_metadata['enroll_by_date'] = additional_metdata.get('registration_deadline')
+    else:
+        # Else case covers OCM courses.
+        advertised_course_run_uuid = json_meta.get('advertised_course_run_uuid')
+        advertised_course_run = _get_course_run_by_uuid(json_meta, advertised_course_run_uuid)
+        if advertised_course_run is not None:
+            normalized_metadata['start_date'] = advertised_course_run.get('start')
+            normalized_metadata['end_date'] = advertised_course_run.get('end')
+            all_seats = advertised_course_run.get('seats', [])
+            seat = _find_best_mode_seat(all_seats)
+            if seat:
+                normalized_metadata['enroll_by_date'] = seat.get('upgrade_deadline')
+            else:
+                logger.info(f"No Seat Found for course run '{advertised_course_run.get('key')}'. Seats: {all_seats}")
+
+    # Add normalized values to net-new keys:
+    json_meta['normalized_metadata'] = normalized_metadata
+
+
 def _update_full_content_metadata_course(content_keys):
     """
     Given content_keys, finds the associated ContentMetadata records with a type of course and looks up the full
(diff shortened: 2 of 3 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -296,8 +368,12 @@ def _update_full_content_metadata_course(content_keys):
...
                 logger.error('Could not find ContentMetadata record for content_key %s.', content_key)
                 continue
-            # exec ed updates the start/end dates in additional metadata, so we have to manually
-            # move that over to our variables that we use
+
+            # Merge the full metadata from discovery's /api/v1/courses into the local metadata object.
+            metadata_record.json_metadata.update(course_metadata_dict)
+
+            # Exec ed provides the start/end dates in additional_metadata, so we should copy those over to the keys that
+            # we use (inside the advertised course run).
```
pwnage101, line 377: Note about this if-block:  Prior to this PR, it ran BEFORE merging the full metadata from `/api/v1/courses/` into the local metadata object, but now it runs AFTER merging full metadata. The `additional_metadata` key is not present in the `/api/v1/search/all/` response, only in `/api/v1/courses/`, so the only way this code worked was by running the task twice on a given course.

Old behavior:

1. Load basic metadata from `/api/v1/search/all/`.
2. Attempt to copy start/end from additional_metdata (fails because `additional_metdata` doesn't exist yet).
3. Merge full metadata from `/api/v1/courses/` (adds `additional_metdata`).
4. Load basic metadata from `/api/v1/search/all/`.
5. Attempt to copy start/end from additional_metdata (success).

New behavior:

1. Load basic metadata from `/api/v1/search/all/`.
2. Merge full metadata from `/api/v1/courses/` (adds `additional_metdata`).
3. Attempt to copy start/end from additional_metdata (success).

enterprise_catalog/apps/api/tasks.py
```diff
@@ -306,13 +382,10 @@ def _update_full_content_metadata_course(content_keys):
                 for run in json_meta.get('course_runs'):
                     if run.get('uuid') == course_run_uuid:
                         run.update({'start': start_date, 'end': end_date})
-                course_run = _get_course_run_by_uuid(json_meta, course_run_uuid)
-                if course_run is not None:
-                    course_run_meta = metadata_by_key.get(course_run.get('key'))
-                    if hasattr(course_run_meta, 'json_metadata'):
-                        course_run_meta.json_metadata.update({'start': start_date, 'end': end_date})
-                        modified_content_metadata_records.append(course_run_meta)
```
pwnage101, line 314: This section is being removed because it is a no-op.  Course runs are never processed in this task.

https://github.com/openedx/enterprise-catalog/blob/d868759/enterprise_catalog/apps/api/tasks.py#L250-L251

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -366,51 +367,85 @@ def test_update_full_metadata(self, mock_oauth_client, mock_partition_course_key
...
+        metadata_1 = ContentMetadata.objects.get(content_key=course_key_1)
+        metadata_2 = ContentMetadata.objects.get(content_key=course_key_2)
+        metadata_3 = ContentMetadata.objects.get(content_key=course_key_3)
 
-        # add aggregation_key and uuid to course objects since they should now exist
-        # after merging the original json_metadata with the course metadata
-        course_data_1.update(metadata_1.json_metadata)
-        course_data_2.update(metadata_2.json_metadata)
-        course_data_1.update({'aggregation_key': 'course:fakeX'})
-        course_data_2.update({'aggregation_key': 'course:testX'})
```
pwnage101, line 410: This is more of the testing pattern we're trying to move away from.  It's like doing "x = y; assert y = x".  Instead we should test each expected key manually.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -62,6 +63,16 @@
...
 
+# The closer a mode is to the beginning of this list, the more likely a seat with that mode will be used to find the
+# upgrade deadline for the course (and course run).
+BEST_MODE_ORDER = [
+    CourseMode.VERIFIED,
+    CourseMode.PROFESSIONAL,
+    CourseMode.NO_ID_PROFESSIONAL_MODE,
+    CourseMode.UNPAID_EXECUTIVE_EDUCATION,
+    CourseMode.AUDIT,
+]
```
iloveagent57, line 74: I think you need to tag a `.value` on each of these enum symbols.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -253,6 +264,67 @@ def update_full_content_metadata_task(self, force=False):  # pylint: disable=unu
...
+    if sorted_seats:
+        return sorted_seats[0]
+
+    return None
+
+
+def _normalize_course_metadata(course_metadata_record):
+    """
+    Add net-new metadata keys with values calculated by normalizing existing keys. This will be helpful for downstream
+    consumers which no longer will need to do their on independent normalization.
```
iloveagent57: nit: "on" -> "own"

enterprise_catalog/apps/api/tasks.py
```diff
@@ -253,6 +264,67 @@ def update_full_content_metadata_task(self, force=False):  # pylint: disable=unu
...
+        advertised_course_run = _get_course_run_by_uuid(json_meta, advertised_course_run_uuid)
+        if advertised_course_run is not None:
+            net_new_fields['start_date'] = advertised_course_run.get('start')
+            net_new_fields['end_date'] = advertised_course_run.get('end')
+            all_seats = advertised_course_run.get('seats', [])
+            seat = _find_best_mode_seat(all_seats)
+            if seat:
+                net_new_fields['enroll_by_date'] = seat.get('upgrade_deadline')
+            else:
+                logger.info(f"No Seat Found. Seats: {all_seats}")
```
iloveagent57: Nit: it'd be helpful to log the content key here, too.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -253,6 +264,67 @@ def update_full_content_metadata_task(self, force=False):  # pylint: disable=unu
...
+
+    # Add normalized values to net-new keys:
+    json_meta.update(net_new_fields)
```
iloveagent57: Open question: is it worth nesting these in a dict called something like `normalized_metadata` inside of the `json_metadata` dictionary?

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -366,51 +367,85 @@ def test_update_full_metadata(self, mock_oauth_client, mock_partition_course_key
...
+                'key': f'course-v1:{course_key_3}+1',
+                'uuid': course_run_3_uuid,
+                # The task should copy these dates into net-new top level fields.
+                'start': '2023-03-01T00:00:00Z',
+                'end': '2023-03-01T00:00:00Z',
+                'seats': [
+                    {
+                        'type': str(CourseMode.VERIFIED),
+                        'upgrade_deadline': '2023-02-01T00:00:00Z',
+                        # WTF?
```
iloveagent57: Nit: do not want?

iloveagent57 (commented):
Couple nits/suggestions.

enterprise_catalog/apps/api/tests/test_tasks.py
pwnage101: oh yea, that was there to remind me to ask WTF the upgrade_deadline_override is all about.  There's no description in the model: https://github.com/openedx/course-discovery/blob/837d3eb8671c80a86c36a67c1a10d4118817f928/course_discovery/apps/course_metadata/models.py#L2863

enterprise_catalog/apps/api/tasks.py
pwnage101: That would help isolate the artifacts of this processing step, potentially avoiding unintentional clobbering in the future.  Seems okay to me!

pwnage101: updated to add a top-level json_metadata dict.

enterprise_catalog/apps/api/tests/test_tasks.py
pwnage101: confirmed offline that upgrade_deadline already takes upgrade_deadline_override into account, so we can safely ignore the latter.

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -356,6 +358,45 @@ def setUpTestData(cls):
...
+        # Test that even if no "best" seat types are present, one is still selected.
+        {
+            'seats': [
+                {'type': CourseMode.PAID_EXECUTIVE_EDUCATION.value, 'sku': 'SKU-1'},
+            ],
+            'expected_seat': {'type': CourseMode.PAID_EXECUTIVE_EDUCATION.value, 'sku': 'SKU-1'},
+        },
+    )
+    @ddt.unpack
+    def test_find_best_mode_seat(self, seats, expected_seat):
```
pwnage101, line 393: FYI: added this test since last review.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -253,6 +264,68 @@ def update_full_content_metadata_task(self, force=False):  # pylint: disable=unu
...
+        """
+        Assign a weight to the seat dictionary based on the position of its type in best mode order list.
+        """
+        mode_type = mode['type']
+        if mode_type in BEST_MODE_ORDER:
+            return len(BEST_MODE_ORDER) - BEST_MODE_ORDER.index(mode_type)
+        else:
+            return 0
+
+    sorted_seats = sorted(seats, key=sort_key, reverse=True)
```
iloveagent57: code golf nit/alternative approach:
```python
preferred_modes = {mode: index for (index, mode) in enumerate(BEST_MODE_ORDER)}
def sort_key(seat):
    mode_type = seat['type']
    return preferred_modes.get(mode_type, len(preferred_modes)
sorted_seats = sorted(seats, key=sort_key) # don't reverse sort
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -253,6 +264,68 @@ def update_full_content_metadata_task(self, force=False):  # pylint: disable=unu
     _update_full_content_metadata_program(content_keys)
 
 
+def _find_best_mode_seat(seats):
+    """
+    Find the seat by best course mode.  See BEST_MODE_ORDER to find which modes are best.
+    """
+    def sort_key(mode):
```
iloveagent57: nit: should this param be `seat` instead of `mode`?

enterprise_catalog/apps/api/constants.py
```diff
@@ -0,0 +1,25 @@
+"""
+Constants for api app.
+"""
+
+from enum import Enum
+
+
+class CourseMode(Enum):
```
iloveagent57: Maybe we just make this a plain ole' class (basically a set of namespaced constants)?  I don't think the `Enum` is doing any work for us?

iloveagent57 (approved):

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -468,77 +539,108 @@ def test_update_full_metadata_program(self, mock_oauth_client, mock_partition_pr
...
+                    'key': course_run_key,
+                    # Use dummy 2022 dates that we will assert are overwritten.
+                    'start': '2022-03-01T00:00:00Z',
+                    'end': '2022-03-01T00:00:00Z',
+                }],
                 'programs': [],
-                'additional_metadata': {
-                    'start_date': '2023-03-01T00:00:00Z',
-                    'end_date': '2023-04-09T23:59:59Z',
-                }
```
pwnage101, line 519: Remove `additional_metadata` from the starter course metadata to properly simulate an object freshly seeded by `/api/v1/search/all/` via a different task.

enterprise_catalog/apps/api/constants.py
```diff
@@ -0,0 +1,23 @@
+"""
+Constants for api app.
+"""
+
+
+class CourseMode():
```
adamstankiewicz, line 6: [nit/inform]: Similar course mode constants are also defined under `enterprise_catalog/apps/catalog/constants.py` (e.g., `CONTENT_COURSE_TYPE_ALLOW_LIST`). We could likely consolidate to ensure that those hardcoded course modes also rely on this new `CourseMode` class.

That said, we might want to define this constant under the `constants.py` file under the catalog app, so they can be more readily shared across both the `api` and `catalog` apps.

I wouldn't say this is blocking feedback on this PR; likely something I can tackle in [ENT-7548](https://2u-internal.atlassian.net/browse/ENT-7548), which relates to ensuring these normalized dates get into Algolia so they're available across all course types.

adamstankiewicz (approved):
LGTM!

PR #675, opened 2023-09-21
feat: catalog query filtering without elasticsearch

johnnagro:
## Description

Given a catalog query and some content metadata - can we determine if the content matches the query _without_ using Elasticsearch (ie `search/all`)? 

This will be useful when and if:
- crawl content once, determine catalogs from our copy (rather than re-crawl content for every query it belongs to)
- grind `search/all` less
- we want to process content metadata update events (from event bus)
- support catalog contents with similar query syntax but broader sources of content metadata (not everything in discovery)

- https://2u-internal.atlassian.net/browse/ENT-7712

## Testing

- I wrote some tests
- i parsed all our production queries and they're syntactically valid based on this code
- if we like this, i'll write a job in prod to compare our query tool to Elasticsearch

### DIFF
--- a/enterprise_catalog/apps/catalog/filters.py
+++ b/enterprise_catalog/apps/catalog/filters.py
@@ -1,0 +1,134 @@
+"""
+Utility functions for catalog query filtering without elasticsearch
+"""
+import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+SUPPORTED_FILTER_COMPARISONS = [
+    'exact',
+    'not',
+    'exclude',
+    'gt',
+    'gte',
+    'lt',
+    'lte',
+]
+
+
+class QueryFilterException(Exception):
+    """
+    An exception for content catalog query filtering
+    """
+
+
+def fix_common_query_key_mistakes(raw_query_key):
+    """
+    In production many queries have odd typos
+    which seem to have been copypasta-proliferated
+    """
+    corrections_for_typos = {
+        'aggregation_key': [
+            'aggregration__key',
+            'aggregation__key',
+        ],
+        'org__exclude': [
+            'org__exempt',
+        ],
+    }
+    for correction, typos in corrections_for_typos.items():
+        if raw_query_key in typos:
+            return correction
+    return raw_query_key
+
+
+def extract_field_and_comparison_kind(raw_query_key):
+    """
+    Taking an query key, extra the content_metadata
+    field name and the kind of comparison matching
+    should be used.
+    """
+    field = None
+    # comparison_kind defaults to "exact match"
+    comparison_kind = 'exact'
+    split_query_key = raw_query_key.split("__")
+    if len(split_query_key) == 2:
+        field, comparison_kind = split_query_key
+    elif len(split_query_key) > 2:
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/catalog/filters.py
```diff
@@ -0,0 +1,126 @@
...
+    In production many queries have odd typos
+    which seem to have been copypasta-proliferated
+    """
+    raw_query_key = raw_query_key.lower()
+    if raw_query_key == 'aggregration__key':
+        return 'aggregation_key'
+    elif raw_query_key == 'aggregation__key':
+        return 'aggregation_key'
+    elif raw_query_key == 'org__exempt':
+        return 'org__exclude'
```
marlonkeating: nit: Could put this in a dict to reduce duplication and allow adding more cases without further stacking the elif tower
```suggestion
    corrections_for_typos = {
      'aggregation_key': ['aggregration__key', 'aggregation__key'],
      'org__exclude': ['org__exempt']
    }
    for correction, typos in corrections_for_typos.items():
      if raw_query in typos:
        return correction;
```

enterprise_catalog/apps/catalog/filters.py
```diff
@@ -0,0 +1,126 @@
...
+def extract_field_and_comparison_kind(raw_query_key):
+    """
+    Taking an query key, extra the content_metadata
+    field name and the kind of comparison matching
+    should be used.
+    """
+    field = None
+    # comparison_kind defaults to "exact match"
+    comparison_kind = 'exact'
+    if len(raw_query_key.split("__")) == 2:
```
marlonkeating: `raw_query_key.split("__")` expression gets duplicated a couple times
```suggestion
    field__comparison = raw_query_key.split("__")
    if len(field__comparison) == 2:
```

enterprise_catalog/apps/catalog/filters.py
```diff
@@ -0,0 +1,126 @@
...
+        if isinstance(query_value, list):
+            field_results = []
+            for query_value_item in query_value:
+                this_field_result = field_comparison(query_value_item, content_value, comparison_kind)
+                logger.debug(f'{query_value_item}, {content_value}, {comparison_kind} -> {this_field_result}')
+                field_results.append(this_field_result)
+            if comparison_kind == 'exact':
+                field_result = any(field_results)
+            else:
+                field_result = all(field_results)
```
marlonkeating: Should the logic be reversed here?  I would expect 'exact' matching to require matching all results.

marlonkeating (changes requested):

enterprise_catalog/apps/catalog/filters.py
johnnagro: its confusing but "exact" in this case is really mean to be more like "includes" so any of them need to match. if the course key is "edX+Demo" and the query key is like "course_key: [ "edX+Demo", "MIT+Demo"]" you'd want this to match where as the else case is an "excludes" which means NONE of them can match... lemme think about the right way to document this and/or if we need to code this so its easier to understand.

marlonkeating (approved):

PR #681, opened 2023-09-29
fix: cache debugging in discovery client

iloveagent57:
## Description
Add some logging to determine if/when we're getting cache misses/hits when fetching catalog query content from discovery.

### DIFF
--- a/enterprise_catalog/apps/api_client/discovery_cache.py
+++ b/enterprise_catalog/apps/api_client/discovery_cache.py
@@ -14,6 +14,8 @@ from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 
 LOGGER = getLogger(__name__)
 
+_CACHE_MISS = object()
+
 
 class CatalogQueryMetadata:
     """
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api_client/discovery_cache.py
```diff
@@ -52,16 +54,24 @@ def _get_catalog_query_metadata(self, catalog_query):
                 Empty dictionary if no data found in cache or from API.
         """
         cache_key = DISCOVERY_CATALOG_QUERY_CACHE_KEY_TPL.format(id=self.catalog_query.id)
-        catalog_query_data = cache.get(cache_key)
-        if not catalog_query_data:
+        catalog_query_data = cache.get(cache_key, _CACHE_MISS)
+        if catalog_query_data is not _CACHE_MISS:
+            LOGGER.info('Cache HIT for CatalogQuery id=%s', self.catalog_query.id)
```
johnnagro, line 59: we may want to set `catalog_query_data` to `None` in this `if` branch so ensure we don't mess up any downstream logic when returning `object()` on a miss?

johnnagro (approved):

PR #689, opened 2023-10-03
feat: mgmt command queue task for logs to splunk

johnnagro:
- mgmt task `logger.info` run from jenkins doesnt end up in splunk
- rework this to run as a task so we can see splunk logs


alex-sheehan-edx (approved):
🚢

PR #694, opened 2023-10-05
feat: remove disused discovery cache

johnnagro:
## Description

Remove the disused Discovery API Client cache.

- https://2u-internal.atlassian.net/browse/ENT-7772


enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -238,3 +238,44 @@ def get_programs(self, query_params=None):
             )
 
         return programs
+
+
+class CatalogQueryMetadata:
```
johnnagro, line 243: this class is being moved out of the discovery cache file, it packages up the data from the discovery api

iloveagent57 (approved):

PR #703, opened 2023-10-25
feat: add content_price field to normalized_metadata

katrinan029:
## Description

Currently, we add business logic in the frontend apps to find the content price for each course. This PR adds `content_price` to `normalized_metadata` to remove that logic from the frontend. 

## Ticket Link

Link to the associated ticket
[JIRA](https://2u-internal.atlassian.net/browse/ENT-7591)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -305,11 +305,12 @@ def _normalize_course_metadata(course_metadata_record):
     if course_metadata_record.is_exec_ed_2u_course:
         # First case covers Exec Ed courses.
         additional_metadata = json_meta.get('additional_metadata', {})
-        entitlements = json_meta.get('entitlements', [])
         normalized_metadata['start_date'] = additional_metadata.get('start_date')
         normalized_metadata['end_date'] = additional_metadata.get('end_date')
         normalized_metadata['enroll_by_date'] = additional_metadata.get('registration_deadline')
-        normalized_metadata['content_price'] = entitlements[0]['price']
+        for entitlement in json_meta.get('entitlements', []):
+            if entitlement.get('mode') == CourseMode.PAID_EXECUTIVE_EDUCATION:
+                normalized_metadata['content_price'] = entitlement.get('price')
     else:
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -290,31 +290,34 @@ def _normalize_course_metadata(course_metadata_record):
...
-        additional_metdata = json_meta.get('additional_metadata', {})
-        normalized_metadata['start_date'] = additional_metdata.get('start_date')
-        normalized_metadata['end_date'] = additional_metdata.get('end_date')
-        normalized_metadata['enroll_by_date'] = additional_metdata.get('registration_deadline')
+        additional_metadata = json_meta.get('additional_metadata', {})
+        entitlements = json_meta.get('entitlements', [])
+        normalized_metadata['start_date'] = additional_metadata.get('start_date')
+        normalized_metadata['end_date'] = additional_metadata.get('end_date')
+        normalized_metadata['enroll_by_date'] = additional_metadata.get('registration_deadline')
+        normalized_metadata['content_price'] = entitlements[0]['price']
```
adamstankiewicz: [question] Will this throw an exception if there happens to not be any entitlements returned (i.e., the fallback to `[]` on line 308)? 

```python3
[][0] # throws "IndexError: list index out of range"
```
It may make sense to iterate over the entitlements (similar to [this method](https://github.com/openedx/enterprise-subsidy/blob/9a7e56bddcc31b7c00a591d51eae0089890c043c/enterprise_subsidy/apps/content_metadata/api.py#L63-L65) in enterprise-subsidy), which would also give the opportunity to verify the entitlement's `mode` matches up with the course mode `paid-executive-education` ([source](https://github.com/openedx/enterprise-catalog/blob/e4c43c878ffdcfb64159fe44b5d2eb3674856973/enterprise_catalog/apps/api/constants.py#L20C5-L20C29)) as well.

katrinan029: I refactored the code to reflect this feedback!

adamstankiewicz (approved):
LGTM!

PR #733, opened 2023-12-20
fix: fix logging related to ENT-7893

iloveagent57:
Was previously logging the `__str__` of catalog query objects, which caused unfortunate line breaks in our logs, making it difficult/impossible to correlate catalog query identifiers with content identifiers in the impacted logging debug lines.  This change makes it so we only log the catalog query id.

https://openedx.atlassian.net/browse/ENT-7893

## Post-review


adamstankiewicz (approved):
:shipit:

PR #749, opened 2024-01-24
feat: indexing bayesian average ranking for courses in algolia

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-8193

## Post-review


enterprise_catalog/settings/base.py
```diff
@@ -380,9 +380,11 @@
 
 # How long we keep API Client data in cache. (seconds)
 ONE_HOUR = 60 * 60
+ONE_DAY = ONE_HOUR * 24
```
johnnagro: i'd make this much, much longer - maybe "forever" - have a background job overwrite it

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -331,6 +337,21 @@ def get_algolia_object_id(content_type, uuid):
     return None
 
 
+def get_course_bayesian_average(course):
+    """
+    https://www.algolia.com/doc/guides/managing-results/must-do/custom-ranking/how-to/bayesian-average/
+    """
+    cache_key = DISCOVERY_AVERAGE_COURSE_REVIEW_CACHE_KEY
```
johnnagro, line 382: i'd break this out into a getter function that falls back to a settings value. the global average isnt going to move much, i'd calculate it by hand and add it to the base.py settings as the fallback

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -149,6 +151,15 @@ def get_course_reviews(self, course_keys=None):
...
         }
 
+        if set_average_course_rating:
+            rolling_rating_sum = 0.0
+            for value in results.values():
+                rolling_rating_sum += float(value.get('avg_course_rating'))
+
+            total_average_course_rating = rolling_rating_sum / len(results)
+            cache_key = DISCOVERY_AVERAGE_COURSE_REVIEW_CACHE_KEY
+            cache.set(cache_key, total_average_course_rating, settings.DISCOVERY_AVERAGE_COURSE_REVIEW_CACHE_TIMEOUT)
```
johnnagro: i'd break this out into its own function rather than include it in the reviews getter, then make a mgmt command which calls the discrete method

johnnagro (changes requested):
okay so, close but imho i'd tweal things a bit:
- i'd calculate and set a constant default that is roughly correct in `base.py`
- i'd make a "get_global_average" that tries to read from cache then falls back
- i'd make a "calculate_and_set_global_average" and call that from a simple mgmt task

enterprise_catalog/apps/api_client/discovery.py
```diff
@@ -149,6 +151,15 @@ def get_course_reviews(self, course_keys=None):
             )
         }
 
+        if set_average_course_rating:
+            rolling_rating_sum = 0.0
+            for value in results.values():
+                rolling_rating_sum += float(value.get('avg_course_rating'))
+
+            total_average_course_rating = rolling_rating_sum / len(results)
```
bcitro: I think you also need to factor in the number of reviews for each course to get the true global average.  So in your rolling sum you'd want to include (`avg_course_rating` * `reviews_count`), and then ultimately divide by the sum of `reviews_count`.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -331,6 +337,21 @@ def get_algolia_object_id(content_type, uuid):
...
+    total_average_rating = cache.get(cache_key)
```
bcitro: I think here we had talked about calculating the current global review average and falling back to that as a default if this cache key isn't set. From this snowflake query I calculated that to be `4.5`:

`SELECT SUM(REVIEWS_COUNT * AVG_COURSE_RATING)/SUM(REVIEWS_COUNT) FROM enterprise.course_reviews`

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -331,6 +337,21 @@ def get_algolia_object_id(content_type, uuid):
...
+    avg_review = course.get('avg_course_rating')
+    ratings_count = course.get('reviews_count')
+    if avg_review is not None and ratings_count is not None:
+        bayes_avg = ((avg_review * ratings_count) + (total_average_rating * 15)) / (ratings_count + 15)
```
bcitro: I would make `15` a constant

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -331,6 +341,69 @@ def get_algolia_object_id(content_type, uuid):
...
+        settings.DISCOVERY_AVERAGE_COURSE_REVIEW_CACHE_TIMEOUT
+    )
+
+
+def get_global_course_review_avg():
+    """
+    Fetch the calculated global course review average from py-cache
+    """
+    cache_key = DISCOVERY_AVERAGE_COURSE_REVIEW_CACHE_KEY
+    total_average_rating = cache.get(cache_key)
```
johnnagro: ah i see now that we can pass the default into the getter, so you could do:
`cache.get(cache_key, COURSE_REVIEW_BASE_AVG_REVIEW_SCORE)`
then you wouldnt need the `if None` check later

enterprise_catalog/apps/catalog/management/commands/tests/test_set_global_average_course_rating_value.py
```diff
@@ -0,0 +1,52 @@
...
+            json_metadata={'avg_course_rating': 4, 'reviews_count': 10}
+        )
+        ContentMetadataFactory(json_metadata={})
+        call_command(self.command_name)
+        expected_total_average = ((5 * 20) + (4 * 10)) / 30
+
+        mock_cache.set.assert_called_with(
+            DISCOVERY_AVERAGE_COURSE_REVIEW_CACHE_KEY,
+            expected_total_average,
+            settings.DISCOVERY_AVERAGE_COURSE_REVIEW_CACHE_TIMEOUT,
```
johnnagro: according to the docs, if you pass `None` as the timeout, it will cache forever:
> The timeout argument is optional and defaults to the timeout argument of the appropriate backend in the [CACHES](https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-CACHES) setting (explained above). It’s the number of seconds the value should be stored in the cache. Passing in None for timeout will cache the value forever. A timeout of 0 won’t cache the value.

https://docs.djangoproject.com/en/4.2/topics/cache/#django.core.cache.cache.set

johnnagro (approved):
some suggestions but overall 👍

enterprise_catalog/apps/catalog/management/commands/tests/test_set_global_average_course_rating_value.py
alex-sheehan-edx: Forever sounds ok to me ¯\_(ツ)_/¯

PR #750, opened 2024-01-26
feat: drop include_exec_ed_2u_courses logic

johnnagro:
## Description

- in prep for dropping the field entirely
- we've already added query exclusions: https://2u-internal.atlassian.net/browse/ENT-7893
- we've already already remove it from the LMS ui: https://github.com/openedx/edx-enterprise/pull/2001


enterprise_catalog/apps/catalog/admin.py
```diff
@@ -76,6 +76,11 @@ def get_catalog(self, obj):
 @admin.register(CatalogQuery)
 class CatalogQueryAdmin(UnchangeableMixin):
     """ Admin configuration for the custom CatalogQuery model. """
+    fields = (
```
johnnagro, line 79: remove the field from the admin ui

iloveagent57 (approved):

PR #755, opened 2024-01-30
feat: smaller batch sizes when filtering over metadata

alex-sheehan-edx:
we were seeing k8s OOMKilled, this small batch size should help keep things under the limits (this table contains large, json metadata fields)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -349,7 +349,7 @@ def set_global_course_review_avg():
     total_number_reviews = 0.0
     course_only_filter = Q(content_type='course')
     # only courses have course reviews
-    for items_batch in batch_by_pk(ContentMetadata, extra_filter=course_only_filter):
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -349,7 +349,7 @@ def set_global_course_review_avg():
     total_number_reviews = 0.0
     course_only_filter = Q(content_type='course')
     # only courses have course reviews
-    for items_batch in batch_by_pk(ContentMetadata, extra_filter=course_only_filter):
+    for items_batch in batch_by_pk(ContentMetadata, batch_size=25, extra_filter=course_only_filter):
```
johnnagro, line 352: we were seeing k8s OOMKilled, this small batch size should help keep things under the limits (this table contains large, json metadata fields)

johnnagro (approved):

enterprise_catalog/apps/catalog/algolia_utils.py
alex-sheehan-edx: (adding to the PR descr)

PR #767, opened 2024-02-08
feat: ability to include unlisted courses in catalogs

johnnagro:
## Description

- introduces a new catalog query parameter `enterprise_force_include_aggregation_keys`
- when a catalog's metadata is downloaded, we include the courseware in addition to whatever the query provides
- discovery ignores keys it doesn't understand
- force-included keys will be added and transformed to appear published and available in the enterprise catalog service
- courses are marked with a magic key
- https://2u-internal.atlassian.net/browse/ENT-8212

## Todo

- [ ] we may need additional metadata transformers, such as enrollment start date

PR #775, opened 2024-02-14
feat: remove advertised run check for now

johnnagro:
## Description

- following on to https://github.com/openedx/enterprise-catalog/pull/767
- in stage we had unpublished courses with advertised runs
- advertised runs seemed safer thing to edit versus all runs
- in practice, we cannot safely advertise a run in prod for un-submitted courses


georgebabey:
👍 This looks good to unblock this, but we should circle back to see if we can have a published run without a published course in production.

georgebabey (approved):

alex-sheehan-edx (commented):
Looks good to me ✅

PR #779, opened 2024-02-22
feat: drop include_exec_ed_2u_courses field

johnnagro:
## Description

- dropping the db field https://2u-internal.atlassian.net/browse/ENT-8508
- we've already removed the logic: https://github.com/openedx/enterprise-catalog/pull/750
- we've already added query exclusions: https://2u-internal.atlassian.net/browse/ENT-7893
- we've already already remove it from the LMS ui: https://github.com/openedx/edx-enterprise/pull/2001


enterprise_catalog/apps/catalog/models.py
```diff
@@ -73,6 +73,7 @@ class CatalogQuery(TimeStampedModel):
         null=True,
         max_length=32,
         editable=False,
+        unique=True,
```
alex-sheehan-edx, line 76: This is a bigger change that maybe we need to notify people about?

alex-sheehan-edx (approved):
Looks fine, should we publicly state we're removing the unique constraint from queries?

enterprise_catalog/apps/catalog/models.py
johnnagro: this was unique before we added this column, we then made it unique together, since we're dropping the include column we can go back to just this column being unique

PR #783, opened 2024-02-26
feat: Added a new REST endpoint to get AI Curation.

saleem-latif:
__Jira Ticket:__ [ENT-8316](https://2u-internal.atlassian.net/browse/ENT-8316)

__Description:__
This PR introduces a new REST endpoint for AI Curations epic. This is a skeleton endpoint that receives, validates the requests and then triggers an async task. It also introduces a GET method to get the state/result of the async task.

## Post-review


enterprise_catalog/apps/ai_curation/api/v1/views.py
```diff
@@ -0,0 +1,50 @@
...
+        try:
+            curation_task = AICurationTask.objects.get(task_id=task_id)
+            return Response({
+                'status': curation_task.status,
+                'result': curation_task.result,
+            })
+        except AICurationTask.DoesNotExist:
+            return Response({'error': 'Task not found.'}, status=404)
+        except ValidationError:
+            return Response({'error': 'Invalid task_id.'}, status=400)
```
muhammad-ammar: Can we please use status constants at all the places?
```python
from rest_framework import status
status.HTTP_400_BAD_REQUEST
```

enterprise_catalog/apps/ai_curation/api/v1/views.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    View for AI Curation.
+    """
+    authentication_classes = []
+    permission_classes = []
+    throttle_classes = [AICurationThrottle]
+
+    def get(self, request):
+        """
+        Return a list of all users.
```
muhammad-ammar: Looks like we need to update this docstring?

enterprise_catalog/apps/ai_curation/api/v1/views.py
```diff
@@ -0,0 +1,50 @@
...
+                'result': curation_task.result,
+            })
+        except AICurationTask.DoesNotExist:
+            return Response({'error': 'Task not found.'}, status=404)
+        except ValidationError:
+            return Response({'error': 'Invalid task_id.'}, status=400)
+
+    def post(self, request):
+        """
+        Trigger the AI curation process.
```
muhammad-ammar, line 51: Can we please add some more details in the docstring?

enterprise_catalog/apps/ai_curation/models.py
```diff
@@ -0,0 +1,40 @@
...
+
+from django.core.serializers.json import DjangoJSONEncoder
+from django.db import models
+from django.utils import timezone
+from model_utils.models import TimeStampedModel
+
+from enterprise_catalog.apps.ai_curation.enums import AICurationStatus
+
+
+class AICurationTask(TimeStampedModel):
```
muhammad-ammar: I think instead of creating a new model we can use `TaskResult` model which is already integrated. I believe `TaskResult` has [everything](https://github.com/celery/django-celery-results/blob/cb9448976bc51865922319c91b6c2218859ba80d/django_celery_results/models.py#L18) that we need.

Example Usage: https://github.com/openedx/enterprise-catalog/blob/master/enterprise_catalog/apps/api/tasks.py
ADR: https://github.com/openedx/enterprise-catalog/blob/master/docs/decisions/0002-celery-task-restructuring.rst#use-a-taskresult-celery-backend

enterprise_catalog/apps/ai_curation/api/v1/views.py
```diff
@@ -0,0 +1,50 @@
...
+from enterprise_catalog.apps.ai_curation.models import AICurationTask
+from enterprise_catalog.apps.ai_curation.tasks import trigger_ai_curations
+
+
+class AICurationView(APIView):
+    """
+    View for AI Curation.
+    """
+    authentication_classes = []
+    permission_classes = []
```
muhammad-ammar, line 28: I know this API is public and we have throttling in place but what are our guidelines for public APIs if there are any?

enterprise_catalog/apps/ai_curation/api/v1/views.py
```diff
@@ -0,0 +1,50 @@
...
+    throttle_classes = [AICurationThrottle]
```
muhammad-ammar: What about having different throttling limits for `GET` and `POST`?

Also any idea how many requests we can expect from a single user per minutes? Any stats regarding current usage?

muhammad-ammar (approved):
LGTM

PR #786, opened 2024-03-01
feat: Filter academy tags with no indexed data. ENT-8406

irfanuddinahmad:
## Description

This PR filters the academy tags returned by the academy detail api to only those which have indexed data in Algolia.

## Ticket Link

[ENT-8406](https://2u-internal.atlassian.net/browse/ENT-8406)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/serializers.py
+++ b/enterprise_catalog/apps/api/v1/serializers.py
@@ -12,6 +12,9 @@ from enterprise_catalog.apps.api.v1.utils import (
     is_any_course_run_active,
     update_query_parameters,
 )
+from enterprise_catalog.apps.catalog.algolia_utils import (
+    get_initialized_algolia_client,
+)
 from enterprise_catalog.apps.catalog.constants import (
     COURSE,
     COURSE_RUN,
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/api/v1/serializers.py
```diff
@@ -402,13 +405,39 @@ def get_highlight_sets(self, obj):
...
+        tags_with_results = []
+        for tag in tags:
+            tag_title = tag['title']
+            search_options = {
+                'facetFilters': [f'academy_tags:{tag_title}',],
+                'attributesToRetrieve': ['aggregation_key',],
+                'hitsPerPage': 1,
+                'page': 0,
+            }
+            page = algolia_client.algolia_index.search('', search_options)
```
muhammad-ammar: I think it would be great if we can send a single call to algolia for all the tags. I think sending a call for each tag is expensive.

enterprise_catalog/apps/api/v1/views/academies.py
```diff
@@ -23,6 +23,12 @@ class AcademiesReadOnlyViewSet(viewsets.ReadOnlyModelViewSet):
     def request_action(self):
         return getattr(self, 'action', None)
 
+    def get_serializer_context(self):
+        context = super().get_serializer_context()
+        academy_uuid = str(self.kwargs['uuid']) if 'uuid' in self.kwargs else ''
```
muhammad-ammar, line 28: is `academy_uuid` set to `''` a valid case?

muhammad-ammar (approved):
LGTM

enterprise_catalog/apps/api/v1/views/academies.py
irfanuddinahmad: it is empty for the list view which uses tha same serializer

PR #797, opened 2024-03-13
feat: removed unnecessary batching for academy uuids and tags

irfanuddinahmad:
## Description

This PR removes the batching for academy uuids and tags. The academy uuids and tags are always going to be less than 15, hence, the batching is unnecessary since our batch size is 100. The undesirable side effect of batching is that an academy specific object id is added. Algolia query results, therefore, have missing enterprise uuids and catalog uuids since these have their own batches and unique object ids.

## Ticket Link

[ENT-8406](https://2u-internal.atlassian.net/browse/ENT-8406)

## Post-review


muhammad-ammar (approved):
LGTM

PR #800, opened 2024-03-18
feat: caching for contains-content-items and curation views

iloveagent57:
## Description

Implements a per-user cache on the curation read-only endpoints for the curation config and highlight sets.
A per-user cache is needed on these because we technically allow requests _without_ specifying a filterable identifier (either customer uuid or curation config uuid) in the URL, even though the learner portal _always_ supplies these in its requests: https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/33652b9dd38631d0e07573f6fc833b406fa64f08/src/components/search/content-highlights/data/service.js#L16
If we don't see much benefit here on the first attempt, we can probably enforce a more strict assumption that the caller of these endpoints always provides either  the customer uuid or a config uuid, which would let us change this to per-view caching (better overall performance).
Still, this should help some, because for highlights-enabled learner portals, we make one request to each of the curation config and search page on _each load_ of the LP dashboard _and_ the LP search page.

This also implements a per-view cache on `contains_content_items`. We can use a per-view cache here becuase the edx-rbac perms are built around a part of the URL path in that view (the uuid of the catalog).

## Ticket Link

https://openedx.atlassian.net/browse/ENT-8655

## Post-review


enterprise_catalog/apps/api/constants.py
```diff
@@ -2,6 +2,11 @@
 Constants for api app.
 """
 
+# Per-view cache timeouts
+CONTAINS_CONTENT_ITEMS_VIEW_CACHE_TIMEOUT_SECONDS = 60 * 30
+CURATION_CONFIG_READ_ONLY_VIEW_CACHE_TIMEOUT_SECONDS = 60 * 30
+HIGHLIGHT_SET_READ_ONLY_VIEW_CACHE_TIMEOUT_SECONDS = 60 * 30
```
pwnage101, line 8: This means 30 minutes, right?  Does that seem a bit high?

macdiesel (approved):

enterprise_catalog/apps/api/v1/tests/test_curation_views.py
```diff
@@ -134,6 +138,33 @@ def test_authorized_list(self, is_catalog_staff, is_role_assigned_via_jwt):
         assert len(curation_config_results) == 1
         assert curation_config_results[0]['uuid'] == str(self.curation_config_one.uuid)
 
+    @ddt.data(True, False)
```
pwnage101, line 141: I think this ddt doesn't test what we're hoping it tests.  Within one test run, the same exact URL is requested twice, instead of one with query params + one without query params.  Isn't the goal to test multiple requests in the same session with different query parameters?

enterprise_catalog/apps/api/constants.py
macdiesel: FWIW I too read this as 30 minutes.

enterprise_catalog/apps/api/v1/tests/test_curation_views.py
pwnage101: Apologies, this test is explicitly testing cache hits, so it is correct.  What I felt was missing was another test that ensures that different query parameters results in a cache miss.

iloveagent57: hmm, that's a good test, too. I was going for a test of what the learner portal frontend actually does (always sends a customer uuid). but we can add an additional test like you're suggesting.

macdiesel: I see an if statement to modify and add the param when ddt sets it to true?
```
if include_customer_param:
            url = url + f'?enterprise_customer={self.enterprise_uuid}'
```

pwnage101 (approved):
LGTM!

enterprise_catalog/apps/api/constants.py
iloveagent57: yup, 30 minutes, but this (IMO) is really slowly-changing data

PR #801, opened 2024-03-20
ENT-8400: Added transcript_languages in Algolia Search

mahamakifdar19:
## Description
This PR adds the `transcript_languages` facet in Algolia search fields

## Ticket Link
[ENT-8400](https://2u-internal.atlassian.net/browse/ENT-8400)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -93,6 +93,7 @@ ALGOLIA_FIELDS = [
     'reviews_count',
     'avg_course_rating',
     'course_bayesian_average',
+    'transcript_languages',
 ]
 
 # default configuration for the index
@@ -136,6 +137,7 @@ ALGOLIA_INDEX_SETTINGS = {
         'aggregation_key',
         'learning_type',
         'learning_type_v2',
+        'transcript_languages',
     ],
     'unretrievableAttributes': [
         'enterprise_catalog_uuids',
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -421,6 +423,25 @@ def get_course_language(course):
...
+    Gets the human-readable video transcript languages associated with the advertised course run. Used for
+    the "transcript_languages" facet in Algolia.
+
+    Arguments:
+        course (dict): a dict representing with course metadata
+
+    Returns:
+        list: a list of available human-readable video transcript languages parsed from a language code.
+    """
+    advertised_course_run = _get_course_run_by_uuid(course, course.get('advertised_course_run_uuid'))
```
irfanuddinahmad, line 437: Do we always have this key? ... better to put some guard code around it

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -421,6 +423,25 @@ def get_course_language(course):
...
+    if not advertised_course_run:
+        return None
```
irfanuddinahmad, line 439: Should this be None or []?

mahamakifdar19: @irfanuddinahmad ahh yes it should be None.

mahamakifdar19: @irfanuddinahmad None, we'll store the facet value only when an advertised course run is available.

enterprise_catalog/apps/catalog/algolia_utils.py
mahamakifdar19: @irfanuddinahmad I believe there's no need for a guard condition because if the key `advertised_course_run_uuid` is not present, `advertised_course_run` will be assigned `None` automatically since the function `_get_course_run_by_uuid` returns `None` and it handles that case.

irfanuddinahmad (approved):

jajjibhai008 (approved):

PR #803, opened 2024-03-21
fix: avoiding api errors from algolia when course metadata gets too large by omitting from indexing

alex-sheehan-edx:
We currently flatten algolia objects with the use of `ALGOLIA_UUID_BATCH_SIZE` as more customers are onboarded, but there is no safety guard against the indexer attempting to index content records that are larger than the max algolia object size. In such cases, the indexer blows up and no content is indexed, so as an immediate solution while we investigate the sudden growth of metadata size, we are simply omitting the record from indexing should it exceed a certain size.

According to the algolia docs: https://support.algolia.com/hc/en-us/articles/4406981897617-Is-there-a-size-limit-for-my-index-records 
record size is evaluated by dumping the object into json and read into memory- so I've done a similar method of calculating payload size. 

## Ticket Link

Link to the associated ticket
https://2u-internal.atlassian.net/browse/ENT-8667

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -2,6 +2,7 @@ import copy
 import functools
 import json
 import logging
+import sys
 import time
 from collections import defaultdict
 from datetime import timedelta
@@ -20,7 +21,9 @@ from enterprise_catalog.apps.api.constants import CourseMode
 from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.catalog.algolia_utils import (
     ALGOLIA_FIELDS,
+    ALGOLIA_JSON_METADATA_MAX_SIZE,
     ALGOLIA_UUID_BATCH_SIZE,
+    _algolia_object_from_product,
     _get_course_run_by_uuid,
     configure_algolia_index,
     create_algolia_objects,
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -654,31 +657,36 @@ def add_metadata_to_algolia_objects(
...
-    # enterprise customer uuids
-    customer_uuids = sorted(list(customer_uuids))
-    batched_metadata = _batched_metadata(
-        json_metadata,
-        customer_uuids,
-        'enterprise_customer_uuids',
-        '{}-customer-uuids-{}',
+
+    json_metadata_size = sys.getsizeof(
+        json.dumps(_algolia_object_from_product(json_metadata, algolia_fields=ALGOLIA_FIELDS)).strip(" "),
```
adamstankiewicz, line 662: nit: is it worth commenting a link to the Algolia docs on how they determine record size? https://support.algolia.com/hc/en-us/articles/4406981897617-Is-there-a-size-limit-for-my-index-records

enterprise_catalog/apps/api/tasks.py
```diff
@@ -654,31 +657,36 @@ def add_metadata_to_algolia_objects(
...
     )
-    _add_in_algolia_products_by_object_id(algolia_products_by_object_id, batched_metadata)
+    if json_metadata_size < ALGOLIA_JSON_METADATA_MAX_SIZE:
```
adamstankiewicz: [nit/suggestion] To avoid having the "happy path" code nested under a conditional, would it make sense to consider handling the reverse of this conditional with an early return?:

```python
if json_metadata_size > ALGOLIA_JSON_METADATA_MAX_SIZE:
  return

# everything else...
```

adamstankiewicz: [nit/suggestion] Is it worth introducing any logging around Algolia objects that weren't included for observability in the logs, if needed?

adamstankiewicz (approved):

enterprise_catalog/apps/api/tasks.py
alex-sheehan-edx: I put it in the PR desc but I guess it's worth it 👍

adamstankiewicz (approved):

PR #825, opened 2024-05-03
fix: Use percentile for filtering courses instead of tfidf values for uniform distribution

saleem-latif:
__Description:__
This PR updates filtering logic to use percentiles of `TF-IDF` files instead of raw values for a better filtering on uniformly distributed courses.

## Post-review


muhammad-ammar (approved):
LGTM

PR #829, opened 2024-05-08
feat: adding guard rails to content query association setter to prevent large negative deltas

alex-sheehan-edx:
https://2u-internal.atlassian.net/browse/ENT-8778
## Post-review

### DIFF
--- a/docs/decisions/0009-query-content-associations-guard-rails.rst
+++ b/docs/decisions/0009-query-content-associations-guard-rails.rst
@@ -1,0 +1,38 @@
+Query Content Associations Guard Rails
+======================================
+
+Status
+------
+
+Accepted
+
+Context
+-------
+We have on occasion seen moments where Discovery has reported an extreme number of incorrect content records associated
+with customer queries. Almost entirely in these situations we have seen the system right itself after some time and
+return the expected number of records. While ensuring for normal variance and content modification rates, we want to
+catch and prevent moments where Discovery or other sources of truth erroneously reports an extreme loss or gain of
+content associated with a query.
+
+Decision
+--------
+The ``associate_content_metadata_with_query`` method will now consider the size of the existing query/content
+association list as compared to the list of content to update to. Should the content association update surpass a
+configurable threshold, the update to the content associations will be blocked and the old state will be retained.
+
+To be considered for a threshold cutoff, the requester of ``associate_content_metadata_with_query`` must meet the
+following criteria:
+- The existing set of content associations must exceed a configurable cutoff
+- The query must have not been modified today
+- The change in number of content association records must exceed a configurable percentage, both in a positive and
+negative direction (ie the query loses or gains more than x% of its prior number of records)
+
+Consequences
+------------
+Content associations will not be updated in specific cases, causing stagnant data to be displayed to customers. The
+guardrail is also not loud/destructive in that it will log a warning about its action but not stop the process, just
+returning the existing content metadata association set. However, given the high levels of the thresholds set, it is
+intended that this only happens when the alternative of letting the content association update going through would be
+more disruptive to the customer's experience than to continue using old data. It is also intended that the check on the
+``modified`` field value will ensure there is a direct and easy way to allow a large content associations update to go
+through.
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -850,6 +850,54 @@ def create_content_metadata(metadata, catalog_query=None, dry_run=False):
     return metadata_list
 
 
+def _check_content_association_threshold(catalog_query, metadata_list):
+    """
+    Helper method to check a given catalog query's content metadata association set and compare it to a new set of
+    metadata records. Should the two sets of records differ in size beyond a configurable percentage value, and are
+    evaluated to be applicable, this method returns True, indicating the threshold of difference has been met.
+
+    Applicability for the threshold is defined as such:
+    - The existing set of content associations must exceed a configurable cutoff value
+    - The query must have not been modified today, meaning it's stale
+    - The change in number of content association records must exceed a configurable percentage, both in a positive and
+    negative direction (ie the query loses or gains more than x% of its prior number of records)
+    """
+    existing_relations_size = catalog_query.contentmetadata_set.count()
+    new_relations_size = len(metadata_list)
+    # To prevent false positives, this content association action stop gap will only apply to reasonably sized
+    # content sets
+    if existing_relations_size > settings.CATALOG_CONTENT_INCLUSION_GUARDRAIL_CONSIDERATION_FLOOR:
+        # If the catalog query hasn't been modified yet today, means there's no immediate reason for such a
+        # large change in content associations
+        if catalog_query.modified.date() < localized_utcnow().date():
+            # Check if the association of content results in a percentage change of
+            # `CATALOG_CONTENT_INCLUSION_GUARDRAIL_ALLOWABLE_DELTA` of content items from the query's content set.
+            percent_change = abs((new_relations_size - existing_relations_size) / existing_relations_size)
+            if percent_change > settings.CATALOG_CONTENT_INCLUSION_GUARDRAIL_ALLOWABLE_DELTA:
+                LOGGER.warning(
+                    "[CONTENT_DELTA_WARNING] associate_content_metadata_with_query is requested to set query: "
+                    "%s to a content metadata set of length of %s when it previous had a content metadata set length "
+                    "of:%s. The current threshold cutoff is a delta of: %s content remaining. The update has been "
+                    "prevented.",
+                    catalog_query,
+                    new_relations_size,
+                    existing_relations_size,
+                    settings.CATALOG_CONTENT_INCLUSION_GUARDRAIL_ALLOWABLE_DELTA,
+                )
+                return True
+            elif percent_change > settings.CATALOG_CONTENT_ASSOCIATIONS_CONTENT_DELTA_WARNING_THRESHOLD:
+                LOGGER.warning(
+                    "[CONTENT_DELTA_WARNING] associate_content_metadata_with_query hit the warning threshold: %s "
+                    "while setting query: %s to a content metadata set of length of %s when it previous had a content "
+                    "metadata set length of:%s.",
+                    settings.CATALOG_CONTENT_ASSOCIATIONS_CONTENT_DELTA_WARNING_THRESHOLD,
+                    catalog_query,
+                    new_relations_size,
+                    existing_relations_size,
+                )
+    return False
+
+
 def associate_content_metadata_with_query(metadata, catalog_query, dry_run=False):
     """
     Creates or updates a ContentMetadata object for each entry in `metadata`,
@@ -864,7 +912,11 @@ def associate_content_metadata_with_query(metadata, catalog_query, dry_run=False
         list: The list of content_keys for the metadata associated with the query.
     """
     metadata_list = create_content_metadata(metadata, catalog_query, dry_run)
-
+    # Stop gap if the new metadata list is extremely different from the current one
+    if _check_content_association_threshold(catalog_query, metadata_list):
+        return list(content_key for content_key in catalog_query.contentmetadata_set.values_list(
+            'content_key', flat=True,
+        ))
     # Setting `clear=True` will remove all prior relationships between
     # the CatalogQuery's associated ContentMetadata objects
     # before setting all new relationships from `metadata_list`.
--- a/enterprise_catalog/apps/catalog/tests/test_models.py
+++ b/enterprise_catalog/apps/catalog/tests/test_models.py
@@ -3,6 +3,7 @@
 import json
 from collections import OrderedDict
 from contextlib import contextmanager
+from datetime import timedelta
 from unittest import mock
 from uuid import uuid4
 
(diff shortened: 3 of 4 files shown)


enterprise_catalog/settings/base.py
```diff
@@ -447,3 +447,13 @@
 OPENAI_API_KEY = 'I am key'
 
 ############################################## AI CURATION ##############################################
+
+CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF = os.environ.get(
```
marlonkeating: Can we get documentation for these variables?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -848,6 +848,31 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
+        if catalog_query.modified.date() < localized_utcnow().date():
+            # Check if the association of content results in a percentage loss of
+            # `CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF` (90% as of May 8th 2024) of content items from the
+            # query's content set.
+            content_delta = new_relations_size / existing_relations_size
+            if content_delta < settings.CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF:
+                LOGGER.warning(
+                    f"associate_content_metadata_with_query is requested to set query: {catalog_query} to a "
+                    f"content metadata set of length of {new_relations_size} when it previous had a content "
+                    f"metadata set length of:{existing_relations_size}. The updated is prevented"
```
marlonkeating:
```suggestion
                    f"metadata set length of:{existing_relations_size}. The update is prevented"
```

enterprise_catalog/apps/catalog/models.py
```diff
@@ -848,6 +848,31 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
+                )
+                return [metadata.content_key for metadata in catalog_query.contentmetadata_set.all()]
+            elif content_delta < settings.CATALOG_CONTENT_ASSOCIATIONS_CONTENT_DELTA_WARNING_THRESHOLD:
+                LOGGER.warning(
+                    "associate_content_metadata_with_query hit the warning threshold while setting query: "
```
marlonkeating: Spotting these messages in the logs might be easier if we prefaced the log lines here
```suggestion
                    "[CONTENT_DELTA_WARNING] associate_content_metadata_with_query hit the warning threshold while setting query: "
```

enterprise_catalog/apps/catalog/models.py
marlonkeating: Could we log the percentage diff and the threshold we're working with?

marlonkeating (changes requested):

enterprise_catalog/apps/catalog/models.py
alex-sheehan-edx: the percent dif is just new / old. But I'm fine with logging the threshold.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -848,6 +848,34 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
+        # large change in content associations
+        if catalog_query.modified.date() < localized_utcnow().date():
+            # Check if the association of content results in a percentage loss of
+            # `CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF` (80% as of May 8th 2024) of content items from the
+            # query's content set.
+            content_delta = new_relations_size / existing_relations_size
+            if content_delta < settings.CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF:
+                LOGGER.warning(
+                    "[CONTENT_DELTA_WARNING] associate_content_metadata_with_query is requested to set query: "
+                    f"{catalog_query} to a content metadata set of length of {new_relations_size} when it previous "
```
marlonkeating:
```suggestion
                    f"{catalog_query} to a content metadata set of length of {new_relations_size} when it previously "
```

docs/decisions/0009-query-content-associations-guard-rails.rst
```diff
@@ -0,0 +1,37 @@
...
+To be considered for a threshold cutoff, the requester of `associate_content_metadata_with_query` must meet the
+following criteria:
+- The existing set of content associations must exceed a configurable cutoff
+- The query must have not been modified today
+- The change in number of content association records must exceed a configurable percentage, both in a positive and
+negative direction (ie the query loses or gains more than x% of its prior number of records)
+
+Consequences
+------------
+Content associations will not be updated in specific cases, causes stagnant data being displayed to customers. The
```
marlonkeating:
```suggestion
Content associations will not be updated in specific cases, causing stagnant data to be displayed to customers. The
```

docs/decisions/0009-query-content-associations-guard-rails.rst
```diff
@@ -0,0 +1,37 @@
...
+guardrail is also not loud/destructive in that it will log a warning about its action but not stop the process, just
+returning the existing content metadata association set. However, given the high levels of the thresholds set, it is
+intended that this only happens when the alternative of letting the content association update going through would be
+more disruptive to the customer's experience than the to continue using old data. It is also intended that the check on
```
marlonkeating:
```suggestion
more disruptive to the customer's experience than to continue using old data. It is also intended that the check on
```

marlonkeating (approved):

docs/decisions/0009-query-content-associations-guard-rails.rst
```diff
@@ -0,0 +1,37 @@
...
+-------
+We have on occasion seen moments where Discovery has reported an extreme number of incorrect content records associated
+with customer queries. Almost entirely in these situations we have seen the system right itself after some time and
+return the expected number of records. While ensuring for normal variance and content modification rates, we want to
+catch and prevent moments where Discovery or other sources of truth erroneously reports an extreme loss or gain of
+content associated with a query.
+
+Decision
+--------
+The `associate_content_metadata_with_query` method will now consider the size of the existing query/content association
```
iloveagent57: in RST, you need double back-ticks to format something as "code". That is:
```
``associate_content_metadata_with-query``
```

docs/decisions/0009-query-content-associations-guard-rails.rst
```diff
@@ -0,0 +1,37 @@
...
+negative direction (ie the query loses or gains more than x% of its prior number of records)
+
+Consequences
+------------
+Content associations will not be updated in specific cases, causing stagnant data to be displayed to customers. The
+guardrail is also not loud/destructive in that it will log a warning about its action but not stop the process, just
+returning the existing content metadata association set. However, given the high levels of the thresholds set, it is
+intended that this only happens when the alternative of letting the content association update going through would be
+more disruptive to the customer's experience than to continue using old data. It is also intended that the check on
+modified at time will ensure there is a direct and easy way to allow a large content associations update to go through.
```
iloveagent57: nit: format as "``modified_at``" here to denote it as a field.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -848,6 +848,37 @@ def associate_content_metadata_with_query(metadata, catalog_query):
     """
     metadata_list = create_content_metadata(metadata, catalog_query)
 
+    existing_relations_size = len(catalog_query.contentmetadata_set.all())
```
iloveagent57: Hard blocker: switch to
```
catalog_query.contentmetadata_set.count()
```
so we don't load the entire CM set into memory.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -848,6 +848,37 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
+    new_relations_size = len(metadata_list)
+    # To prevent false positives, this content association action stop gap will only apply to reasonably sized
+    # content sets
+    if existing_relations_size > settings.CATALOG_CONTENT_ASSOCIATIONS_CONTENT_BASE_SIZE_CUTOFF:
```
iloveagent57: nit: maybe `CATALOG_CONTENT_INCLUSION_GUARDRAIL_CONSIDERATION_FLOOR`
and `CATALOG_CONTENT_INCLUSION_GUARDRAIL_ALLOWABLE_DELTA` for these setting names?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -848,6 +848,37 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
+        # If the catalog query hasn't been modified yet today, means there's no immediate reason for such a
+        # large change in content associations
+        if catalog_query.modified.date() < localized_utcnow().date():
+            # Check if the association of content results in a percentage loss of
+            # `CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF` (80% as of May 8th 2024) of content items from the
```
iloveagent57: nit: adding in the percentage as of some date is actually misleading - the reader should be forced to go lookup the current value, rather than being presented with an historical guess.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -848,6 +848,37 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
+            # content_delta = new_relations_size / existing_relations_size
+            if percent_change > settings.CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF:
+                LOGGER.warning(
+                    "[CONTENT_DELTA_WARNING] associate_content_metadata_with_query is requested to set query: "
+                    f"{catalog_query} to a content metadata set of length of {new_relations_size} when it previous "
+                    f"had a content metadata set length of:{existing_relations_size}. The current threshold cutoff is "
+                    f"a delta of: {settings.CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF} content remaining. The "
+                    "update has been prevented."
+                )
+                return [metadata.content_key for metadata in catalog_query.contentmetadata_set.all()]
```
iloveagent57: Use `values_list('content_key', flat=True)` here since you only need the keys, which will be a much smaller result set.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -848,6 +848,37 @@ def associate_content_metadata_with_query(metadata, catalog_query):
...
+        if catalog_query.modified.date() < localized_utcnow().date():
+            # Check if the association of content results in a percentage loss of
+            # `CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF` (80% as of May 8th 2024) of content items from the
+            # query's content set.
+
+            percent_change = abs((new_relations_size - existing_relations_size) / existing_relations_size)
+
+            # content_delta = new_relations_size / existing_relations_size
+            if percent_change > settings.CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF:
+                LOGGER.warning(
```
iloveagent57, line 877: nit: prefer to use `%s` in log strings instead of f-strings: https://docs.python.org/3/howto/logging.html#optimization

enterprise_catalog/apps/catalog/tests/test_models.py
```diff
@@ -392,3 +394,75 @@ def test_enrollment_url_exec_ed(
         else:
             assert 'happy-little-sku' in actual_enrollment_url
             assert 'proxy-login' in actual_enrollment_url
+
+    @override_settings(DISCOVERY_CATALOG_QUERY_CACHE_TIMEOUT=0)
+    @mock.patch('enterprise_catalog.apps.api_client.discovery.DiscoveryApiClient')
+    def test_associate_content_metadata_with_query_guardrails(self, mock_client):
```
iloveagent57, line 507: great job on a thorough test!

enterprise_catalog/apps/catalog/models.py
iloveagent57: one more suggestion: move everything in this if-block to a helper function - it's a lot of code to read that doesn't have much bearing on the rest of the main logic.

enterprise_catalog/settings/base.py
```diff
@@ -447,3 +447,18 @@
 OPENAI_API_KEY = 'I am key'
 
 ############################################## AI CURATION ##############################################
+
+# Decimal percentage cutoff value for queries' content metadata association setting. If len(new set) / len(old set) is
+# less than the setting value, the update may be prevented
+CATALOG_CONTENT_ASSOCIATION_CONTENT_DELTA_CUTOFF = os.environ.get(
```
iloveagent57: nit: you don't need to use `os.environ` as part of your default anymore. you can do this in `base.py`:
```python
NEW_SETTING = 'my-default-value'
```
and over on the production/stage yaml config side of things, override the default in the environment with
```
NEW_SETTING: "my-production-value"
```

iloveagent57 (changes requested):
This is a good idea, and the ADR is very useful, thank you!

docs/decisions/0009-query-content-associations-guard-rails.rst
alex-sheehan-edx: technically the field on the table is `modified`

alex-sheehan-edx: let me use back-ticks and clean up the language a bit

enterprise_catalog/apps/catalog/models.py
alex-sheehan-edx: wow, I need to switch up my patterns 😅

iloveagent57 (approved):

PR #832, opened 2024-05-15
feat: Added ADR for video indexing

irfanuddinahmad:
## Post-review

### DIFF
--- a/docs/decisions/0010-indexing-videos-for-enterprise.rst
+++ b/docs/decisions/0010-indexing-videos-for-enterprise.rst
@@ -1,0 +1,35 @@
+Indexing videos for search and driving enrollments
+==================================================
+
+Status
+------
+
+Draft
+
+
+Context
+-------
+Microlearning, broadly defined, are short form educational experiences (<30 minutes) that are designed to deliver against a specific learning objective while minimizing learner fatigue. The scope of this decision record is a microlearning experience that focuses on videos. The videos will be utilized to market the associated course to prospective learners and help drive enrollments, which will eventually result in an increase of revenue.
+
+
+Solution Architecture
+---------------------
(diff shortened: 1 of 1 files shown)


nrobertson1992:
Regarding the "alternatives considered" section -
1) how much additional effort is involved with the solution being through edx-discovery search?
2) if in the future B2C wanted to adopt this feature, would they be able to easily extend our work into a b2c video use case?

Just thinking a couple steps ahead. B2C has no plans right now to do this. But want to pause and think through whether or not this architecture creates a path for B2C to pursue this if they wanted to.

irfanuddinahmad:
> Regarding the "alternatives considered" section -
> 
> 1. how much additional effort is involved with the solution being through edx-discovery search?
> 2. if in the future B2C wanted to adopt this feature, would they be able to easily extend our work into a b2c video use case?
> 
> Just thinking a couple steps ahead. B2C has no plans right now to do this. But want to pause and think through whether or not this architecture creates a path for B2C to pursue this if they wanted to.

@nrobertson1992 this work does not hinder any video indexing work planned by B2C in Discovery. If we go via Discovery, our effort would effectively double since we would now need to create two indexes and sync them.

docs/decisions/0010-indexing-videos-for-enterprise.rst
```diff
@@ -0,0 +1,35 @@
...
+
+
+Decision
+--------
+Video catalog is an enterprise catalog and should reside within the Enterprise Catalog service. Video and course metadata relationships can be easily managed while existing indexing routines in enterprise-catalog repository can be reused.
```
iloveagent57, line 22: Is the idea that there is _one_ catalog record that references all of the video content? Or will the content be included in a subset of customer catalogs as a new content type?

irfanuddinahmad: @iloveagent57 the catalog relationship will be picked up from the associated course of any video. While indexing a course in Algolia, its associated videos will be fetched and also indexed with the same catalog/query/customer uuids as those of the course

PR #835, opened 2024-05-22
feat: add default availability filter in catalog algolia query

jajjibhai008:
**Description**

***Problem***
When we send a query to our Xpert from the frontend-app-enterprise-public-catalog, some of the default filters are applied. Consequently, we do not obtain the same count on the XpertStat card as the results shown below.

***Solution***
We have applied an `availability` filter in the backend Algolia query to synchronize the result counts on the frontend-app-enterprise-public-catalog.

## Post-review


muhammad-ammar (approved):
LGTM

saleem-latif (approved):

PR #858, opened 2024-06-25
feat: add get_content_filter_hash endpoint

marlonkeating:
[Jira Ticket](https://2u-internal.atlassian.net/browse/ENT-8661)

Following up on the changes in https://github.com/openedx/enterprise-catalog/pull/837, this adds a method for retrieving content query hashes for query text, which can then be passed to `get_query_by_hash`.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -2589,6 +2589,25 @@ class CatalogQueryViewTests(APITestMixin):
         response_json = response.json()
         assert response_json == ['You must provide at least one of the following query parameters: hash.']
 
+    def test_get_content_filter_hash(self):
+        """
+        Test that get content filter hash returns md5 hash of query
+        """
+        url = reverse('api:v1:get-content-filter-hash')
+        test_query = json.dumps({"content_type": ["political", "unit", "market"]})
+        response = self.client.generic('GET', url, content_type='application/json', data=test_query)
+        assert response.json() == '35584b583415a5bd4e51cc70d898a0eb'  # pylint: disable=no-member
+
+    def test_get_content_filter_hash_bad_query(self):
+        """
+        Test that get content filter hash returns md5 hash of query
+        """
+        url = reverse('api:v1:get-content-filter-hash')
+        test_query = 'bad query'
+        response = self.client.generic('GET', url, content_type='application/json', data=test_query)
+        err_detail = "Failed to parse catalog query: JSON parse error - Expecting value: line 1 column 1 (char 0)"
+        assert response.json() == {"detail": err_detail}  # pylint: disable=no-member
+
     def test_catalog_query_retrieve(self):
         """
         Test that the Catalog Query viewset supports retrieving individual queries
--- a/enterprise_catalog/apps/api/v1/views/catalog_query.py
+++ b/enterprise_catalog/apps/api/v1/views/catalog_query.py
@@ -3,7 +3,7 @@ from django.utils.functional import cached_property
 from edx_rbac.mixins import PermissionRequiredForListingMixin
 from rest_framework import viewsets
 from rest_framework.decorators import action
-from rest_framework.exceptions import NotFound
+from rest_framework.exceptions import NotFound, ParseError
 from rest_framework.renderers import JSONRenderer
 from rest_framework.response import Response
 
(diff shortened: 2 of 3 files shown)


enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -2589,6 +2589,15 @@ def test_get_query_by_hash_requires_hash(self):
         response_json = response.json()
         assert response_json == ['You must provide at least one of the following query parameters: hash.']
 
+    def test_get_content_filter_hash(self):
+        """
+        Test that get content filter hash returns md5 hash of query
+        """
+        url = reverse('api:v1:get-content-filter-hash')
```
pmakwana93, line 2596: why do we need to reverse the URL?

enterprise_catalog/apps/api/v1/views/catalog_query.py
```diff
@@ -106,3 +109,11 @@ def get_query_by_hash(self, request, **kwargs):
             raise NotFound('Catalog query not found.') from exc
         serialized_data = self.serializer_class(query)
         return Response(serialized_data.data)
+
+    @action(detail=True, methods=['get'])
+    def get_content_filter_hash(self, request, **kwargs):
+        """
+        Get md5 hash of a catalog query
+        """
+        content_filter_hash = get_content_filter_hash(json.loads(request.body))
```
pmakwana93: Are we guaranteed to get a valid response? What about failure scenarios where the response is invalid or the API call times out?

enterprise_catalog/apps/api/v1/tests/test_views.py
marlonkeating: Not strictly needed, but if we hypothetically need to change the url scheme in the future, we can more easily track references this way rather than having to hunt down url snippets.

enterprise_catalog/apps/api/v1/views/catalog_query.py
marlonkeating: Good call-out!  Added error handling for invalid json input.

enterprise_catalog/apps/api/v1/tests/test_views.py
iloveagent57: Using `reverse()` to turn a view name into a full URL is a pretty standard practice for django unit tests.
https://stackoverflow.com/a/11241936 has a good overview of what/why `reverse` is.

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -2589,6 +2589,24 @@ def test_get_query_by_hash_requires_hash(self):
...
+        test_query = '{ "content_type": [ "political", "unit", "market" ] }'
+        response = self.client.generic('GET', url, data=test_query)
```
iloveagent57: nit: you should be able to write this as `self.client.get(url, data=test_query)`

enterprise_catalog/apps/api/v1/views/catalog_query.py
```diff
@@ -106,3 +109,15 @@ def get_query_by_hash(self, request, **kwargs):
             raise NotFound('Catalog query not found.') from exc
         serialized_data = self.serializer_class(query)
         return Response(serialized_data.data)
+
+    @action(detail=True, methods=['get'])
+    def get_content_filter_hash(self, request, **kwargs):
+        """
+        Get md5 hash of a catalog query
+        """
+        catalog_query = request.body.decode()
```
iloveagent57: You should be able to use `request.data` to get this as de-serialized JSON: https://www.django-rest-framework.org/api-guide/requests/#request-parsing

enterprise_catalog/apps/api/v1/views/catalog_query.py
```diff
@@ -106,3 +109,15 @@ def get_query_by_hash(self, request, **kwargs):
...
+        try:
+            content_filter_hash = get_content_filter_hash(json.loads(catalog_query))
```
iloveagent57: (and if you use `request.data`, you won't explicitly have to `json.loads()` here)

enterprise_catalog/apps/api/v1/views/catalog_query.py
iloveagent57: Also, is this really a `catalog_query`, or is it the `content_filter` value of a catalog query?

iloveagent57 (commented):
This is a good idea. Handful of nits/suggestions then it should be good to go!

enterprise_catalog/apps/api/v1/tests/test_views.py
marlonkeating: You would think, but it looks like `client.get` doesn't support the parameter, throwing the error message `ValueError: not enough values to unpack (expected 2, got 1)`

iloveagent57 (approved):

PR #861, opened 2024-06-28
temp: debug MIT Micromasters exclusion

pmakwana93:
MIT Micromaster Programs are not showing up in Catalog Search due to not getting tagged with the correct metadata in Algolia.

- This change adds debug logging to enterprise-catalog's Algolia metadata creation logic after consulting with @iloveagent57 
- This will help us identify why MIT Micromasters program courses are not getting tagged with enterprise_catalog_query_titles = 'a la carte'
catalog_queries_by_key  field is used to derive the setting of the enterprise_catalog_query_titles field in Algolia. Reference is [here](https://github.com/openedx/enterprise-catalog/blob/07d4a048cf608247bb0f8bde4dc4feb6d27cc6ef/enterprise_catalog/apps/api/tasks.py#L541)

Tested the change via unit tests.

JIRA: https://2u-internal.atlassian.net/browse/ENT-8812

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -873,6 +873,18 @@ def _get_algolia_products_for_batch(
                             academy_tags_by_key[content_key].add(str(tag.title))
                             academy_tags_by_catalog_uuid[str(catalog.uuid)].add(str(tag.title))
 
+        # adding temporary debug logging for MIT's Finance Micromasters Program
+        if content_key in (
+            "MITx+15.415.2x",
+            "MITx+15.435x",
+            "MITx+15.516x",
+            "MITx+15.455x",
+            "MITx+15.415.1x",
+            "MITx+FIN.CFx"
+        ):
+            logger.info('Debugging content key = {} ...'.format(content_key))
+            logger.info('catalog_queries_by_key: {}'.format(catalog_queries_by_key[content_key]))
+
     # Second pass.  This time the goal is to capture indirect relationships on programs:
     #  * For each program:
     #    - Absorb all UUIDs associated with every associated course.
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -873,6 +873,18 @@ def _get_algolia_products_for_batch(
...
                             academy_tags_by_catalog_uuid[str(catalog.uuid)].add(str(tag.title))
 
+        # adding temporary debug logging for MIT's Finance Micromasters Program
+        if content_key in (
+            "MITx+15.415.2x",
+            "MITx+15.435x",
+            "MITx+15.516x",
+            "MITx+15.455x",
+            "MITx+15.415.1x",
+            "MITx+FIN.CFx"
```
katrinan029, line 883: question (non-blocking): out of curiosity, why are we only checking for these content keys? are they the only MIT micromaster courses?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -873,6 +873,18 @@ def _get_algolia_products_for_batch(
...
+        ):
+            logger.info('Debugging content key = {} ...'.format(content_key))
```
katrinan029, line 885: suggestion (non-blocking): i'd recommend adding text that could be more easily searched (I.e. `Debugging MIT Micromasters content key = {} ...'.format(content_key)` and `Debugging MIT Micromasters catalog_queries_by_key: {}'.format(catalog_queries_by_key[content_key])`.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -906,6 +918,18 @@ def _get_algolia_products_for_batch(
...
+        # adding temporary logging for MIT's Finance Micromasters Program
+        if program_content_key in (
+            "MITx+15.415.2x",
+            "MITx+15.435x",
+            "MITx+15.516x",
+            "MITx+15.455x",
+            "MITx+15.415.1x",
+            "MITx+FIN.CFx"
+        ):
+            logger.info('Pass 2 - Debugging program_content_key = {} ...'.format(program_content_key))
```
katrinan029, line 930: same comments as above

katrinan029 (approved):

enterprise_catalog/apps/api/tasks.py
pmakwana93: I picked courses keys from one specific MIT Micromaster Program (Finance) to debug

enterprise_catalog/apps/api/tasks.py
pmakwana93: Thanks, will keep that in mind for next time

enterprise_catalog/apps/api/tasks.py
pmakwana93: Thanks

PR #872, opened 2024-07-10
fix: adding page size param to highlights

kiram15:
Adding customization to the page_size so that we can increase it to 12 (to fetch all in one page), since 12 is the limit of highlights one customer can add. 

[Jira link 
](https://2u-internal.atlassian.net/browse/ENT-9210)
## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views/curation/highlights.py
+++ b/enterprise_catalog/apps/api/v1/views/curation/highlights.py
@@ -210,6 +210,7 @@ class HighlightSetBaseViewSet(PermissionRequiredForListingMixin, BaseViewSet):
     renderer_classes = [JSONRenderer, XMLRenderer]
     serializer_class = HighlightSetSerializer
     lookup_field = 'uuid'
+    pagination_class = PageNumberWithSizePagination
 
     # Fields required for controlling access in the `list()` action
     list_lookup_field = 'enterprise_curation__enterprise_uuid'
--- a/enterprise_catalog/apps/api/v1/views/curation/highlights.py
+++ b/enterprise_catalog/apps/api/v1/views/curation/highlights.py
@@ -22,9 +22,6 @@ from enterprise_catalog.apps.api.v1.constants import SegmentEvents
 from enterprise_catalog.apps.api.v1.event_utils import (
     track_highlight_set_changes,
 )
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/api/v1/views/curation/highlights.py
```diff
@@ -210,6 +210,7 @@ class HighlightSetBaseViewSet(PermissionRequiredForListingMixin, BaseViewSet):
     renderer_classes = [JSONRenderer, XMLRenderer]
     serializer_class = HighlightSetSerializer
     lookup_field = 'uuid'
+    pagination_class = PageNumberWithSizePagination
```
adamstankiewicz: [inform/suggestion] While `PageNumberWithSizePagination` is the `pagination_class` mostly used throughout this app, there is an opportunity to rely on `DefaultPagination` from `edx-drf-extensions` ([source](https://github.com/openedx/edx-drf-extensions/blob/c53f5f69faf9e77673c48876106f8757eeb0f23e/edx_rest_framework_extensions/paginators.py#L9)) which is a more standard, out-of-the-box `pagination_class` within Open edX that includes the same `page_size` as `PageNumberWithSizePagination ` but with a more consistent serialized pagination response compared to the default `PageNumberPagination` or the custom `PageNumberWithSizePagination` class.

```python
from edx_rest_framework_extensions.paginators import DefaultPagination

pagination_class = DefaultPagination
```

Relying on `DefaultPagination` here and throughout the IDA would effectively eliminate the need for the custom `PageNumberWithSizePagination` paginator class altogether.

adamstankiewicz: Related, I also might wonder if we should start ensuring the default pagination class in the base `REST_FRAMEWORK` settings ([source](https://github.com/openedx/enterprise-catalog/blob/3bf77c50c0e622f10fb854d78cd27a98b3611089/enterprise_catalog/settings/base.py#L140-L141)) for our Enterprise IDAs is set to `DefaultPagination` so all APIs use it by default without needing to explicitly opt-in to it.

```python
# Django Rest Framework
REST_FRAMEWORK = {
    'DEFAULT_PAGINATION_CLASS': 'edx_rest_framework_extensions.paginators.DefaultPagination',
}
```

marlonkeating (approved):

PR #873, opened 2024-07-12
feat: Added management command for fetching video metadata

irfanuddinahmad:
## Post-review


enterprise_catalog/apps/video_catalog/management/commands/fetch_video_metadata.py
```diff
@@ -0,0 +1,51 @@
...
+        Fetch video content metadata from LMS.
+        """
+        shortlisted_videos = VideoShortlist.objects.filter(is_processed=False)
+        if options.get('force', False):
+            shortlisted_videos = VideoShortlist.objects.all()
+        for shorlisted_video in shortlisted_videos:
+            try:
+                fetch_video(shorlisted_video)
+                shorlisted_video.is_processed = True
+                shorlisted_video.save()
```
saleem-latif: Instead of making an update call in every iteration, it would be more efficient to store a list of video ids and run a single update query in the end.

```
VideoShortlist.objects.filter(is_processed=False, id__in=video_ids).update(is_processed=True)
```

saleem-latif (changes requested):

enterprise_catalog/apps/video_catalog/management/commands/fetch_video_metadata.py
irfanuddinahmad: done

saleem-latif (approved):

PR #886, opened 2024-07-23
feat: include parent key in serialized content metadata responses.

iloveagent57:
ENT-8875 | enterprise-access, a consumer of the content metadata API, will soon want to rely on the presence of the `parent_content_key` field for use in the `content_assignments` feature. This change ensures said field is present in responses to the `get_content_metadata` endpoint.

## Post-review

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,6 +90,7 @@ docs/_build/
 
 # emacs
 *~
+.projectile
 
 # transifex exe
 tx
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -90,6 +90,7 @@ docs/_build/
 
 # emacs
 *~
+.projectile
```
adamstankiewicz, line 93: 🚀

pwnage101 (approved):

adamstankiewicz (approved):
LGTM!

PR #889, opened 2024-07-24
fix: Added partners field to video content for indexing

irfanuddinahmad:
## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -22,9 +22,15 @@ from enterprise_catalog.apps.catalog.constants import (
     LEARNER_PATHWAY,
     PROGRAM,
     PROGRAM_TYPES_MAP,
+    VIDEO,
 )
 from enterprise_catalog.apps.catalog.models import ContentMetadata
 from enterprise_catalog.apps.catalog.utils import batch_by_pk, localized_utcnow
+from enterprise_catalog.apps.video_catalog.models import (
+    Video,
+    VideoSkill,
+    VideoTranscriptSummary,
+)
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


macdiesel (approved):

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1343,6 +1497,17 @@ def _algolia_object_from_product(product, algolia_fields):
...
+    elif searchable_product.get('content_type') == VIDEO:
+        searchable_product.update({
+            'partners': get_video_partners(searchable_product),
+            'transcript_summary': get_transcript_summary(searchable_product),
+            'video_skills': get_video_skills(searchable_product),
+            'course_run_key': get_video_course_run_key(searchable_product),
+            'org': get_video_org(searchable_product),
+            'logo_image_urls': get_video_logo_image_urls(searchable_product),
+            'image_url': get_video_image_url(searchable_product),
+            'duration': get_video_duration(searchable_product),
```
iloveagent57, line 1509: Probably not a huge deal, but there's going to be one DB query per key in this, because each function is doing a `Video.objects.get()`.  You could instead do that query for the video once before running the functions, and then pass in the result to each function.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1343,6 +1497,17 @@ def _algolia_object_from_product(product, algolia_fields):
...
             'learning_type_v2': get_learning_type_v2(searchable_product),
         })
+    elif searchable_product.get('content_type') == VIDEO:
+        searchable_product.update({
+            'partners': get_video_partners(searchable_product),
+            'transcript_summary': get_transcript_summary(searchable_product),
+            'video_skills': get_video_skills(searchable_product),
+            'course_run_key': get_video_course_run_key(searchable_product),
+            'org': get_video_org(searchable_product),
+            'logo_image_urls': get_video_logo_image_urls(searchable_product),
```
adamstankiewicz, line 1507: [clarification] Now that `partners` is added (similar to other content types), what is the intent for these `org` and `logo_image_urls` fields? These 2 fields are now essentially duplicative with `partners`. Is there a plan to migrate the Learner Portal's consumption of `org` and `logo_image_urls` for video search results ([source](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/2e013ab60e555433a6422c28d3e369b093073287/src/components/search/SearchVideoCard.jsx#L58-L64)) to `partners` similar to other search result components? That way, the indexed data in Algolia could likely remove these fields to keep the indexed record size as small as possible.

PR #890, opened 2024-07-24
fix: ContentMetadata.bulk_update() changes the modified value

iloveagent57:
Prior to this change, the `ContentMetadata.modified` field was mostly meaningless, because `TimeStampedModel` only sets `modified` during calls to `save()`, but we mostly to updates of CM records via `ContentMetadata.objects.bulk_update()`.
## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -493,6 +493,26 @@ class EnterpriseCatalog(TimeStampedModel):
         return xapi_activity_id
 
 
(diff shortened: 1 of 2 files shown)


pwnage101 (approved):
non-blocking question: Before this PR the modified field was completely meaningless, but now it's eagerly updated for a record even if none of the other requested bulk fields have changed for that particular record; is there a way to efficiently set modified only if the record has changed?

iloveagent57:
Presumably the caller would not be calling `bulk_update` on a record(s) unless the record(s) have changed.

PR #899, opened 2024-08-06
refactor: use serializer to compute normalized_metadata

iloveagent57:
Introduces an explicit serializer, `NormalizedContentMetadataSerializer`, for populating the `normalized_metadata` key inside of `json_metadata`.  I'm hoping for two things:
1. To expand the number of fields in this serializer, and to eventually utilize it in emiting openedx events.
2. To illustrate a pattern for future adoption of serializer-based data transformation.

## Post-review

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
         architecture: x64
     - name: Start container
       run: |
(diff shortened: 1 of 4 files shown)


pwnage101 (approved):
LGTM!

PR #904, opened 2024-08-08
fix: normalized serializer handles null advertised run

iloveagent57:
## Post-review


adamstankiewicz (approved):
LGTM!

PR #909, opened 2024-08-14
fix: compute normalized_metadata.enroll_by_date in enterprise-canonical fashion

iloveagent57:
* Take the seat `upgrade_deadline_override` into account, and prefer it over `upgrade_deadline`.
* Prefer the course run's `enrollment_end` for exec ed 2u courses (over legacy `registration_deadline`).
* Prefer the course run's `enrollment_end` for other course types when no advertised course run seat exists (or if the seat has no upgrade deadline). 
ENT-7472

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/serializers.py
+++ b/enterprise_catalog/apps/catalog/serializers.py
@@ -127,22 +127,23 @@ class NormalizedContentMetadataSerializer(ReadOnlySerializer):
 
     @extend_schema_field(serializers.DateTimeField)
     def get_enroll_by_date(self, obj) -> str:
-        if obj.is_exec_ed_2u_course:
-            return self.additional_metadata.get('registration_deadline')
-
         if not self.advertised_course_run:
             return None
+
+        if obj.is_exec_ed_2u_course:
+            return (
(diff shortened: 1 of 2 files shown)


pwnage101 (approved):

enterprise_catalog/apps/catalog/serializers.py
```diff
@@ -127,22 +127,23 @@ def get_end_date(self, obj) -> str:
...
-        if obj.is_exec_ed_2u_course:
-            return self.additional_metadata.get('registration_deadline')
-
         if not self.advertised_course_run:
             return None
 
+        if obj.is_exec_ed_2u_course:
+            return (
+                self.advertised_course_run.get('enrollment_end')
+                or self.additional_metadata.get('registration_deadline')
```
marlonkeating, line 136: Do you know if we might have a course with no advertised run, but a populated `additional_metadata` section?

iloveagent57: Hmm, I'm not sure - but in that case, maybe it's valid to assert that the enroll_by_date is null?

PR #785, opened 2024-02-26
Take Exec Ed course data from course runs

marlonkeating:
## Description
Updates references for Exec Ed courses originally contained in the `additional_metadata` field, to now be taken from course runs.

## Ticket Link
[ENT-8248: update custom logic for non-ocm courses in enterprise-catalog](https://2u-internal.atlassian.net/browse/ENT-8248)

## Outstanding Questions
- [ ] Where all should these data changes be documented?

## Post-review
* [ ] Squash commits into discrete sets of changes
* [ ] Ensure that once the changes have been [deployed to stage](https://gocd.tools.edx.org/go/pipeline/activity/stage-enterprise_catalog), prod is manually deployed

PR #912, opened 2024-08-16
feat: take exec ed course data from course run instead of additional_metadata (take 2)

marlonkeating:
## Description
Updates references for Exec Ed courses originally contained in the `additional_metadata` field, to now be taken from course runs.  Fixing issues that were found in earlier attempt: https://github.com/openedx/enterprise-catalog/pull/785

## Ticket Link
[ENT-8248: update custom logic for non-ocm courses in enterprise-catalog](https://2u-internal.atlassian.net/browse/ENT-8248)

## Outstanding Questions
- [ ] Do we still need to process Exec Ed courses differently from OCM courses?

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tests/test_tasks.py
+++ b/enterprise_catalog/apps/api/tests/test_tasks.py
@@ -26,6 +26,7 @@ from enterprise_catalog.apps.catalog.constants import (
 from enterprise_catalog.apps.catalog.models import CatalogQuery, ContentMetadata
 from enterprise_catalog.apps.catalog.serializers import (
     DEFAULT_NORMALIZED_PRICE,
+    NormalizedContentMetadataSerializer,
     _find_best_mode_seat,
 )
 from enterprise_catalog.apps.catalog.tests.factories import (
@@ -52,6 +53,37 @@ def mock_task(self, *args, **kwargs):  # pylint: disable=unused-argument
 # An actual celery task would have a name attribute, and we use
 # it in a few places, so we patch it in here.
 mock_task.name = 'mock_task'
+
+
+def _hydrate_normalized_metadata(metadata_record):
+    """
+    Populate normalized_metadata fields for ContentMetadata
+    """
+    normalized_metadata_input = {
+        'course_metadata': metadata_record.json_metadata,
+    }
+    metadata_record.json_metadata['normalized_metadata'] =\
+        NormalizedContentMetadataSerializer(normalized_metadata_input).data
+    metadata_record.json_metadata['normalized_metadata_by_run'] = {}
+    for run in metadata_record.json_metadata.get('course_runs', []):
+        metadata_record.json_metadata['normalized_metadata_by_run'].update({
+            run['key']: NormalizedContentMetadataSerializer({
+                'course_run_metadata': run,
+                'course_metadata': metadata_record.json_metadata,
+            }).data
+        })
+
+
+def _hydrate_course_normalized_metadata():
+    """
+    Populate normalized_metadata fields for all course ContentMetadata
+    Needed for tests that generate test ContentMetadata, which does not have
+    normalized_metadata populated by default.
+    """
+    all_course_metadata = ContentMetadata.objects.filter(content_type=COURSE)
+    for course_metadata in all_course_metadata:
+        _hydrate_normalized_metadata(course_metadata)
+        course_metadata.save()
 
 
 @ddt.ddt
@@ -826,6 +858,8 @@ class IndexEnterpriseCatalogCoursesInAlgoliaTaskTests(TestCase):
         self.course_run_metadata_unpublished.catalog_queries.set([course_run_catalog_query])
         self.course_run_metadata_unpublished.save()
 
+        _hydrate_course_normalized_metadata()
+
     def _set_up_factory_data_for_algolia(self):
         expected_catalog_uuids = sorted([
             str(self.enterprise_catalog_courses.uuid),
@@ -1025,6 +1059,7 @@ class IndexEnterpriseCatalogCoursesInAlgoliaTaskTests(TestCase):
         test_course_1.save()
         test_course_2.save()
         test_course_3.save()
+        _hydrate_course_normalized_metadata()
 
         actual_algolia_products_sent = []
 
@@ -1123,6 +1158,7 @@ class IndexEnterpriseCatalogCoursesInAlgoliaTaskTests(TestCase):
         test_course_1.save()
         test_course_2.save()
         test_course_3.save()
+        _hydrate_course_normalized_metadata()
 
         actual_algolia_products_sent = []
 
@@ -2137,6 +2173,7 @@ class IndexEnterpriseCatalogCoursesInAlgoliaTaskTests(TestCase):
         )
         course_run_for_duplicate = ContentMetadataFactory(content_type=COURSE_RUN, parent_content_key='duplicateX')
         course_run_for_duplicate.catalog_queries.set([self.enterprise_catalog_course_runs.catalog_query])
+        _hydrate_course_normalized_metadata()
 
         actual_algolia_products_sent_sequence = []
 
--- a/enterprise_catalog/apps/api/v1/export_utils.py
+++ b/enterprise_catalog/apps/api/v1/export_utils.py
@@ -169,66 +169,98 @@ def program_hit_to_row(hit):
     return csv_row
 
 
+def _base_csv_row_data(hit):
+    """ Returns the formatted, shared attributes common across all course types. """
+    title = hit.get('title')
+    aggregation_key = hit.get('aggregation_key')
+    language = hit.get('language')
+    transcript_languages = ', '.join(hit.get('transcript_languages', []))
+    marketing_url = hit.get('marketing_url')
+    short_description = strip_tags(hit.get('short_description', ''))
+    subjects = ', '.join(hit.get('subjects', []))
+    skills = ', '.join([skill['name'] for skill in hit.get('skills', [])])
+    outcome = strip_tags(hit.get('outcome', ''))  # What You’ll Learn
+
+    # FIXME: currently ignores partner names when a course has multiple partners
+    partner_name = hit['partners'][0]['name'] if hit.get('partners') else None
+
+    empty_advertised_course_run = {}
+    advertised_course_run = hit.get('advertised_course_run', empty_advertised_course_run)
+    course_run_key = advertised_course_run.get('key')
+    min_effort = advertised_course_run.get('min_effort')
+    max_effort = advertised_course_run.get('max_effort')
+    weeks_to_complete = advertised_course_run.get('weeks_to_complete')  # Length
+
+    if start_date := advertised_course_run.get('start'):
+        start_date = parser.parse(start_date).strftime(DATE_FORMAT)
+
+    if end_date := advertised_course_run.get('end'):
+        end_date = parser.parse(end_date).strftime(DATE_FORMAT)
+
+    if enroll_by := advertised_course_run.get('enroll_by'):
+        enroll_by = datetime.datetime.fromtimestamp(enroll_by).strftime(DATE_FORMAT)
+
+    return {
+        'title': title,
+        'partner_name': partner_name,
+        'start_date': start_date,
+        'end_date': end_date,
+        'enroll_by': enroll_by,
+        'aggregation_key': aggregation_key,
+        'course_run_key': course_run_key,
+        'language': language,
+        'transcript_languages': transcript_languages,
+        'marketing_url': marketing_url,
+        'short_description': short_description,
+        'subjects': subjects,
+        'skills': skills,
+        'min_effort': min_effort,
+        'max_effort': max_effort,
+        'weeks_to_complete': weeks_to_complete,
+        'outcome': outcome,
+        'advertised_course_run': advertised_course_run,
+    }
+
+
 def course_hit_to_row(hit):
     """
     Helper function to construct a CSV row according to a single Algolia result course hit.
     """
+    row_data = _base_csv_row_data(hit)
     csv_row = []
-    csv_row.append(hit.get('title'))
+    csv_row.append(row_data.get('title'))
+    csv_row.append(row_data.get('partner_name'))
 
-    if hit.get('partners'):
-        csv_row.append(hit['partners'][0]['name'])
-    else:
-        csv_row.append(None)
+    advertised_course_run = row_data.get('advertised_course_run')
 
-    empty_advertised_course_run = {}
-    advertised_course_run = hit.get('advertised_course_run', empty_advertised_course_run)
-    if start_date := advertised_course_run.get('start'):
-        start_date = parser.parse(start_date).strftime(DATE_FORMAT)
-    csv_row.append(start_date)
-
-    if end_date := advertised_course_run.get('end'):
-        end_date = parser.parse(end_date).strftime(DATE_FORMAT)
-    csv_row.append(end_date)
+    csv_row.append(row_data.get('start_date'))
+    csv_row.append(row_data.get('end_date'))
 
     # upgrade_deadline deprecated in favor of enroll_by
     if upgrade_deadline := advertised_course_run.get('upgrade_deadline'):
         upgrade_deadline = datetime.datetime.fromtimestamp(upgrade_deadline).strftime(DATE_FORMAT)
     csv_row.append(upgrade_deadline)
-
-    if enroll_by := advertised_course_run.get('enroll_by'):
-        enroll_by = datetime.datetime.fromtimestamp(enroll_by).strftime(DATE_FORMAT)
-    csv_row.append(enroll_by)
-
-    pacing_type = advertised_course_run.get('pacing_type')
-    key = advertised_course_run.get('key')
-
+    csv_row.append(row_data.get('enroll_by'))
     csv_row.append(', '.join(hit.get('programs', [])))
     csv_row.append(', '.join(hit.get('program_titles', [])))
 
+    pacing_type = advertised_course_run.get('pacing_type')
     csv_row.append(pacing_type)
 
     csv_row.append(hit.get('level_type'))
-
     csv_row.append(hit.get('first_enrollable_paid_seat_price'))
-    csv_row.append(hit.get('language'))
-    csv_row.append(', '.join(hit.get('transcript_languages', [])))
-    csv_row.append(hit.get('marketing_url'))
-    csv_row.append(strip_tags(hit.get('short_description', '')))
-
-    csv_row.append(', '.join(hit.get('subjects', [])))
-    csv_row.append(key)
-    csv_row.append(hit.get('aggregation_key'))
-
-    skills = [skill['name'] for skill in hit.get('skills', [])]
-    csv_row.append(', '.join(skills))
-
-    advertised_course_run = hit.get('advertised_course_run', {})
-    csv_row.append(advertised_course_run.get('min_effort'))
-    csv_row.append(advertised_course_run.get('max_effort'))
-    csv_row.append(advertised_course_run.get('weeks_to_complete'))  # Length
-
-    csv_row.append(strip_tags(hit.get('outcome', '')))  # What You’ll Learn
+    csv_row.append(row_data.get('language'))
+    csv_row.append(row_data.get('transcript_languages'))
+    csv_row.append(row_data.get('marketing_url'))
+    csv_row.append(row_data.get('short_description'))
+    csv_row.append(row_data.get('subjects'))
+    csv_row.append(row_data.get('course_run_key'))
+    csv_row.append(row_data.get('aggregation_key'))
+    csv_row.append(row_data.get('skills'))
+    csv_row.append(row_data.get('min_effort'))
+    csv_row.append(row_data.get('max_effort'))
+    csv_row.append(row_data.get('weeks_to_complete'))
+    csv_row.append(row_data.get('outcome'))
 
     csv_row.append(strip_tags(hit.get('prerequisites_raw', '')))  # Pre-requisites
 
@@ -238,71 +270,33 @@ def course_hit_to_row(hit):
     return csv_row
 
 
-def fetch_and_format_registration_date(obj):
-    enroll_by_date = obj.get('registration_deadline')
-    stripped_enroll_by = enroll_by_date.split("T")[0]
-    formatted_enroll_by = None
-    try:
-        enroll_by_datetime_obj = datetime.datetime.strptime(stripped_enroll_by, '%Y-%m-%d')
-        formatted_enroll_by = enroll_by_datetime_obj.strftime('%m-%d-%Y')
-    except ValueError as exc:
-        logger.info(f"Unable to format registration deadline, failed with error: {exc}")
-    return formatted_enroll_by
-
-
 def exec_ed_course_to_row(hit):
     """
     Helper function to construct a CSV row according to a single executive education course hit.
     """
+    row_data = _base_csv_row_data(hit)
     csv_row = []
-    csv_row.append(hit.get('title'))
+    csv_row.append(row_data.get('title'))
+    csv_row.append(row_data.get('partners'))
 
-    if hit.get('partners'):
-        csv_row.append(hit['partners'][0]['name'])
-    else:
-        csv_row.append(None)
-    if hit.get('additional_metadata'):
-        start_date = None
-        additional_md = hit['additional_metadata']
-        if additional_md.get('start_date'):
-            start_date = parser.parse(additional_md['start_date']).strftime(DATE_FORMAT)
-        csv_row.append(start_date)
-
-        end_date = None
-        if additional_md.get('end_date'):
-            end_date = parser.parse(additional_md['end_date']).strftime(DATE_FORMAT)
-        csv_row.append(end_date)
-        formatted_enroll_by = fetch_and_format_registration_date(additional_md)
-    else:
-        csv_row.append(None)  # no start date
-        csv_row.append(None)  # no end date
-        formatted_enroll_by = None
-
-    csv_row.append(formatted_enroll_by)
-
-    adv_course_run = hit.get('advertised_course_run', {})
-    key = adv_course_run.get('key')
+    csv_row.append(row_data.get('start_date'))
+    csv_row.append(row_data.get('end_date'))
+    csv_row.append(row_data.get('enroll_by'))
 
     price = float(hit['entitlements'][0]['price'])
     csv_row.append(math.trunc(price))
-    csv_row.append(hit.get('language'))
-    csv_row.append(', '.join(hit.get('transcript_languages', [])))
-    csv_row.append(hit.get('marketing_url'))
-    csv_row.append(strip_tags(hit.get('short_description', '')))
-
-    csv_row.append(', '.join(hit.get('subjects', [])))
-    csv_row.append(key)
-    csv_row.append(hit.get('aggregation_key'))
-
-    skills = [skill['name'] for skill in hit.get('skills', [])]
-    csv_row.append(', '.join(skills))
-
-    csv_row.append(adv_course_run.get('min_effort'))
-    csv_row.append(adv_course_run.get('max_effort'))
-    csv_row.append(adv_course_run.get('weeks_to_complete'))  # Length
-
-    csv_row.append(strip_tags(hit.get('outcome', '')))  # What You’ll Learn
-
+    csv_row.append(row_data.get('language'))
+    csv_row.append(row_data.get('transcript_languages'))
+    csv_row.append(row_data.get('marketing_url'))
+    csv_row.append(row_data.get('short_description'))
+    csv_row.append(row_data.get('subjects'))
+    csv_row.append(row_data.get('course_run_key'))
+    csv_row.append(row_data.get('aggregation_key'))
+    csv_row.append(row_data.get('skills'))
+    csv_row.append(row_data.get('min_effort'))
+    csv_row.append(row_data.get('max_effort'))
+    csv_row.append(row_data.get('weeks_to_complete'))
+    csv_row.append(row_data.get('outcome'))
     csv_row.append(strip_tags(hit.get('full_description', '')))
 
     return csv_row
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -225,7 +225,7 @@ def _should_index_course(course_metadata):
         return not is_course_run_active(advertised_course_run)
 
     def deadline_passed_checker():
-        return _has_enroll_by_deadline_passed(course_json_metadata, advertised_course_run)
+        return _has_enroll_by_deadline_passed(course_json_metadata)
 
     for should_not_index_function, log_message in (
         (no_advertised_course_run_checker, 'no advertised course run'),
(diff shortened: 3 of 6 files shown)


enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -242,7 +242,13 @@ def course_hit_to_row(hit):
...
 
 def fetch_and_format_registration_date(obj):
-    enroll_by_date = obj.get('registration_deadline')
+    """
+    Args:
+        obj: CourseRun object
+    Returns:
+        string: date to register by
+    """
+    enroll_by_date = obj.get('end')
```
pmakwana93: Does `end` date always exist? Is it possible for some courses to be open ended and have NULL as the end date?

enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -264,26 +270,19 @@ def exec_ed_course_to_row(hit):
...
-            start_date = parser.parse(additional_md['start_date']).strftime(DATE_FORMAT)
-        csv_row.append(start_date)
 
-        end_date = None
-        if additional_md.get('end_date'):
-            end_date = parser.parse(additional_md['end_date']).strftime(DATE_FORMAT)
-        csv_row.append(end_date)
-        formatted_enroll_by = fetch_and_format_registration_date(additional_md)
+    adv_course_run = hit.get('advertised_course_run', {})
+    if (start_date := adv_course_run.get('start'), end_date := adv_course_run.get('end')) and start_date and end_date:
```
pmakwana93: nice use of the assignment operator

enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -264,26 +270,19 @@ def exec_ed_course_to_row(hit):
...
+        csv_row.append(start_date and parser.parse(start_date).strftime(DATE_FORMAT))
```
pmakwana93: what is appended to csv_row is start_date is None? Is None a valid input? What happens when we str format a None row?

enterprise_catalog/apps/catalog/serializers.py
```diff
@@ -131,22 +121,25 @@ def get_enroll_by_date(self, obj) -> str:
...
+        else:
+            logger.info(
+                f"No Seat Found for course run '{self.advertised_course_run.get('key')}'. "
+                f"Seats: {all_seats}"
+            )
+            return None
 
     @extend_schema_field(serializers.FloatField)
     def get_content_price(self, obj) -> float:
+        # TODO?: Unit test exec ed and non exec ed
```
pmakwana93: Remove TODO?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -236,20 +236,18 @@ def _has_enroll_by_deadline_passed(course_json_metadata, advertised_course_run):
...
     this is based on the verified upgrade deadline.
-    For 2u exec ed courses, this is based on the registration deadline.
+    For 2u exec ed courses, this is based on the enrollment_end deadline.
     """
     enroll_by_deadline_timestamp = 0
     if course_json_metadata.get('course_type') == EXEC_ED_2U_COURSE_TYPE:
-        additional_metadata = course_json_metadata.get('additional_metadata') or {}
-        registration_deadline = additional_metadata.get('registration_deadline')
-        if registration_deadline:
+        enrollment_end = advertised_course_run.get('end') or {}
```
brobro10000: [question] Just curious why were setting the `enrollment_end` date to the `end` date when there is already an `enrollment_end` field on the course run. 
Example endpoint: https://discovery.edx.org/api/v1/courses/LSE+PRE/

There is also recent changes that determine an `enroll_by` date field in Algolia to be utilized as part of variant assignments that look at the `enrollment_end` and the `upgrade_deadline` to determine to `enroll_by` date. This is for the purposes of making the `enroll_by` date content agnostic to support exec ed and OCM courses. 

Recently merged PRs:
https://github.com/openedx/enterprise-catalog/pull/915
https://github.com/openedx/enterprise-catalog/pull/921
https://github.com/openedx/enterprise-catalog/pull/917

enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -264,26 +270,19 @@ def exec_ed_course_to_row(hit):
...
-        if additional_md.get('start_date'):
-            start_date = parser.parse(additional_md['start_date']).strftime(DATE_FORMAT)
-        csv_row.append(start_date)
 
-        end_date = None
-        if additional_md.get('end_date'):
-            end_date = parser.parse(additional_md['end_date']).strftime(DATE_FORMAT)
-        csv_row.append(end_date)
-        formatted_enroll_by = fetch_and_format_registration_date(additional_md)
+    adv_course_run = hit.get('advertised_course_run', {})
```
adamstankiewicz: Given both Exec Ed and OCM course types (i.e., `exec_ed_course_to_row` & `course_hit_to_row`) rely on the many of the same formatted attributes:
* `start`
* `end`
* `enroll_by` - [inform] recent contribution from Titans, which takes into account both Exec Ed and OCM upstream via `_get_course_run` ([source](https://github.com/openedx/enterprise-catalog/blob/eb81cba6efdaa5d118b0386e33b158b5f709ca0b/enterprise_catalog/apps/catalog/algolia_utils.py#L1110)).
* `language`
* etc.

I wonder if it's worth creating a helper method to abstract away some of the shared logic across Exec Ed and OCM course types. That way, the `exec_ed_course_to_row` and `course_hit_to_row` methods stay focused on which **content-type specific attributes** to include and which **CSV cells to include and in what order**.

Something along the lines of:

```python
# export_utils.py
def _base_csv_row_data(hit):
    """ Returns the formatted, shared attributes common across all course types. """
    title = hit.get('title')
    aggregation_key = hit.get('aggregation_key')
    language = hit.get('language')
    transcript_languages = ', '.join(hit.get('transcript_languages', []))
    marketing_url = hit.get('marketing_url')
    short_description = strip_tags(hit.get('short_description', ''))
    subjects = ', '.join(hit.get('subjects', []))
    skills = [skill['name'] for skill in hit.get('skills', [])]
    outcome = strip_tags(hit.get('outcome', ''))  # What You’ll Learn

    # FIXME: currently ignores partner names when a course has multiple partners
    partner_name = hit['partners'][0]['name'] if hit.get('partners') else None

    empty_advertised_course_run = {}
    advertised_course_run = hit.get('advertised_course_run', empty_advertised_course_run)
    course_run_key = advertised_course_run.get('key')
    min_effort = advertised_course_run.get('min_effort')
    max_effort = advertised_course_run.get('max_effort')
    weeks_to_complete = advertised_course_run.get('weeks_to_complete')  # Length

    if start_date := advertised_course_run.get('start'):
        start_date = parser.parse(start_date).strftime(DATE_FORMAT)

    if end_date := advertised_course_run.get('end'):
        end_date = parser.parse(end_date).strftime(DATE_FORMAT)

    if enroll_by := advertised_course_run.get('enroll_by'):
        enroll_by = datetime.datetime.fromtimestamp(enroll_by).strftime(DATE_FORMAT)

    return {
        'title': title,
        'partner_name': partner_name,
        'start_date': start_date,
        'end_date': end_date,
        'enroll_by': enroll_by,
        'aggregation_key': aggregation_key,
        'course_run_key': course_run_key,
        'language': language,
        'transcript_languages': transcript_languages,
        'marketing_url': marketing_url,
        'short_description': short_description,
        'subjects': subjects,
        'skills': skills,
        'min_effort': min_effort,
        'max_effort': max_effort,
        'weeks_to_complete': weeks_to_complete,
        'outcome': outcome,
        'advertised_course_run': advertised_course_run,
    }
```

```python
# export_utils.py
def course_hit_to_row(hit):
    """
    Helper function to construct a CSV row according to a single Algolia result course hit.
    """
    csv_row = []

    base_csv_row_data = _base_csv_row_data(hit)
    csv_row.append(base_csv_row_data.get('title'))

    csv_row.append(base_csv_row_data.get('partner_name'))

    csv_row.append(base_csv_row_data.get('start_date'))
    csv_row.append(base_csv_row_data.get('end_date'))

    advertised_course_run = base_csv_row_data.get('advertised_course_run')

    # upgrade_deadline deprecated in favor of enroll_by
    if upgrade_deadline := advertised_course_run.get('upgrade_deadline'):
        upgrade_deadline = datetime.datetime.fromtimestamp(upgrade_deadline).strftime(DATE_FORMAT)
    csv_row.append(upgrade_deadline)

    csv_row.append(base_csv_row_data.get('enroll_by'))

    csv_row.append(', '.join(hit.get('programs', [])))
    csv_row.append(', '.join(hit.get('program_titles', [])))

    csv_row.append(advertised_course_run.get('pacing_type'))

    csv_row.append(hit.get('level_type'))

    csv_row.append(hit.get('first_enrollable_paid_seat_price'))
    csv_row.append(base_csv_row_data.get('language'))
    csv_row.append(base_csv_row_data.get('transcript_languages'))
    csv_row.append(base_csv_row_data.get('marketing_url'))
    csv_row.append(base_csv_row_data.get('short_description'))

    csv_row.append(base_csv_row_data.get('subjects'))
    csv_row.append(base_csv_row_data.get('course_run_key'))
    csv_row.append(base_csv_row_data.get('aggregation_key'))

    csv_row.append(base_csv_row_data.get('skills'))

    csv_row.append(base_csv_row_data.get('min_effort'))
    csv_row.append(base_csv_row_data.get('max_effort'))
    csv_row.append(base_csv_row_data.get('weeks_to_complete'))  # Length

    csv_row.append(base_csv_row_data.get('outcome'))  # What You’ll Learn

    csv_row.append(strip_tags(hit.get('prerequisites_raw', '')))  # Pre-requisites

    catalogs = fetch_catalog_types(hit)  # Catalog types
    csv_row.append(', '.join(catalogs))

    return csv_row
```

```python
def exec_ed_course_to_row(hit):
    """
    Helper function to construct a CSV row according to a single executive education course hit.
    """
    csv_row = []
    base_csv_row_data = _base_csv_row_data(hit)
    csv_row.append(base_csv_row_data.get('title'))

    csv_row.append(base_csv_row_data.get('partner_name'))

    csv_row.append(base_csv_row_data.get('start_date'))
    csv_row.append(base_csv_row_data.get('end_date'))
    csv_row.append(base_csv_row_data.get('enroll_by'))

    price = float(hit['entitlements'][0]['price'])
    csv_row.append(math.trunc(price))
    csv_row.append(base_csv_row_data.get('language'))
    csv_row.append(base_csv_row_data.get('transcript_languages'))
    csv_row.append(base_csv_row_data.get('marketing_url'))
    csv_row.append(base_csv_row_data.get('short_description'))

    csv_row.append(base_csv_row_data.get('subjects'))
    csv_row.append(base_csv_row_data.get('course_run_key'))
    csv_row.append(base_csv_row_data.get('aggregation_key'))

    csv_row.append(base_csv_row_data.get('skills'))

    csv_row.append(base_csv_row_data.get('min_effort'))
    csv_row.append(base_csv_row_data.get('max_effort'))
    csv_row.append(base_csv_row_data.get('weeks_to_complete'))  # Length

    csv_row.append(base_csv_row_data.get('outcome'))  # What You’ll Learn

    csv_row.append(strip_tags(hit.get('full_description', '')))

    return csv_row
```

enterprise_catalog/apps/api/v1/export_utils.py
adamstankiewicz: [clarification, relates to Hamzah's comment below] `obj.get('end')` is referencing the **end date** of the course, not the **registration deadline**, right?

Presumably, the registration deadline (as determined by the field `enrollment_end`) is much earlier than the `end` date, since you need to register well in advance of the course ending :) 

Also, a recent contribution this week ([source](https://github.com/openedx/enterprise-catalog/blob/eb81cba6efdaa5d118b0386e33b158b5f709ca0b/enterprise_catalog/apps/catalog/algolia_utils.py#L1110)) from Titans updated the upstream `_get_course_run` to include an `enroll_by` which handles both Exec Ed and OCM course types (i.e., `upgrade_deadline` for OCM or `enrollment_end` for Exec Ed). Given this, `fetch_and_format_registration_date` likely isn't needed anymore.

enterprise_catalog/apps/api/v1/export_utils.py
brobro10000: +1

enterprise_catalog/apps/api/v1/export_utils.py
marlonkeating: I did some digging into production data and find it does happen, but very rarely, and it seems like a data issue rather than a legitimate configuration.

enterprise_catalog/apps/api/v1/export_utils.py
marlonkeating: We shouldn't be str formatting any `None` values here, since we check for falsy values before trying to format them.

enterprise_catalog/apps/catalog/algolia_utils.py
marlonkeating: Good catch!

enterprise_catalog/apps/api/v1/export_utils.py
marlonkeating: Thanks for the info!

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -450,6 +450,23 @@ def test_update_full_metadata(self, mock_oauth_client, mock_partition_course_key
                 ]
             }],
             'advertised_course_run_uuid': course_run_3_uuid,
+            'advertised_course_run': {
```
adamstankiewicz: [clarification/sanity check] Does the base metadata for top-level course includes both a `advertised_course_run_uuid` _and_ the serialized `advertised_course_run` as depicted here? My understanding is this is not a field that actually exists on real course metadata, given the `json_metadata` discerned from discovery ([example API response](https://discovery.edx.org/api/v1/courses/UCT+ARM/)) does _not_ include this sibling `advertised_course_run` field 🤔

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -640,8 +657,10 @@ def test_update_full_metadata_exec_ed(self, mock_oauth_client, mock_partition_co
                 'key': course_run_key,
                 'uuid': course_run_uuid,
                 # Use dummy 2022 dates that we will assert are overwritten.
```
adamstankiewicz: nit: this comment is no longer needed because there is no overwriting step in this flow anymore.

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -640,8 +657,10 @@ def test_update_full_metadata_exec_ed(self, mock_oauth_client, mock_partition_co
...
-                'start': '2022-03-01T00:00:00Z',
-                'end': '2022-03-01T00:00:00Z',
+                'start': '2023-03-01T00:00:00Z',
+                'end': '2023-04-09T23:59:59Z',
+                'enrollment_end': '2023-02-01T00:00:00Z',
+                "first_enrollable_paid_seat_price": 2900,
```
adamstankiewicz: This test case relates to Exec Ed course; given this, I don't believe there will be a `first_enrollable_paid_seat_price` (i.e., `first_enrollable_paid_seat_price: None`). This is due to the fact that Exec Ed courses do not have paid `seats`; they have a paid `entitlement` from the top-level course object.

Example API response for top-level Exec Ed course: https://discovery.edx.org/api/v1/courses/UCT+ARM/

```
{
  "entitlements": [
        {
            "mode": "paid-executive-education",
            "price": "894.00",
            "currency": "USD",
            "sku": "2B541DB",
            "expires": null
        }
    ],
}
```

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -243,20 +243,18 @@ def _has_enroll_by_deadline_passed(course_json_metadata, advertised_course_run):
...
+        if enrollment_end:
             enroll_by_deadline_timestamp = datetime.datetime.strptime(
-                registration_deadline,
+                enrollment_end,
                 '%Y-%m-%dT%H:%M:%S%z',
             ).timestamp()
     else:
         enroll_by_deadline_timestamp = _get_verified_upgrade_deadline(advertised_course_run)
-
     return enroll_by_deadline_timestamp < localized_utcnow().timestamp()
```
adamstankiewicz: [clarification] Could we use the `normalized_metadata` that exists from the `course_json_metadata` passed into this method? That is, could we refer to the consolidated content metadata across course types already exposed without having to introduce similar conditional branching logic based on `course_type`? 

`normalized_metadata` has this branching logic abstracted away.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -66,12 +70,16 @@ class AlgoliaUtilsTests(TestCase):
         {
             'course_type': EXEC_ED_2U_COURSE_TYPE,
             'expected_result': True,
-            'additional_metadata': {'registration_deadline': '2073-03-21T23:59:59Z'},
+            'start': _fake_upgrade_deadline(1),
+            'end': _fake_upgrade_deadline(30),
```
adamstankiewicz: [clarification/suggestion] Why are we using a method seemingly specific to `upgrade_deadlines` for `start` and `end` date properties?

Should there be a distinct helper method for general timestamp vs. reusing the same `_fake_upgrade_deadline` for non-upgrade deadline use cases?

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -66,12 +70,16 @@ class AlgoliaUtilsTests(TestCase):
...
         },
         {
             'course_type': EXEC_ED_2U_COURSE_TYPE,
             'expected_result': False,
-            'additional_metadata': {'registration_deadline': '2021-03-21T23:59:59Z'},
+            'start': _fake_upgrade_deadline(-30),
+            'end': _fake_upgrade_deadline(-1),
+            'enrollment_end': _fake_upgrade_deadline(-30)
         },
         {
```
adamstankiewicz, line 94: [clarification] Why are the `additional_metadata` inputs in other tests above removed, while the test case below continues to have it included?

```
'additional_metadata': {'rando-key': 'blah'},
```

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -92,7 +100,10 @@ def test_should_index_course(
         course_run_availability='current',
         seats=None,
         course_type=COURSE,
-        additional_metadata=None
+        additional_metadata=None,
```
adamstankiewicz: [clarification] Similar to above comment, should this arg be removed?

enterprise_catalog/apps/catalog/serializers.py
```diff
@@ -131,19 +121,21 @@ def get_enroll_by_date(self, obj) -> str:
...
+            return seat.get('upgrade_deadline_override') or seat.get('upgrade_deadline')
+
+        if enrollment_end := self.advertised_course_run.get('enrollment_end'):
+            return enrollment_end
+        else:
+            logger.info(
+                f"No Seat Found for course run '{self.advertised_course_run.get('key')}'. "
+                f"Seats: {all_seats}"
+            )
+            return None
```
adamstankiewicz: [clarification] It looks like the fallback to `enrollment_end` for OCM course types when an `upgrade_deadline` is missing was removed (i.e.,`return upgrade_deadline or self.advertised_course_run.get('enrollment_end')`, previously line 146).

So, if there is an upgrade_deadline, the enrollment_end is no longer considered (this is the case in the current state as well). Given this, might it make sense to consider the _earliest_ of `upgrade_deadline`, `upgrade_deadline_override` or `enrollment_end` given that `enrollment_end` theoretically could be before `upgrade_deadline`.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -493,7 +515,8 @@ def test_get_upcoming_course_runs(self, searchable_course, expected_course_runs)
                         'is_marketable': True,
                         'availability': 'Current',
                         'start': "2018-01-04T00:00:00Z",
-                        'end': "3022-12-31T23:59:00Z",
+                        'end': NEXT_MONTH_DATE,
+                        'enrollment_end': NEXT_MONTH_DATE,
```
adamstankiewicz: [sanity check] Should these values differ? Presumably the `enrollment_end` should be _before_ the course `end`? Might relate to another comment above about (incorrectly) associating `enrollment_end` with the `end` date.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -535,12 +560,13 @@ def test_get_upcoming_course_runs(self, searchable_course, expected_course_runs)
                     'pacing_type': 'instructor_paced',
                     'availability': 'Current',
                     'start': "2018-01-04T00:00:00Z",
-                    'end': "3022-12-31T23:59:00Z",
+                    'end': NEXT_MONTH_DATE,
+                    'enrollment_end': NEXT_MONTH_DATE,
```
adamstankiewicz: [sanity check] Should these values differ? Presumably the `enrollment_end` should be _before_ the course `end`? Might relate to another comment above about (incorrectly) associating `enrollment_end` with the `end` date.

enterprise_catalog/apps/catalog/tests/test_serializers.py
```diff
@@ -56,13 +56,13 @@ def test_enroll_by_date_is_exec_ed_course_no_enrollment_end(self):
         )
         course_content.json_metadata['course_type'] = EXEC_ED_2U_COURSE_TYPE
         course_content.json_metadata['additional_metadata'] = {
-            'will': 'not be ignored',
-            'registration_deadline': '1999-12-31T23:59:59Z',
+            'will': 'be ignored',
+            'registration_deadline': None,
         }
```
adamstankiewicz: [clarification] Given the removal of `additional_metadata` in this before, should this also be removed as well given the use of `enrollment_end` from the course run instead?

enterprise_catalog/apps/catalog/tests/test_serializers.py
```diff
@@ -56,13 +56,13 @@ def test_enroll_by_date_is_exec_ed_course_no_enrollment_end(self):
...
         course_content.json_metadata['course_runs'] = [
             {
                 'uuid': 'the-course-run-uuid',
-                'enrollment_end': None
+                'enrollment_end': '1999-12-31T23:59:59Z',
```
adamstankiewicz: [sanity check] Given this test case seems to intend to work with Exec Ed _without_ an `enrollment_end` (i.e., "test_enroll_by_date_is_exec_ed_course_**no_enrollment_end**"), I'm confused why we're passing an `enrollment_end` here 🤔

enterprise_catalog/apps/catalog/algolia_utils.py
adamstankiewicz: [clarification] This is still referring to `end`, which is _not_ the `enrollment_end` as described in your updated docstring. See related comment below for why we wouldn't need to consider any branching logic here regardless.

adamstankiewicz (commented):
Thanks for updating 😺 @brobro10000 and I did a pair PR review on the changes and left some additional questions/clarifications/suggestions/etc.

Let us know if you want to rubber duck on anything!

enterprise_catalog/apps/catalog/serializers.py
marlonkeating: That's a good question. I don't have context on whether we want the earliest listed deadline among the options, or the deadline that comes from the most _authoritative_ source ( `enrollment_end` > `upgrade_deadline_override` > `upgrade_deadline` ?)

adamstankiewicz: I believe we'd want something like:

```python
upgrade_deadline = seat.get('upgrade_deadline_override') or seat.get('upgrade_deadline')
enrollment_end = self.advertised_course_run.get('enrollment_end')
enroll_by = min(filter(None, [upgrade_deadline, enrollment_end]))
```

In the case of B2B enrollments (i.e., because we enroll enterprise learners directly to the paid `verified` track), the `upgrade_deadline` and `enrollment_end` pretty much have synonymous weight; either one of them means the learner can't successfully enroll as a subsidized enterprise learner (i.e., can't enroll in paid track due to upgrade deadline, or can't enroll due to course enrollment ended).

enterprise_catalog/apps/catalog/algolia_utils.py
marlonkeating: From what I can see we should be able to use `normalized_metadata`...  I just read through https://github.com/openedx/enterprise-catalog/blob/master/docs/decisions/0007-normalized-metadata-course-types.rst, and is it still the position that we aren't making an active effort to change over to using `normalized_metadata` in business logic, it is available for use where convenient?

adamstankiewicz: When `normalized_metadata` was introduced, we limited the scope to be an additive-only change, with the plan to incrementally migrate to `normalized_metadata` anywhere we had existing branching logic between course types (i.e., OCM vs EE) around metadata like dates/price/etc. We have switched over to `normalized_metadata` in several places already on various consuming services, so this is another incremental opportunity to simplify the existing implementation by relying on the abstraction. The hope is that relying on `normalized_metadata` makes content metadata easier to reason about when parsing it within business logic such as here.

enterprise_catalog/apps/catalog/tests/test_serializers.py
marlonkeating: Good catch!

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
marlonkeating: Good catch!

enterprise_catalog/apps/api/tests/test_tasks.py
marlonkeating: Good catch, removed!

enterprise_catalog/apps/api/tests/test_tasks.py
marlonkeating: Thanks for the info!  Removed `first_enrollable_paid_seat_price` from test data.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
marlonkeating: Good call, refactored into a more general helper method.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -243,20 +243,24 @@ def _has_enroll_by_deadline_passed(course_json_metadata, advertised_course_run):
...
```
brobro10000: [nit] For the case where an advertised course run does not have an `enrollment_end` and falls back to an empty `{}`, how would the return statement handle the logic?

`return {} < localized_utcnow().timestamp()`

enterprise_catalog/apps/catalog/serializers.py
```diff
@@ -130,12 +130,18 @@ def get_end_date(self, obj) -> str:  # pylint: disable=unused-argument
...
-        seat = _find_best_mode_seat(all_seats)
+
         upgrade_deadline = None
-        if seat:
+        if seat := _find_best_mode_seat(all_seats):
             upgrade_deadline = seat.get('upgrade_deadline_override') or seat.get('upgrade_deadline')
-        return upgrade_deadline or self.course_run_metadata.get('enrollment_end')
+
+        enrollment_end = self.course_run_metadata.get('enrollment_end')
+        return min(filter(None, [upgrade_deadline, enrollment_end]), default=None)
```
brobro10000, line 150: 🥳

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -25,11 +25,20 @@
...
 
+def _days_from_now_ts(days_from_now=0):
+    deadline = _days_from_now(days_from_now)
+    return {'str': deadline, 'timestamp': to_timestamp(deadline)}
+
+
+TOMORROW, NEXT_WEEK, NEXT_MONTH = [_days_from_now_ts(days) for days in [1, 7, 30]]
+NEXT_MONTH_DATE = _days_from_now(30)
+
+
```
brobro10000: 🥳

brobro10000 (commented):
Looks good on my end, curious about 1 nit though 👍🏽

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -243,20 +243,24 @@ def _has_enroll_by_deadline_passed(course_json_metadata, advertised_course_run):
...
+    For 2u exec ed courses, this is based on the enrollment_end deadline.
+    """
+    enroll_by_deadline_timestamp = None
+    if 'normalized_metadata' in course_json_metadata:
+        enroll_by_deadline = course_json_metadata['normalized_metadata']['enroll_by_date']
+        enroll_by_deadline_timestamp = datetime.datetime.strptime(
+            enroll_by_deadline,
+            '%Y-%m-%dT%H:%M:%S%z',
+        ).timestamp()
+    elif course_json_metadata.get('course_type') == EXEC_ED_2U_COURSE_TYPE:
```
adamstankiewicz: [curious/suggestion] When will `normalized_metadata` _not_ exist in `course_json_metadata`, where the fallback `elif` is needed? I think the  conditional around whether `normalized_metadata` exists and the fallback `elif` could probably be removed 🤔 Rationale: `normalized_metadata` is added to `ContentMetadata`'s `json_metadata` during the `update_content_metadata` so I believe it's a fair assumption it will always exist ([source](https://github.com/openedx/enterprise-catalog/blob/c847837ade2f6238c38dd7d1666943eee07d280c/enterprise_catalog/apps/api/tasks.py#L296-L297)).

Given this, I believe the pre-existing Exec Ed vs. OCM checks here and below (i.e., `_get_verified_upgrade_deadline`) could be removed to only rely on `normalized_metadata` as the logic below is essentially duplicative of `normalized_metadata`.

enterprise_catalog/apps/catalog/algolia_utils.py
adamstankiewicz: Similar question if only accounting for the `normalized_metadata` (i.e., sans the Exec Ed and OCM conditional checks to remove largely duplicative logic), and the `enroll_by_date` is `None`.

image

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1103,6 +1107,7 @@ def _get_course_run(full_course_run):
         'availability': full_course_run.get('availability'),
         'start': full_course_run.get('start'),
         'end': full_course_run.get('end'),
+        'enrollment_end': full_course_run.get('enrollment_end'),
```
adamstankiewicz: [clarification] Is `enrollment_end` already abstracted into the `enroll_by` field below? I would consider this `enrollment_end` deprecated similar to `upgrade_deadline`, in favor of relying on the transformed "interface" for `enroll_by`.

enterprise_catalog/apps/catalog/serializers.py
```diff
@@ -130,12 +130,18 @@ def get_end_date(self, obj) -> str:  # pylint: disable=unused-argument
     def get_enroll_by_date(self, obj) -> str:  # pylint: disable=unused-argument
         if not self.course_run_metadata:
             return None
+
+        if self.course.is_exec_ed_2u_course:
```
adamstankiewicz: Nice cleanup here to short-circuit for the Exec Ed case!

enterprise_catalog/apps/catalog/algolia_utils.py
marlonkeating: Removed check by recommendation [here](https://github.com/openedx/enterprise-catalog/pull/912#discussion_r1759617354)

enterprise_catalog/apps/catalog/algolia_utils.py
marlonkeating: Got it, removed the check so we're only going by `normalized_metadata` now.

enterprise_catalog/apps/catalog/algolia_utils.py
marlonkeating: Good point!  Removed it.

brobro10000 (approved):
🥳  LGTM 👍🏽

adamstankiewicz (approved):

PR #915, opened 2024-08-19
feat: add course run fields for Algolia indexing

brobro10000:
Adds net new fields to be indexed into Algolia related to course runs in support of variant based assignments.

Fields:
  - `enroll_by` - The course run enrollment deadline. This field handles both exec-ed and OCM course types
  -  `is_active` - Computed value determining if the course is active (`is_enrollable`, `is_marketable`,  `is_published`) and either `Current` or `Upcoming`
  - `status` - The `status` field from course runs to determine if its one of corresponding [choices](https://github.com/openedx/course-discovery/blob/721de405939748ad1c59333a665c4128b21be5fb/course_discovery/apps/course_metadata/choices.py#L5)
  - `is_marketable` - Whether the course is marketable based on [these](https://github.com/openedx/course-discovery/blob/6aa995766f356fd37c28ec791c85d51cae79bb7f/course_discovery/apps/course_metadata/models.py#L2920) parameters

Also takes into account the`enroll_by` date to determine if the course run should be indexed alongside the `end` date. 

## Post-review


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1201,6 +1204,27 @@ def _get_verified_upgrade_deadline(full_course_run):
     return (datetime.datetime(3000, 1, 1)).timestamp()
 
 
+def _get_is_available_course_run(full_course_run):
+    is_marketable = full_course_run.get('is_marketable', False)
+    is_enrollable = full_course_run.get('is_enrollable', False)
+    availability = full_course_run.get('availability')
+    is_available = is_marketable and is_enrollable and (availability == 'Current')
```
adamstankiewicz: [clarification] Might this preclude an upcoming, not yet "Current" course run from being indexed/assigned?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1198,7 +1217,61 @@ def _get_verified_upgrade_deadline(full_course_run):
...
+    is_marketable: true
+    is_enrollable: true
+    availability: 'Current' || 'Upcoming'
+    is_published: 'Published'
+
+    It resolves the logic into an indexed field on the course run labeled 'is_available'
+    """
+    is_active = is_course_run_active(full_course_run)
+    availability = full_course_run.get('availability')
+    is_current_or_upcoming_availability = availability in {'Current', 'Upcoming'}
```
adamstankiewicz: [suggestion] I believe we'll also want to include "Starting Soon", per how the `availability` is computed ([source](https://github.com/openedx/course-discovery/blob/721de405939748ad1c59333a665c4128b21be5fb/course_discovery/apps/course_metadata/models.py#L2565-L2584)). Alternatively, possibly could also consider an exclusion list with "Archived" vs. an allow list.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -308,6 +308,9 @@ def test_get_course_subjects(self, course_metadata, expected_subjects):
                     'min_effort': 10,
                     'max_effort': 14,
                     'weeks_to_complete': 13,
+                    'status': 'Published',
```
adamstankiewicz: I believe the `status` field is lowercase in discovery API response?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1198,7 +1217,61 @@ def _get_verified_upgrade_deadline(full_course_run):
...
+    `_get_verified_upgrade_deadline` retrieves OCM related course run end dates
+    `enrollment_end` corresponds to an Exec-ed related course run end date
+
+    If no end date is provided in either field, it returns the ALGOLIA_DEFAULT_TIMESTAMP
+    since Algolia cannot filter on null values
+    """
+    upgrade_deadline_timestamp = _get_verified_upgrade_deadline(full_course_run=full_course_run)
+    enrollment_end = full_course_run.get('enrollment_end', ALGOLIA_DEFAULT_TIMESTAMP)
+    if _is_default_algolia_timestamp(upgrade_deadline_timestamp) == _is_default_algolia_timestamp(enrollment_end):
+        return ALGOLIA_DEFAULT_TIMESTAMP
```
adamstankiewicz: [clarification] Can you help me understand this conditional a bit? When would we expect `upgrade_deadline_timestamp` and/or `enrollment_end` to be the default Algolia timestamp of `datetime.datetime(3000, 1, 1)` (`ALGOLIA_DEFAULT_TIMESTAMP`)?

enterprise_catalog/apps/catalog/utils.py
```diff
@@ -138,3 +140,28 @@ def batch_by_pk(ModelClass, extra_filter=Q(), batch_size=10000):
...
+    Ex. to_timestamp("2024-07-30T00:00:00Z") -> 1722297600.0
+
+    The decimal represents a timestamp epoch time down to the millisecond
+
+    This is useful if we need to pass epoch time to an indexable Algolia value
+    which requires it to be in epoch format in order for the indexed field to be
+    filtered/sorted.
+    """
+    try:
+        if isinstance(date_str_or_timestamp, (int, float)):
```
adamstankiewicz: [curious] Is this method expected to be called with an existing timestamp value thats already transformed? When would this happen? Wondering if you could always assume this method accepts a datetime string?

enterprise_catalog/apps/catalog/utils.py
```diff
@@ -138,3 +140,28 @@ def batch_by_pk(ModelClass, extra_filter=Q(), batch_size=10000):
...
+            return date_str_or_timestamp
+        parse_datetime = parser.parse(date_str_or_timestamp)
+        timestamp = time.mktime(parse_datetime.timetuple())
```
adamstankiewicz: [suggestion] might recommend using `datetime` here, e.g. something like:

```python
from datetime import datetime

try:
    dt = datetime.fromisoformat(timestamp_str)
    epoch_time = int(dt.timestamp())
    return epoch_time
except (ValueError, TypeError) as error:
    LOGGER.error(f"[to_timestamp][{error}] Could not parse date string: {date_str_or_timestamp}")
    return None
```

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1135,9 +1145,18 @@ def get_course_runs(course):
     course_runs = course.get('course_runs') or []
     for full_course_run in course_runs:
         this_course_run = _get_course_run(full_course_run)
-        if this_course_run.get('end'):
+        if this_course_run.get('enroll_by') and this_course_run.get('end'):
```
adamstankiewicz: [question] What if a course run has `end` date but no `enroll_by` date?

adamstankiewicz (commented):
Still perusing some of the PR, but here's a few comments for now.

enterprise_catalog/apps/catalog/algolia_utils.py
brobro10000: This conditional handles follows the logic of `_get_verified_upgrade_deadline` which returns a default value far in the future if no upgrade deadline value exist.  I continued this logic when determining the `enrollment_end` date to default to the `ALGOLIA_DEFAULT_TIMESTAMP` if none exist.

Further down we choose the minimum epoch value between `_get_verified_upgrade_deadline` and  `enrollment_end`. 

Considering this logic, the way I structured this function these two lines seem redundant 😓 
Great catch.

enterprise_catalog/apps/catalog/utils.py
brobro10000: In `_get_course_run_enroll_by_date_timestamp` we are retrieving the `enrollment_end` field or defaulting to the `ALGOLIA_DEFAULT_TIMESTAMP` which is a use case for including a check to see if the value passed into the argument is already a timestamp. So for this use case `to_timestamp` would need to check if the passed argument is already a timestamp

I guess this was an attempt to start coding defensively  while still returning a sane default.

enterprise_catalog/apps/catalog/algolia_utils.py
brobro10000: Because defaults should always exist for the `enroll_by` date (because of the ALGOLIA_DEFAULT_VALUE), I don't anticipate it happening. Given some of the logic below this line is dependent on the existence of both an `enroll_by` and `end` date from `this_course_run`, I still think the `and` conditional should exist, but I'll go through and check if theres a way to mitigate the scope to be non-dependent on both values 👍🏽 .

adamstankiewicz: Ah, good point that the `enroll_by` here intends to be the already-transformed value for indexing, even if not set (`ALGOLIA_DEFAULT_VALUE`). Thanks for the clarification.

enterprise_catalog/apps/catalog/constants.py
```diff
@@ -1,10 +1,14 @@
+import datetime
```
adamstankiewicz: nit: `from datetime import datetime`

to avoid `datetime.datetime(3000, ...)`

enterprise_catalog/apps/catalog/algolia_utils.py
brobro10000: I've updated to logic to more explicitly handle the instances of both and `enroll_by` or an `end` date.

enterprise_catalog/apps/catalog/utils.py
brobro10000: Updated to handle this logic outside of the function and keep it single responsibility in its inputs and returns 👍🏽

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1198,7 +1222,49 @@ def _get_verified_upgrade_deadline(full_course_run):
...
+    is_not_archived_availability = availability != 'Archived'
+    is_active = course_run_is_active and is_not_archived_availability
+    if not is_active:
+        logger.info(
+            f'[_get_is_active_course_run] course run is not active '
+            f'key: {full_course_run.get("key")}, '
+            f'is_marketable: {full_course_run.get("is_marketable")}, '
+            f'is_enrollable: {full_course_run.get("is_enrollable")}, '
+            f'availability: {availability}, '
+            f'status: {full_course_run.get("status")}'
```
adamstankiewicz, line 1246: [nit/curious] if `is_available` doesn't use `status`, is it worth logging here?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1100,6 +1107,10 @@ def _get_course_run(full_course_run):
         'max_effort': full_course_run.get('max_effort'),
         'weeks_to_complete': full_course_run.get('weeks_to_complete'),
         'upgrade_deadline': _get_verified_upgrade_deadline(full_course_run),
```
adamstankiewicz: [nit/suggestion] I wonder if it's worth adding a comment on this line that indicates `upgrade_deadline` as "deprecated", considering its now abstracted via the new field `enroll_by` below.

```python
'upgrade_deadline': _get_verified_upgrade_deadline(full_course_run),  # deprecated in favor of `enroll_by`
```

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1135,9 +1146,22 @@ def get_course_runs(course):
...
+                    course_run_enroll_by_datetime + datetime.timedelta(days=LATE_ENROLLMENT_THRESHOLD_DAYS)
+                # check for runs within the late enrollment threshold
+                is_late_enrollment_enroll_by_date_before_now = \
+                    course_run_late_enrollment_enroll_by_datetime.replace(tzinfo=UTC) < localized_utcnow()
+            if this_course_run.get('end'):
+                course_run_end = parser.parse(this_course_run.get('end'))
+                # check for runs within course run end date
+                is_course_run_end_date_before_now = course_run_end < localized_utcnow()
+            if is_late_enrollment_enroll_by_date_before_now or is_course_run_end_date_before_now:
                 # skip old course runs
```
adamstankiewicz: [sanity check] I think when I was originally reviewing the original logic/ticketing, I may have been assuming the late enrollment was related to the `end` field, but it's not; late enrollment is pertinent to `enroll_by` as this logic is demonstrating. My assumption at ticketing was that we'd be making `_get_course_runs` less restrictive, but this is in fact, making it more restrictive by adding a second filtering parameter on `enroll_by`. Now, this method is skipping either course runs that have ended per `end` _or_ not-yet-ended runs with an elapsed (30+ days) `enroll_by`.

Perhaps this is actually what we want here, though 🤔 As you know, the primary consumer of these course runs currently is that related `CatalogWorkbookView` (i.e., supports the "Download results" CTA on [explore-catalog.edx.org](https://explore-catalog.edx.org/)), allowing CSV downloads of the search results.

I think the question is whether the "Download results" CTA from explore-catalog should or shouldn't also include course runs ineligible for late enrollment. My two cents would be to exclude them as you have here, but this is a slight possible change to existing behavior of "Download results".

I think this change makes sense though, such that we aren't including course runs in the downloaded CSV which don't have a chance to be enrolled into verified anyways (with or without late enrollment), making the results more useful for purposes of finding course runs that are potentially eligible to current/future customers.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1135,9 +1146,22 @@ def get_course_runs(course):
     course_runs = course.get('course_runs') or []
     for full_course_run in course_runs:
         this_course_run = _get_course_run(full_course_run)
-        if this_course_run.get('end'):
-            course_run_end = parser.parse(this_course_run.get('end'))
-            if course_run_end < localized_utcnow():
+        if this_course_run.get('enroll_by') or this_course_run.get('end'):
```
adamstankiewicz: Could this be simplified to reduce nested conditionals?

```python
for full_course_run in course_runs:
    this_course_run = _get_course_run(full_course_run)
    enroll_by = this_course_run.get('enroll_by')
    end = this_course_run.get('end')
    is_late_enrollment_enroll_by_date_before_now = False
    is_end_date_before_now = False
    
    # determine whether enroll_by (if any) is before now + late enrollment offset
    if enroll_by:
        enroll_by_datetime = datetime.datetime.fromtimestamp(enroll_by)
        # offsets enroll_by date by the late enrollment threshold
        late_enrollment_enroll_by_datetime = \
            enroll_by_datetime + datetime.timedelta(days=LATE_ENROLLMENT_THRESHOLD_DAYS)
        # check for runs within the late enrollment threshold
        is_late_enrollment_enroll_by_date_before_now = \
            enroll_by_datetime.replace(tzinfo=UTC) < localized_utcnow()

    # determine whether end (if any) is before now
    if end:
        course_run_end = parser.parse(end)
        # check for runs within course run end date
        is_end_date_before_now = course_run_end < localized_utcnow()

    if is_late_enrollment_enroll_by_date_before_now or is_end_date_before_now:
        # skip course runs which have already ended or has elapsed its enroll-by date (including late enrollment offset).
        continue

    # ...
```

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1135,9 +1146,22 @@ def get_course_runs(course):
...
+                is_late_enrollment_enroll_by_date_before_now = \
+                    course_run_late_enrollment_enroll_by_datetime.replace(tzinfo=UTC) < localized_utcnow()
+            if this_course_run.get('end'):
+                course_run_end = parser.parse(this_course_run.get('end'))
+                # check for runs within course run end date
+                is_course_run_end_date_before_now = course_run_end < localized_utcnow()
+            if is_late_enrollment_enroll_by_date_before_now or is_course_run_end_date_before_now:
                 # skip old course runs
                 continue
         output.append(_get_course_run(full_course_run))
```
adamstankiewicz: [curious] Do we need to transform `full_course_run` with `_get_course_run` again here, given we already do it above via `this_course_run = _get_course_run(full_course_run)`?

Would it make sense to update this line to `output.append(this_course_run)`?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1100,6 +1107,10 @@ def _get_course_run(full_course_run):
         'max_effort': full_course_run.get('max_effort'),
         'weeks_to_complete': full_course_run.get('weeks_to_complete'),
         'upgrade_deadline': _get_verified_upgrade_deadline(full_course_run),
+        'enroll_by': _get_course_run_enroll_by_date_timestamp(full_course_run),
+        'is_active': _get_is_active_course_run(full_course_run),
+        'is_marketable': full_course_run.get('is_marketable'),
+        'status': full_course_run.get('status'),
```
adamstankiewicz: Ah, upon further reflection, we might want to reference where we landed with the filter defined for late enrollment in `getAvailableCourseRuns` within the Learner Portal ([source](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/17f426af2819734fa86c849b73d13807cf349c6f/src/components/app/data/utils.js#L519-L543)).

It's a good reminder that `is_marketable` also takes into account `status`, where an "unpublished" course run would have `is_marketable=False`; given this, we might not be able to use it as is. 

[suggestion/consideration] Perhaps we instead consider exposing `is_active` and a new boolean field `is_late_enrollment_eligible` (or similar).

```python
{
    'is_active': _get_is_active_course_run(full_course_run),
    'is_eligible_for_late_enrollment': _get_is_eligible_for_late_enrollment(full_course_run),
}

def _get_is_eligible_for_late_enrollment(course_run):
    # this could likely re-use the enroll_by offset logic used for filtering within `_get_course_runs`, perhaps extracting it to a helper function?
    enroll_by = course_run.get('enroll_by')
    if not enroll_by:
        return True
    return _is_enroll_by_eligible_for_enrollment(enroll_by)  # possible helper function
```

Then, I believe, consumers of these course run metadata could filter on `is_active=True` under normal circumstances and `is_active=True and is_eligible_for_late_enrollment=True` when opted into late enrollment (e.g., from the admin portal assignment flow) 🤔

enterprise_catalog/apps/catalog/algolia_utils.py
brobro10000: I would still argue yes its still applicable. The `is_course_run_active` field still uses `status` in some capacity to determine if a course run is active, so logging it downstream of its usage in determining the `is_active` is still valuable 👍🏽 

``` py
def is_course_run_active(course_run):
    """
    Checks whether a course run is active. That is, whether the course run is published,
    enrollable, and marketable.

    Arguments:
        course_run (dict): The metadata about a course run.

    Returns:
        bool: True if course run is "active"
    """
    course_run_status = course_run.get('status') or ''
    is_published = course_run_status.lower() == 'published'
    is_enrollable = course_run.get('is_enrollable', False)
    is_marketable = course_run.get('is_marketable', False)

    return is_published and is_enrollable and is_marketable
```

enterprise_catalog/apps/catalog/algolia_utils.py
brobro10000: opted for further refinement with the walrus symbol (`:=`) 

``` py
    for full_course_run in course_runs:
        this_course_run = _get_course_run(full_course_run)
        is_late_enrollment_enroll_by_date_before_now = False
        is_end_date_before_now = False
        if enroll_by := this_course_run.get('enroll_by'):
            course_run_enroll_by_datetime = datetime.datetime.fromtimestamp(enroll_by)
            # offsets enroll_by date by the late enrollment threshold
            course_run_late_enrollment_enroll_by_datetime = \
                course_run_enroll_by_datetime + datetime.timedelta(days=LATE_ENROLLMENT_THRESHOLD_DAYS)
            # check for runs within the late enrollment threshold
            is_late_enrollment_enroll_by_date_before_now = \
                course_run_late_enrollment_enroll_by_datetime.replace(tzinfo=UTC) < localized_utcnow()
        if end := this_course_run.get('end'):
            course_run_end = parser.parse(end)
            # check for runs within course run end date
            is_end_date_before_now = course_run_end < localized_utcnow()
        if is_late_enrollment_enroll_by_date_before_now or is_end_date_before_now:
            # skip old course runs
            continue
        output.append(this_course_run)
    return output
```

enterprise_catalog/apps/catalog/algolia_utils.py
brobro10000: To capture our conversation synchronously, were opting not to move forward with the `is_eligible_for_late_enrollment` due to the lag time between indexing and the enroll-by date.

Theoretically, a failed re-indexing could result in the unintended consequence of the `is_eligible_for_late_enrollment` being incorrect due to the indexing time being sooner then the consumer accessing the value. Any consumer of the `enroll_by` field would have to calculate the late enrollment eligibility at the time of usage.

Were also removing the `is_marketable` field from indexing.

enterprise_catalog/apps/catalog/algolia_utils.py
brobro10000: To capture our synchronous conversation, we determined that checking for a sooner enroll-by date based on offset is a no-op but does no harm to filter on that value to capture enroll-by dates within the threshold of late enrollment. Since the consumers of the enroll-by date would need to calculate whether they are within the late enrollment threshold regardless, the course run end date should capture those users beyond the late enrollment offset but within the course run end date if a reindexing were to fail and the data turned stale.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1135,12 +1145,24 @@ def get_course_runs(course):
...
+                course_run_enroll_by_datetime + datetime.timedelta(days=LATE_ENROLLMENT_THRESHOLD_DAYS)
+            # check for runs within the late enrollment threshold
+            is_late_enrollment_enroll_by_date_before_now = \
+                course_run_late_enrollment_enroll_by_datetime.replace(tzinfo=UTC) < localized_utcnow()
+        if end := this_course_run.get('end'):
+            course_run_end = parser.parse(end)
+            # check for runs within course run end date
+            is_end_date_before_now = course_run_end < localized_utcnow()
+        if is_late_enrollment_enroll_by_date_before_now or is_end_date_before_now:
+            # skip old course runs
```
adamstankiewicz: nit: this is now skipping _either_ old course runs OR enroll-by date (with late enrollment offset) elapsed, so the comment could probably be updated?

```python
# skip course runs which have ended or elapsed their enroll_by date with late enrollment offset days
```

enterprise_catalog/apps/catalog/algolia_utils.py
adamstankiewicz: Ah, good point about the child method `is_course_run_active`; you're correct!

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1099,7 +1106,10 @@ def _get_course_run(full_course_run):
         'min_effort': full_course_run.get('min_effort'),
         'max_effort': full_course_run.get('max_effort'),
         'weeks_to_complete': full_course_run.get('weeks_to_complete'),
-        'upgrade_deadline': _get_verified_upgrade_deadline(full_course_run),
+        'upgrade_deadline': _get_verified_upgrade_deadline(full_course_run),  # deprecated in favor of `enroll_by`
+        'enroll_by': _get_course_run_enroll_by_date_timestamp(full_course_run),
+        'is_active': _get_is_active_course_run(full_course_run),
+        'status': full_course_run.get('status'),
```
adamstankiewicz: I'm still a bit uncertain whether we truly need `status` to support late enrollment in the Admin Portal? It's also no longer included in the CSV columns anymore either given the filtering in `course_hit_runs`.

[thought process] The default behavior for the Admin Portal reading these `course_runs` will filter only to show runs with `is_active=True`; when late enrollment is enabled for a budget, the Admin Portal will consider all course runs regardless of `is_active`.

If we did nothing else (i.e., no other filtering) in Admin Portal for this latter case, we would be at least guaranteed to only show runs within the late enrollment threshold at the time of last Algolia indexing (happy path, fresh data daily). Then, we'd also have the Admin Portal filtering to sanity check the current timestamp against `(enroll_by + late enrollment)`, ensuring we only show runs within that late enrollment 30-day offset for the edge case of relying on potentially stale Algolia data (e.g., a failing indexing job the past week for some reason).

[question/sanity check] At what point will the Admin Portal logic need to read the `status` (i.e., published vs. unpublished) here? I can see the Admin Portal needing to ensure `availability != 'Archived'`, but still a bit unsure where `status` would be consumed.

adamstankiewicz (commented):
Looks great; mostly left a comment/question for your consideration about `status` before ✅.

adamstankiewicz (approved):

PR #916, opened 2024-08-21
feat: more restricted runs and contains_content_items logic

iloveagent57:
* contains_content_items endpoint now both utilize the same shared logic around restricted runs
    * contains_content_items top-level key returns true based on explicit content membership *or*
    membership in configured restricted runs mapping
    * The previous point implies that requests about restricted course runs
    will response with contains_content_items=true, even if the run is
    not tied directly to the catalog via ContentMetadata records.
    ENT-9386
## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_enterprise_customer_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_enterprise_customer_views.py
@@ -367,6 +367,12 @@ class EnterpriseCustomerViewSetTests(APITestMixin):
         # Create a two catalogs that have the content we're looking for
         content_key = 'fake-key+101x'
         second_catalog = EnterpriseCatalogFactory(enterprise_uuid=self.enterprise_uuid)
+        # Add some non-null, but irrelevant restricted runs to this catalog
+        second_catalog.catalog_query.content_filter[RESTRICTED_RUNS_ALLOWED_KEY] = {
+            'something+else1': ['course-v1:something+else+restrictedrun']
+        }
+        second_catalog.catalog_query.save()
+
         relevant_content = ContentMetadataFactory(content_key=content_key)
         self.add_metadata_to_catalog(second_catalog, [relevant_content])
         url = self._get_contains_content_base_url() + '?course_run_ids=' + content_key + \
@@ -374,8 +380,10 @@ class EnterpriseCustomerViewSetTests(APITestMixin):
         self.assert_correct_contains_response(url, True)
 
         response = self.client.get(url)
-        catalog_list = response.json()['catalog_list']
+        response_payload = response.json()
+        catalog_list = response_payload['catalog_list']
         assert set(catalog_list) == {str(second_catalog.uuid)}
+        self.assertIsNone(response_payload['restricted_runs_allowed'])
 
     def test_contains_catalog_key_restricted_runs_allowed(self):
         """
@@ -408,12 +416,16 @@ class EnterpriseCustomerViewSetTests(APITestMixin):
         # make sure to also request a course key that has no restricted runs,
         # and then assert that it is *not* included in the response payload.
         url = self._get_contains_content_base_url() + \
-            '?course_run_ids=org+key1&course_run_ids=org+key2&course_run_ids=org+key3'
+            '?course_run_ids=org+key1&course_run_ids=org+key2&course_run_ids=org+key3&get_catalog_list=true'
 
         response = self.client.get(url)
         response_payload = response.json()
 
         self.assertTrue(response_payload.get('contains_content_items'))
+        self.assertEqual(
+            set(response_payload['catalog_list']),
+            set([str(catalog.uuid), str(other_catalog.uuid)])
+        )
         self.assertEqual(
             response_payload['restricted_runs_allowed'],
             {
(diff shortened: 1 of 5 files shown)


enterprise_catalog/apps/catalog/api.py
```diff
@@ -0,0 +1,68 @@
...
+        if not catalog.restricted_runs_allowed:
+            continue
+        for restricted_course_key, restricted_runs in catalog.restricted_runs_allowed.items():
+            if restricted_course_key not in requested_course_keys:
+                continue
+            course_dict = serialized_data[restricted_course_key]
+            for course_run_key in restricted_runs:
+                run_dict = course_dict[course_run_key]
+                run_dict['catalog_uuids'].add(str(catalog.uuid))
+    return serialized_data or None
```
iloveagent57: No implementation changed here.

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -106,13 +108,21 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
+            EnterpriseCatalog.objects.select_related(
+                'catalog_query',
+            ).filter(
+                enterprise_uuid=enterprise_uuid,
+            ))
 
         any_catalog_contains_content_items = False
         catalogs_that_contain_course = []
         for catalog in customer_catalogs:
             contains_content_items = catalog.contains_content_keys(course_run_ids + program_uuids)
```
iloveagent57: BTW, I believe `EnterpriseCatalog.contains_content_keys()` implicitly does an ANY type of query, not an ALL type of query.

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -106,13 +108,21 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
-            if contains_content_items:
+            contains_requested_restricted_items = catalog_contains_any_restricted_course_run(
+                catalog, course_run_ids,
+            )
+            if contains_content_items or contains_requested_restricted_items:
```
iloveagent57, line 124: ...hence why I'm using an `or` here and not `and`.

enterprise_catalog/apps/catalog/api.py
pwnage101: Why did I review this code in a separate PR and you're adding it again?

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -125,31 +135,11 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
         if (get_catalogs_containing_specified_content_ids or get_catalog_list):
             response_data['catalog_list'] = catalogs_that_contain_course
 
-        response_data[RESTRICTED_RUNS_ALLOWED_KEY] = self._get_restricted_runs_allowed_for_query(
-            course_run_ids, customer_catalogs,
+        response_data[RESTRICTED_RUNS_ALLOWED_KEY] = get_restricted_runs_allowed_for_query(
+            course_run_ids, customer_catalogs
```
pwnage101: I think after some more rubber ducking today, I think we might want to pass `catalogs_that_contain_course` to `get_restricted_runs_allowed_for_query()` instead of all of the customer's catalogs. This avoids needing to cover an edge case in the frontend where the parent course does not match the base query, but one of its restricted runs does.

We would use help texts and documentation to guide ECS to only write custom filters that match BOTH the course and any of its restricted runs. I believe this change would serve as a safeguard to prevent queries that might not look how we want them to (but essentially work by accident) from proliferating and making it hard to undo. Always easier to release something more restrictive and relax restrictions later, rather than the opposite.

```suggestion
        response_data[RESTRICTED_RUNS_ALLOWED_KEY] = get_restricted_runs_allowed_for_query(
            course_run_ids, catalogs_that_contain_course
```
(this assumes we update the `catalogs_that_contain_course.append()` line to add catalogs instead of catalog UUIDs)

pwnage101 (changes requested):

pwnage101 (commented):
The rest LGTM!

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -106,50 +108,33 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
             )
-        customer_catalogs = list(EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid))
+        customer_catalogs = list(
+            EnterpriseCatalog.objects.select_related(
+                'catalog_query',
+            ).filter(
+                enterprise_uuid=enterprise_uuid,
+            ))
 
-        any_catalog_contains_content_items = False
```
adamstankiewicz, line 111: [kudos] Nice cleanup of `any_catalog_contains_content_items`, and only relying on the empty vs. non-empty `catalogs_that_contain_course`.

adamstankiewicz (approved):
LGTM, pending any test updates (happy to re-review once tests updated).

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -106,50 +108,33 @@ def contains_content_items(self, request, enterprise_uuid, course_run_ids, progr
...
         for catalog in customer_catalogs:
             contains_content_items = catalog.contains_content_keys(course_run_ids + program_uuids)
-            if contains_content_items:
-                any_catalog_contains_content_items = True
-                if not (get_catalogs_containing_specified_content_ids or get_catalog_list):
-                    # Break as soon as we find a catalog that contains the specified content
-                    break
-                catalogs_that_contain_course.append(catalog.uuid)
+            contains_requested_restricted_items = catalog_contains_any_restricted_course_run(
+                catalog, course_run_ids,
```
pwnage101: Nit: since these are not always course run IDs, should we take this opportunity to rename this variable to something more generic, like content_keys?

enterprise_catalog/apps/api/v1/tests/test_enterprise_customer_views.py
```diff
@@ -430,6 +430,62 @@ def test_contains_catalog_key_restricted_runs_allowed(self):
             }
         )
 
+    def test_restricted_course_disallowed_if_course_not_in_catalog(self):
```
pwnage101, line 445: I still feel like there's a test missing for the exact case I was thinking about.  Can you come up with a situation that returns contains_content_items = true, AND restricted_runs_allowed = None?

edit: sorry, give me a moment to imagine the scenario conditions...

pwnage101: nevermind, this test (`test_restricted_course_disallowed_if_course_not_in_catalog`) is exactly what I had in mind. I got thrown off by the fact that the other test sends a course run to the endpoint.

pwnage101 (approved):

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
pwnage101: even if just in the scope of this function (i.e. no change to endpoint signature)

enterprise_catalog/apps/catalog/api.py
iloveagent57: This is no longer true - implementation has now changed to take both courses and course runs into account.

PR #917, opened 2024-08-22
feat: Update CatalogWorkbookView course run headers

brobro10000:
Upgrade CatalogWorkbookView `export_utils` to include net new headers for course runs corresponding to addtional fields being indexed in Algolia.

Corresponding [PR](https://github.com/openedx/enterprise-catalog/pull/915)https://github.com/openedx/enterprise-catalog/pull/915 related to net new Algolia fields.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/export_utils.py
+++ b/enterprise_catalog/apps/api/v1/export_utils.py
@@ -5,7 +5,6 @@ import math
 from dateutil import parser
 from django.utils.html import strip_tags
 
-from enterprise_catalog.apps.api.constants import LATE_ENROLLMENT_THRESHOLD_DAYS
 from enterprise_catalog.apps.catalog.algolia_utils import ALGOLIA_INDEX_SETTINGS
 
 
@@ -18,6 +17,7 @@ CSV_COURSE_HEADERS = [
     'Start',
     'End',
     'Verified Upgrade Deadline',
+    'Enroll-by Date'
     'Program Type',
     'Program Name',
     'Pacing',
@@ -58,7 +58,6 @@ CSV_COURSE_RUN_HEADERS = [
     'End Date',
     'Verified Upgrade Deadline',
     'Enroll-by Date',
-    'Late Enrollment Eligible',
     'Min Effort',
     'Max Effort',
     'Length',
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -345,6 +355,12 @@ def course_run_to_row(hit, course_run):
         upgrade_deadline = datetime.datetime.fromtimestamp(raw_deadline).strftime(DATE_FORMAT)
```
adamstankiewicz, line 352: nit: consider adding a `# deprecated in favor of enroll_by` type comment for this conditional block

enterprise_catalog/apps/api/v1/export_utils.py
```diff
@@ -345,6 +356,17 @@ def course_run_to_row(hit, course_run):
...
+    enroll_by = None
+    late_enrollment_eligible = None
+    if course_run.get('enroll_by'):
+        raw_enroll_by_timestamp = course_run.get('enroll_by')
+        enroll_by_dt = datetime.datetime.fromtimestamp(raw_enroll_by_timestamp)
+        enroll_by_late_enrollment_offset = enroll_by_dt + datetime.timedelta(days=LATE_ENROLLMENT_THRESHOLD_DAYS)
+        late_enrollment_eligible = enroll_by_late_enrollment_offset < datetime.datetime.now()
+        enroll_by = enroll_by_dt.strftime(DATE_FORMAT)
+    csv_row.append(enroll_by)
+    csv_row.append(late_enrollment_eligible)
```
adamstankiewicz: While `late_enrollment_eligible` is accurate at the time of calling this API, I'm worried only including the boolean against the download timestamp (i.e., `datetime.datetime.now()`) could be misleading when reading an already downloaded CSV report.

For example, if the CSV is downloaded when late enrollment is eligible, and then the user re-opens the CSV 4 days later and still sees "late enrollment is eligible", despite the 30-day offset elapsing within that 4 day window. I think you may have also mentioned something to this affect during the pairing session? 🤔 

If we do keep anything about a late enrollment deadline in the CSV, I would include the actual formatted timestamp string versus the boolean.

All this said, I'm still not 100% sure we should include it, given it's not relevant to most customers / use cases and could cause more confusion than not (e.g., given its not relevant to most customers, will they understand what is meant by "late enrollment" and when its applicable? 

[suggestion] Given the uncertainty, perhaps we ship without the column for late enrollment enroll by and add it if/when there's customer/support/sales feedback around misleading enroll by dates due to late enrollment.

Thoughts?

adamstankiewicz (commented):
Looking good; one comment/thought about the `late_enrollment_eligible` column for consideration before the ✅.

Also, curious... are there no tests related to `CatalogWorkbookView`?

adamstankiewicz (approved):

enterprise_catalog/apps/api/v1/export_utils.py
brobro10000: Makes sense. Going with removing the `late_enrollment_eligible` Header with the understanding that we are accounting for the normal enrollment flow where-as late enrollment is not a typical enrollment path you could encounter.

adamstankiewicz (approved):
LGTM

PR #930, opened 2024-09-06
feat: support course run keys in `get_content_metadata`; introduce `normalized_metadata_by_run` in persisted `ContentMetadata` course JSON

adamstankiewicz:
## Description

Related tickets:
* https://2u-internal.atlassian.net/browse/ENT-8876
* https://2u-internal.atlassian.net/browse/ENT-9029

Two primary changes:
1. Updates existing `get_content_metadata` API to support mixed content types in its optional `?content_keys` query parameter, i.e. both course and (with this PR) course run keys. Similar to other content metadata related APIs (e.g., in enterprise-subsidy), passing a course run key will typically return the top-level course.
    * This change was abstracted to its own PR: https://github.com/openedx/enterprise-catalog/pull/940

2. Extends the _full_ `json_metadata` of `ContentMetadata` course objects to include a new `normalized_metadata_by_run` key with a value containing a mapping of normalized metadata per course run key of the top-level course, using the same `NormalizedContentMetadataSerializerTests`.
   * This simplifies consumers of enterprise-catalog APIs such that run-specific dates, prices, etc. across content types doesn't need to occur at the edges. Instead, the interface is created at the time of content ingestion via `update_content_metadata`.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -288,21 +288,21 @@ def _update_full_content_metadata_course(content_keys, dry_run=False):
             # Merge the full metadata from discovery's /api/v1/courses into the local metadata object.
             metadata_record.json_metadata.update(course_metadata_dict)
 
-            # Exec ed provides the start/end dates in additional_metadata, so we should copy those over to the keys that
-            # we use (inside the advertised course run).
-            if metadata_record.is_exec_ed_2u_course:
-                json_meta = metadata_record.json_metadata
-                start_date = json_meta.get('additional_metadata', {}).get('start_date')
-                end_date = json_meta.get('additional_metadata', {}).get('end_date')
-                course_run_uuid = json_meta.get('advertised_course_run_uuid')
-                for run in json_meta.get('course_runs', []):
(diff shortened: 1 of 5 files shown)


enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
```diff
@@ -52,7 +52,7 @@ def get_queryset(self, **kwargs):
         queryset = self.enterprise_catalog.content_metadata
         content_filter = kwargs.get('content_keys_filter')
         if content_filter:
-            queryset = queryset.filter(content_key__in=content_filter)
+            queryset = self.enterprise_catalog.get_matching_content(content_keys=content_filter)
```
adamstankiewicz: [inform] This change allows the `get_content_metadata` API to support both top-level course keys and course run keys. Previously, passing course run key would usually return an empty list of `results`.

iloveagent57: [note] we talked offline about pulling this change into its own PR.

iloveagent57 (approved):
GTG once that one line is pulled into another PR.

PR #939, opened 2024-09-09
feat: fetch and filter out restricted course runs

iloveagent57:
* When enabled, fetches restricted b2b runs during syncing of full course metadata from discovery.
* Subsequently prevents restricted runs from being written to ContentMetadata.json_metadata before saving.
* In a future PR, we'll make use of the pruned restricted runs on a per-customer-catalog fashion.

### Set up a restricted course run locally
It'll be associated with DemoX: https://2u-internal.atlassian.net/wiki/spaces/SOL/pages/1317306375/Setting+up+restricted+runs+locally
### Enable the setting
In your `private.py` settings file, set `SHOULD_FETCH_RESTRICTED_COURSE_RUNS = True`
### Sync full course metadata
1. `make worker-restart worker-logs`
2. In a different shell, do `make app-shell` and then `./manage.py update_full_content_metadata --force`
3. Observe in the worker logs a line like `[restricted runs] Course edX+DemoX has restricted runs {'course-v1:edX+DemoX+restricted_run'} that will be removed.`
### Verify non-presence of restricted run in API
Fetch content metadata for the DemoX course in enterprise-catalog, i.e.
```
curl --location 'http://localhost:18160/api/v1/enterprise-catalogs/7467c9d2-433c-4f7e-ba2e-c5c7798527b2/get_content_metadata/?content_keys=edX%2BDemoX&traverse_pagination=True' \
--header 'Authorization: JWT [your jwt]' \
```
Observe that the restricted run is not included in the payload.

ENT-9404
## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/content_metadata_utils.py
+++ b/enterprise_catalog/apps/catalog/content_metadata_utils.py
@@ -6,13 +6,16 @@ from logging import getLogger
 
 from enterprise_catalog.apps.catalog.utils import get_content_key
 
-from .constants import FORCE_INCLUSION_METADATA_TAG_KEY
+from .constants import (
+    COURSE_RUN_RESTRICTION_TYPE_KEY,
+    FORCE_INCLUSION_METADATA_TAG_KEY,
+)
 
 
 LOGGER = getLogger(__name__)
 
 
-def tansform_force_included_courses(courses):
+def transform_force_included_courses(courses):
     """
     Transform a list of forced/unlisted course metadata
     ENT-8212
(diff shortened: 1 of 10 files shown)


enterprise_catalog/apps/catalog/content_metadata_utils.py
```diff
@@ -41,3 +44,124 @@ def transform_course_metadata_to_visible(course_metadata):
...
+
+    course_json_metadata['course_runs'] = non_restricted_runs
+    course_json_metadata['course_run_keys'] = non_restricted_keys
+    course_json_metadata['course_run_statuses'] = sorted(list(non_restricted_statuses))
+
+    # also recompute the first enrollable paid seat price for the course
+    # now that we've removed any restricted runs
+    course_json_metadata['first_enrollable_paid_seat_price'] = get_course_first_paid_enrollable_seat_price(
+        course_json_metadata,
+    )
```
pwnage101, line 167: It's not clear to me that we currently store the `first_enrollable_paid_seat_price` at the top-level of the course metadata, or ever read it again. Even the updated `algolia_utils.py` in this PR isn't updated to read this stored value, still relying on the same function call. Are there any future plans to read this key?

pwnage101:
The rest LGTM!

enterprise_catalog/apps/catalog/content_metadata_utils.py
iloveagent57: I looked into this since it was mentioned in the ticket description. I'm fairly sure it's exposed at the course level via the ES results from course-discovery:
https://github.com/openedx/course-discovery/blob/7e6e7376cf0893450ea38b2071a6c6a590eb61a8/course_discovery/apps/course_metadata/search_indexes/documents/course.py#L45
https://github.com/openedx/course-discovery/blob/34355f412fe2a039fdf4479269f5bd6c01ade18d/course_discovery/apps/course_metadata/models.py#L1736
I'm not sure if `algolia_utils.py` should just read this field or not. Maybe it can lazily use this function if the course-level value for this key is falsey?

pwnage101: Ah, so we inherit it but just don't touch it or even read it, if I'm understanding this correctly.  Our algolia indexing code just re-calculates it anyway based on the runs.

Best I can tell from reading the Discovery code, this field value is tainted by any restricted runs that exist for the course, regardless of whatever query parameter we pass to Discovery APIs.  Your code essentially "fixes" it to revert it to be restricted-run-unaware.  I think that's good for now, in case anything downstream does ever read it.  Your suggestion to update `algolia_utils.py` to use the field and fallback to the function sounds reasonable.

pwnage101:
Hm, in the process of drafting an approach for [ENT-9405](https://2u-internal.atlassian.net/browse/ENT-9405) (draft PR: https://github.com/openedx/enterprise-catalog/pull/947) it occurs to me that this current PR likely does not actually store restricted runs into the database as "orphan" ContentMetadata records because it only updates the call to /courses to get the full metadata, but an update to the /search/all call is missing.

We definitely need both to be updated, so the changes in this PR are still necessary, but just FYI my PR introduces new API calls to fetch the restricted runs from /search/all and stores them as ContentMetadata records.  Just so you know so we don't step over toes!

PR #940, opened 2024-09-09
feat: expand `get_content_metadata` REST API to support course run keys

adamstankiewicz:
When a catalog containing only courses is used with `get_content_metadata`, passing _course run_ keys to its `?content_keys` filter, the API will return the associated top-level course records instead of an empty result set.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -1394,6 +1394,33 @@ class EnterpriseCatalogGetContentMetadataTests(APITestMixin):
         self.assertEqual(response.json()['results'], [])
 
     @mock.patch('enterprise_catalog.apps.api_client.enterprise_cache.EnterpriseApiClient')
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -1393,6 +1393,33 @@ def test_get_content_metadata_no_catalog_query(self):
         self.assertEqual(response.status_code, status.HTTP_200_OK)
         self.assertEqual(response.json()['results'], [])
 
+    @mock.patch('enterprise_catalog.apps.api_client.enterprise_cache.EnterpriseApiClient')
+    def test_get_content_metadata_content_filters_course_run_Key(self, mock_api_client):
```
iloveagent57: nit: capital 'K' in "Key"

iloveagent57 (approved):

enterprise_catalog/apps/api/v1/tests/test_views.py
adamstankiewicz: ha! good catch 👀

PR #943, opened 2024-09-16
feat: include `content_price` in Algolia indexed `course_runs` for course objects

adamstankiewicz:
[ENT-9504](https://2u-internal.atlassian.net/browse/ENT-9504)

* Ensures the `course_runs` list indexed in Algolia course objects contains `content_price`.
* Refactors to pull `enroll_by` for the indexed `course_runs` from `NormalizedContentMetadataSerializer` vs. largely replicating the same logic in multiple places (DRY).

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/serializers.py
+++ b/enterprise_catalog/apps/catalog/serializers.py
@@ -9,8 +9,8 @@ from drf_spectacular.utils import extend_schema_field
 from rest_framework import serializers
 
 from enterprise_catalog.apps.api.constants import CourseMode
-
-from .algolia_utils import _get_course_run_by_uuid
+from enterprise_catalog.apps.catalog.constants import EXEC_ED_2U_COURSE_TYPE
+from enterprise_catalog.apps.catalog.utils import get_course_run_by_uuid
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -25,11 +25,15 @@
 PAST_COURSE_RUN_UUID_1 = uuid4()
 
 
-def _fake_upgrade_deadline(days_from_now=0):
+def _days_from_now(days_from_now=0):
```
adamstankiewicz, line 28: [inform] Rename carried over from https://github.com/openedx/enterprise-catalog/pull/912

enterprise_catalog/apps/catalog/serializers.py
```diff
@@ -99,20 +99,19 @@ class NormalizedContentMetadataSerializer(ReadOnlySerializer):
...
-    def course(self):
-        return self.instance.get('course')
+    def course_metadata(self):
+        return self.instance.get('course_metadata')
 
     @cached_property
-    def course_metadata(self):
-        return self.course.json_metadata
+    def is_exec_ed_2u_course(self):
+        return self.course_metadata.get('course_type') == EXEC_ED_2U_COURSE_TYPE
```
adamstankiewicz, line 107: [inform] To support using `NormalizedContentMetadataSerializer` where only the _metadata_ for course is available vs. the actual database `ContentMetadata` object itself, consumers must pass in `course_metadata: course_record.json_metadata` instead of having the serializer attempt to parse `json_metadata`.

As a result, the helper property `is_exec_ed_2u_course` that exists on the `ContentMetadata` model is accounted for here as well such that the `get_content_price` can still utilize it.

brobro10000 (approved):
LGTM 👍🏽

PR #945, opened 2024-09-17
fix: ensure late enrollment eligibility is correct in Algolia indexing of runs

adamstankiewicz:
1. Ensures we include eligible `course_runs` in the Algolia index for top-level course objects, with a fix related to late enrollment eligibility. Excludes run(s) that:
    * Has ended
    * Has an `enroll_by` in the past, but it's outside of the 30-day late enrollment buffer window.
1. Updates `enroll_by` to allow `null` values in Algolia, but also exposes a `has_enroll_by` boolean, per [Algolia docs](https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-attributes/#filter-by-null-or-missing-attributes) for supporting filtering with `null` / missing attributes.
1. Includes `is_late_enrollment_eligible` boolean field in the `_get_course_run` payload to provide an single, abstracted property for clients (e.g., Admin Portal) to consume when conditionally supporting late enrollment.

## Post-review


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1120,7 +1122,8 @@ def _get_course_run(course, course_run):
         'max_effort': course_run.get('max_effort'),
         'weeks_to_complete': course_run.get('weeks_to_complete'),
         'upgrade_deadline': _get_verified_upgrade_deadline(course_run),  # deprecated in favor of `enroll_by`
-        'enroll_by': _get_course_run_enroll_by_date_timestamp(normalized_content_metadata),
+        'enroll_by': enroll_by,
+        'has_enroll_by': bool(enroll_by),
```
adamstankiewicz, line 1126: [inform] No longer falls back to the `ALGOLIA_DEFAULT_TIMESTAMP` (year 3000), as was the case with `upgrade_deadline`. See Algolia docs for more context/details: https://www.algolia.com/doc/guides/managing-results/refine-results/filtering/how-to/filter-by-attributes/#filter-by-null-or-missing-attributes

This ensures we can do proper checks against runs that have no `enroll_by` date (e.g., as is the case on a lot of stage courses).

brobro10000 (approved):
Approving to unblock 👍🏽  Great job tackling this, thank you!

pwnage101 (approved):

PR #947, opened 2024-09-19
Relate restricted runs to CatalogQuery objects (ENT-9405)

pwnage101:
Two main parts of this:
1. Pre-fetch restricted runs from discovery as needed. This is required because they might not yet be imported into the database at the time we are processing a CatalogQuery which allows them, but we need them imported at that point in time so that we can relate them to the CatalogQuery.
2. Actually create a new M2M relationship between ContentMetadata and CatalogQuery, and relate restricted runs to queries as defined in the content filter (original goal of this ticket).

ENT-9405


enterprise_catalog/apps/catalog/models.py
```diff
@@ -629,6 +630,22 @@ class ContentMetadata(TimeStampedModel):
...
+    # `catalog_queries` field above generally doesn't relate course runs
+    # directly to CatalogQuery objects. As a result, downstream code broadly
+    # assumes that a run belongs to a CatalogQuery to which its parent course
+    # also belongs. Relating runs directly to CatalogQueries is a fundamental
+    # departure from the original architecture and base assumptions, hence the
+    # new field just to be on the safe side.
+    catalog_queries_for_restricted_course_run = models.ManyToManyField(
+        CatalogQuery,
+        related_name='restricted_course_runs',
+    )
```
iloveagent57, line 647: Can we create an explicit M2M model to represent this and use `through`?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1129,37 +1209,93 @@ def update_contentmetadata_from_discovery(catalog_query, dry_run=False):
...
+        restricted_metadata | resticted_run_metadata,
+        catalog_query,
+        dry_run
+    )
+    LOGGER.info(
+        'Associated %d restricted runs (%d unique) with catalog query %s',
+        len(associated_restricted_run_keys),
+        len(set(associated_restricted_run_keys)),
+        catalog_query,
+    )
```
iloveagent57, line 1294: Maybe throw this whole chunk into a new function?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1129,37 +1209,93 @@ def update_contentmetadata_from_discovery(catalog_query, dry_run=False):
...
-            catalog_query,
-        )
+    # Partition metadata dicts by unrestricted vs. restricted.
+    restricted_metadata = []
+    unrestricted_metadata = []
+    for metadata_dict in metadata:
+        if is_content_restricted(metadata_dict):
+            restricted_metadata += metadata_dict
+        else:
+            unrestricted_metadata += metadata_dict
```
iloveagent57, line 1238: Is it necessary to mix the restricted run fetching in with the unrestricted fetching? Since we're going to make a second set of requests to discovery anyway, can we do as just two completely separate chunks of work?
1. Request the content filter without requesting restricted runs.
2. Request _only_ the restricted runs defined in the content filter.

alexjmpb:
@iloveagent57 @pwnage101 Greetings! just reaching out to confirm if we can close this draft since this was already implemented on https://github.com/openedx/enterprise-catalog/pull/965.

PR #949, opened 2024-09-24
feat: index enrollStart and hasEnrollStart in Algolia course_runs

adamstankiewicz:
* Includes `enroll_start_date` in the `NormalizedContentMetadataSerializer`. This will persist this field in the `json_metadata` blob for course records.
* Includes `enroll_start` in the indexed course runs metadata for `course` objects within Algolia; this will primarily be consumed by Admin Portal during the assignment allocation flow.

## Post-review


enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -517,11 +529,12 @@ def test_get_upcoming_course_runs(self, searchable_course, expected_course_runs)
                         'max_effort': 6,
                         'weeks_to_complete': 6,
                         'seats': [{
-                            'type': 'verified',
```
adamstankiewicz, line 520: [inform] Updates the course run type to be Exec Ed when dealing with the late enrollment cases, as late enrollment is typically used with Exec Ed content.

brobro10000 (approved):
Smart idea to serialize this value as part of normalized metadata. Approving to unblock, but I'd also have someone else take a look at this as well, just as a sanity check 👍🏽

iloveagent57 (approved):

PR #952, opened 2024-09-26
spike: [nomerge] spike on restricted run modeling

iloveagent57:
## Post-review


enterprise_catalog/apps/catalog/migrations/0041_spike_part_two.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 4.2.15 on 2024-09-26 19:02
+
+import collections
+from django.db import migrations
+import jsonfield.encoder
+import jsonfield.fields
+
+
+class Migration(migrations.Migration):
```
iloveagent57, line 9: The nice thing about this migration is its a no-op in terms of SQL:
```
root@app:/edx/app/enterprise_catalog/enterprise_catalog# ./manage.py sqlmigrate catalog 0041
--
-- Alter field data on book
--
-- (no-op)
--
-- Rename field data on book to _data
--
-- (no-op)
--
-- Alter field data on restrictedbook
--
-- (no-op)
--
-- Rename field data on restrictedbook to _data
--
-- (no-op)
```

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1197,3 +1197,113 @@ def current_options(cls):
...
+        null=True,
+        load_kwargs={'object_pairs_hook': collections.OrderedDict},
+        dump_kwargs={'indent': 4, 'cls': JSONEncoder, 'separators': (',', ':')},
+    )
+
+    @property
+    def data(self):
+        if overwritten_copy := getattr(self, 'overwritten_copy', None):
+            return overwritten_copy[0]._data
+        return self._data
```
iloveagent57, line 1272: We'd do a similar thing on the existing `CatalogQuery` model, renaming `json_metadata` to `_json_metadata`, but leaving the `db_column='json_metadata'`, so that there's no change to the DB schema.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1197,3 +1197,113 @@ def current_options(cls):
...
+        unique=True,
+    )
+
+    @property
+    def books(self):
+        prefetch_qs = models.Prefetch(
+            'restricted_books',
+            queryset=RestrictedBook.objects.filter(library=self),
+            to_attr='overwritten_copy',
+        )
```
iloveagent57, line 1221: This prefetch is an extra query, but we can avoid it in all cases where a `CatalogQuery` doesn't specify restricted runs to fetch in its content filter.  We'll make use of `overwritten_copy` in the `data` property below.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1197,3 +1197,113 @@ def current_options(cls):
...
+        blank=True,
+        null=True,
+        load_kwargs={'object_pairs_hook': collections.OrderedDict},
+        dump_kwargs={'indent': 4, 'cls': JSONEncoder, 'separators': (',', ':')},
+    )
+
+    @property
+    def data(self):
+        if overwritten_copy := getattr(self, 'overwritten_copy', None):
+            return overwritten_copy[0]._data
```
iloveagent57, line 1271: because of prefetch, this will always be a list, but because of how we _filtered_ the prefetch, it should only ever include 1 element (or zero if the filter returns no results).

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1197,3 +1197,113 @@ def current_options(cls):
                 'no_async': current_config.no_async,
             }
         return {}
+
+
+
+class Library(models.Model):
```
iloveagent57, line 1204: analogous to `CatalogQuery`

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1197,3 +1197,113 @@ def current_options(cls):
...
+            queryset=RestrictedBook.objects.filter(library=self),
+            to_attr='overwritten_copy',
+        )
+        return self.book_set.prefetch_related(prefetch_qs)
+
+    def __str__(self):
+        return self.name
+
+
+class AbstractBook(models.Model):
```
iloveagent57, line 1248: we'd similarly extract an `AbstractContentMetadata` class.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1197,3 +1197,113 @@ def current_options(cls):
...
+    )
+
+    @property
+    def data(self):
+        if overwritten_copy := getattr(self, 'overwritten_copy', None):
+            return overwritten_copy[0]._data
+        return self._data
+
+
+class Book(AbstractBook):
```
iloveagent57, line 1275: analogous to `ContentMetadata`

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1197,3 +1197,113 @@ def current_options(cls):
...
+    class Meta:
+        app_label = 'catalog'
+
+    def __str__(self):
+        return f"title: {self.title}, data={self.data}"
+
+
+class RestrictedBook(AbstractBook):
```
iloveagent57, line 1283: analagous to `RestrictedCourseMetadata`

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1197,3 +1197,113 @@ def current_options(cls):
...
+        null=True,
+        related_name='restricted_books',
+        on_delete=models.deletion.CASCADE,
+    )
+
+    def save(self, *args, **kwargs):
+        if not self.id:
+            super().save(*args, **kwargs)
+
+        self.title = self.parent_book.title
```
iloveagent57, line 1309: force this to always inherit the parent book's title.  this would be analagous to always inheriting the parent content metadata's `content_key`.

PR #953, opened 2024-09-27
[do not merge] restricted runs modeling spike (again)

pwnage101:


enterprise_catalog/apps/catalog/migrations/0040_restrictedcoursemetadata_and_more.py
```diff
@@ -0,0 +1,58 @@
...
+            options={
+                'verbose_name': 'Restricted Content Metadata',
+                'verbose_name_plural': 'Restricted Content Metadata',
+            },
+        ),
+        migrations.RenameField(
+            model_name='contentmetadata',
+            old_name='json_metadata',
+            new_name='_json_metadata',
+        ),
```
pwnage101, line 40: This is what makes it so that the name of the field in the database doesn't actually change, but the reference to it in django is renamed.

alexjmpb:
Hi @pwnage101 hope that you are doing well. Can we close this PR? Is this an issue that we keep an eye on?

PR #958, opened 2024-09-30
feat: use django parse_datetime() instead of strptime()

iloveagent57:
Use this django utility that accepts strings in valid ISO 8601 datetime formats (or some close alternative) and returns a datetime object. https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.dateparse.parse_datetime

## Post-review


pwnage101 (approved):
leave it to the pros

brobro10000 (approved):
+1 on the pros :) great job

adamstankiewicz (approved):

PR #959, opened 2024-10-03
feat: Added models and commands for enterprise-jobs association

irfanuddinahmad:
**Description**: This PR adds a new app 'jobs' with commands and models for persisting the jobs associated with an enterprise. The relationship is established via common skills between the job and the skills associated with courses in the Enterprises's catalog.

**JIRA**: ENT-9469, ENT-9470

## Post-review

* [*] Squash commits into discrete sets of changes


enterprise_catalog/apps/jobs/models.py
```diff
@@ -0,0 +1,114 @@
...
+    .. no_pii:
+    """
+    job = models.ForeignKey(
+        Job,
+        blank=True,
+        null=True,
+        related_name='enterprises',
+        on_delete=models.deletion.SET_NULL,
+    )
+    enterprise_uuid = models.UUIDField(primary_key=True, default=uuid4, editable=False)
```
mahamakifdar19: I may be mistaken, but I wanted to point out that since `JobEnterprise` serves as a junction table between `Job` and `Enterprise`, each `Enterprise` can be associated with multiple `Job` records through various entries in `JobEnterprise`. 
However, in the current `JobEnterprise` model, `enterprise_uuid` is defined as the primary key, which I believe would restrict the ability to have multiple entries for the same `Enterprise`. 

Perhaps we should consider modifying the `JobEnterprise` model to make `enterprise_uuid` a foreign key that references the `Enterprise` model.

```
id = models.UUIDField(primary_key=True, default=uuid4, editable=False)  # Separate primary key
enterprise = models.ForeignKey(Enterprise, on_delete=models.CASCADE, related_name='job_enterprises')
```

mahamakifdar19 (approved):

enterprise_catalog/apps/jobs/models.py
```diff
@@ -0,0 +1,114 @@
...
+
+    .. no_pii:
+    """
+    job = models.ForeignKey(
+        Job,
+        blank=True,
+        null=True,
+        related_name='skills',
+        on_delete=models.deletion.SET_NULL,
+    )
```
muhammad-ammar: We are setting  `blank=True` and `null=True`, which means the `JobSkill` object can exist without an associated `Job`. This design might allow orphaned `JobSkill` records.

enterprise_catalog/apps/jobs/models.py
```diff
@@ -0,0 +1,114 @@
...
+
+    .. no_pii:
+    """
+    job = models.ForeignKey(
+        Job,
+        blank=True,
+        null=True,
+        related_name='enterprises',
+        on_delete=models.deletion.SET_NULL,
+    )
```
muhammad-ammar: We are setting  `blank=True` and `null=True`, which means the `JobEnterprise` object can exist without an associated `Job`. This design might allow orphaned `JobEnterprise ` records.

enterprise_catalog/apps/jobs/management/commands/fetch_jobs_skills.py
```diff
@@ -0,0 +1,65 @@
...
+                        job=job,
+                        skill_id=skill.get('id'),
+                        name=skill.get('name'),
+                        significance=item.get('significance')
+                    )
+            except Exception as exc:  # pylint: disable=broad-exception-caught
+                job_id = result.get('id')
+                logger.exception(f'Could not store job skills. job id {job_id} {exc}')
+
+    def handle(self, *args, **options):
```
muhammad-ammar, line 51: **Questions:**

1. What happens if a job no longer exists in Discovery, but that job and its related data still exist in the Job and JobSkills models? Don't we need to delete that Job and its related data?
2. Do we need to fetch all jobs every time, or can we only fetch jobs that have been updated since the last fetch?

enterprise_catalog/apps/jobs/management/commands/process_enterprise_jobs.py
```diff
@@ -0,0 +1,67 @@
...
+                            enterprise_customer_uuids:{enterprise_uuid}',
+                        'maxFacetHits': 50
+                    }
+                    response = algolia_client.algolia_index.search_for_facet_values('skill_names', '', search_query)
+                    for hit in response.get('facetHits', []):
+                        if hit.get('count') > 1:
+                            JobEnterprise.objects.update_or_create(
+                                job=job,
+                                enterprise_uuid=enterprise_uuid
+                            )
```
muhammad-ammar, line 59: What happens if a job is no longer associated with an enterprise? Don't we need to delete Job and related JobEnterprise object?

enterprise_catalog/apps/jobs/management/commands/process_enterprise_jobs.py
```diff
@@ -0,0 +1,67 @@
...
+        enterprise_uuids = set()
+        enterprise_catalogs = EnterpriseCatalog.objects.all()
+        for enterprise_catalog in enterprise_catalogs:
+            enterprise_uuids.add(enterprise_catalog.enterprise_uuid)
+
+        jobs = Job.objects.all()
+        for enterprise_uuid in enterprise_uuids:
+            for job in jobs:
+                try:
+                    job_skills = JobSkill.objects.filter(job=job).order_by('-significance')[:3]  # Get top 3 skills
```
muhammad-ammar, line 45: Command name and docstings says that we are associating jobs with enterprises but this entire association is created based on skills. Can we update the docstrings and add more details?

enterprise_catalog/apps/jobs/management/commands/process_enterprise_jobs.py
```diff
@@ -0,0 +1,67 @@
...
+                            )
+                            break
+                except Exception as ex:  # pylint: disable=broad-exception-caught
+                    logger.error(
+                        '[PROCESS_ENTERPRISE_JOBS] Failure in processing \
+                        enterprise "%s" and job: "%s". Ex:  "%s".',
+                        enterprise_uuid,
+                        job.job_id,
+                        str(ex)
+                    )
```
muhammad-ammar: I think the log can be improved.

```python
logger.error(
        '[PROCESS_ENTERPRISE_JOBS] Failure in processing enterprise "%s" and job: "%s".',
        enterprise_uuid,
        job.job_id,
        exc_info=True
    )
```

enterprise_catalog/apps/jobs/models.py
```diff
@@ -0,0 +1,114 @@
...
+from simple_history.models import HistoricalRecords
+
+
+class Job(TimeStampedModel):
+    """
+    Model for storing job related information.
+
+    .. no_pii:
+    """
+    job_id = models.IntegerField(unique=True, help_text=_('The job ID received from API.'))
```
muhammad-ammar, line 16: Do we really need to store the job id? Cannot we only use the `external_id`?

enterprise_catalog/apps/jobs/management/commands/fetch_jobs_skills.py
```diff
@@ -0,0 +1,65 @@
...
+        try:
+            results, has_next = DiscoveryApiClient().get_jobs_skills(page=page)
+            self._process_job_skills(results)
+
+            while has_next:
+                page += 1
+                results, has_next = DiscoveryApiClient().get_jobs_skills(page=page)
+                self._process_job_skills(results)
+        except Exception as exc:  # pylint: disable=broad-exception-caught
+            logger.exception(f'Could not retrieve job skills for page {page} {exc}')
```
muhammad-ammar, line 65: Shouldn't we allow it to fail loudly? If an error occurs, no one will know that something went wrong.

enterprise_catalog/apps/jobs/management/commands/fetch_jobs_skills.py
```diff
@@ -0,0 +1,65 @@
...
+                    skill = item.get('skill')
+                    JobSkill.objects.update_or_create(
+                        job=job,
+                        skill_id=skill.get('id'),
+                        name=skill.get('name'),
+                        significance=item.get('significance')
+                    )
+            except Exception as exc:  # pylint: disable=broad-exception-caught
+                job_id = result.get('id')
+                logger.exception(f'Could not store job skills. job id {job_id} {exc}')
```
muhammad-ammar, line 49: Shouldn't we allow it to fail loudly? If an error occurs, no one will know that something went wrong.

enterprise_catalog/apps/jobs/management/commands/process_enterprise_jobs.py
```diff
@@ -0,0 +1,67 @@
...
+                    response = algolia_client.algolia_index.search_for_facet_values('skill_names', '', search_query)
+                    for hit in response.get('facetHits', []):
+                        if hit.get('count') > 1:
+                            JobEnterprise.objects.update_or_create(
+                                job=job,
+                                enterprise_uuid=enterprise_uuid
+                            )
+                            break
+                except Exception as ex:  # pylint: disable=broad-exception-caught
+                    logger.error(
```
muhammad-ammar, line 63: Shouldn't we allow it to fail loudly? If an error occurs, no one will know that something went wrong.

enterprise_catalog/apps/jobs/models.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __str__(self):
+        """
+        Return human-readable string representation.
+        """
+        return (
+            "<JobEnterprises for '{job_id}'>".format(
+                job_id=str(self.job)
+            )
+        )
```
muhammad-ammar: I think the above return a string something like below. Is this expected?

```python
<JobEnterprises for '<Job id="123">'>
```

enterprise_catalog/apps/jobs/models.py
```diff
@@ -0,0 +1,114 @@
...
+
+    def __str__(self):
+        """
+        Return human-readable string representation.
+        """
+        return (
+            "<JobSkills for '{job_id}'>".format(
+                job_id=str(self.job)
+            )
+        )
```
muhammad-ammar, line 78: I think the above return a string something like below. Is this expected?

```python
< JobSkills for '<Job id="123">'>
```

enterprise_catalog/apps/jobs/management/commands/fetch_jobs_skills.py
irfanuddinahmad: Good point. Something we can consider in a subsequent iteration

enterprise_catalog/apps/jobs/management/commands/fetch_jobs_skills.py
irfanuddinahmad: No. We donot want to stop because of an issue in one of the rows

enterprise_catalog/apps/jobs/management/commands/fetch_jobs_skills.py
irfanuddinahmad: No. We donot want to stop because of an issue in one of the rows

enterprise_catalog/apps/jobs/models.py
irfanuddinahmad: future proofing ...

PR #963, opened 2024-10-08
Data model for explicit storage of restricted runs and their courses

pwnage101:
ENT-9569


enterprise_catalog/apps/catalog/models.py
```diff
@@ -252,11 +252,42 @@ def content_metadata(self):
...
+        2. The ContentMetdata.json_metadata attribute being a property that
+           dynamically uses (1) or falls back to the stored value in _json_metadata.
+
+        Returns:
+            Queryset of ContentMetadata: Same as self.content_metadata, but courses may have augmented
+            json_metadata.
+        """
+        if not self.catalog_query:
+            return ContentMetadata.objects.none()
+        # FYI: prefetch causes a performance penalty by introducing a 2nd database query.
```
iloveagent57: Note: we can probably avoid this penalty by examining the related `catalog_query.content_filter` for the presence of the key that allows restricted runs.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -252,11 +252,42 @@ def content_metadata(self):
...
-            Queryset: The queryset of associated content metadata
+            Queryset of ContentMetadata: The queryset of associated content metadata
         """
         if not self.catalog_query:
             return ContentMetadata.objects.none()
-        return self.catalog_query.contentmetadata_set.all()
+        return self.catalog_query.contentmetadata_set.filter(
+            # Exclude any restricted runs.
+            restricted_courses_which_allow_restricted_run__isnull=True,
+        )
```
iloveagent57: 1. We probably want a `prefetch_related()` here, too, right?
2. Can we avoid dealing with restricted runs at all if the related `catalog_query.content_filter` doesn't ask for them?

enterprise_catalog/apps/catalog/migrations/0040_rename_json_metadata_contentmetadata__json_metadata_and_more.py
```diff
@@ -0,0 +1,85 @@
...
+        migrations.RenameField(
+            model_name='contentmetadata',
+            old_name='json_metadata',
+            new_name='_json_metadata',
+        ),
+        migrations.RenameField(
+            model_name='historicalcontentmetadata',
+            old_name='json_metadata',
+            new_name='_json_metadata',
+        ),
```
iloveagent57: We don't want to rename this field in the database.

pwnage101: Ah, these lines in the migration were actually correct but `sqlmigrate` generated a non-no-op (in other words, `migrate` would have renamed the DB field).  The solution was to set `db_column='json_metadata',` on the field definition, then `sqlmigrate` produces a no-op:

```
(enterprise-catalog) ➜  enterprise-catalog git:(pwnage101/ENT-9569) ✗ python manage.py makemigrations
[...]
Was contentmetadata.json_metadata renamed to contentmetadata._json_metadata (a JSONField)? [y/N] y
Was historicalcontentmetadata.json_metadata renamed to historicalcontentmetadata._json_metadata (a JSONField)? [y/N] y
Migrations for 'catalog':
  enterprise_catalog/apps/catalog/migrations/0040_alter_contentmetadata_json_metadata_and_more.py
    - Alter field json_metadata on contentmetadata
    - Rename field json_metadata on contentmetadata to _json_metadata
    - Alter field json_metadata on historicalcontentmetadata
    - Rename field json_metadata on historicalcontentmetadata to _json_metadata
    - Alter field content_filter on catalogquery
    - Create model HistoricalRestrictedCourseMetadata
    - Create model RestrictedCourseMetadata

(enterprise-catalog) ➜  enterprise-catalog git:(pwnage101/ENT-9569) ✗ python manage.py sqlmigrate catalog 0040
[...]
BEGIN;
--
-- Alter field json_metadata on contentmetadata
--
-- (no-op)
--
-- Rename field json_metadata on contentmetadata to _json_metadata
--
-- (no-op)
--
-- Alter field json_metadata on historicalcontentmetadata
--
-- (no-op)
--
-- Rename field json_metadata on historicalcontentmetadata to _json_metadata
--
-- (no-op)
--
[...]
COMMIT;
```

enterprise_catalog/apps/catalog/migrations/0040_restrictedcoursemetadata_and_more.py
```diff
@@ -0,0 +1,118 @@
...
+        migrations.RenameField(
+            model_name='historicalcontentmetadata',
+            old_name='json_metadata',
+            new_name='_json_metadata',
+        ),
+        migrations.AlterField(
+            model_name='catalogquery',
+            name='content_filter',
+            field=jsonfield.fields.JSONField(default=dict, dump_kwargs={'cls': jsonfield.encoder.JSONEncoder, 'ensure_ascii': False, 'indent': 4, 'separators': (',', ':')}, help_text="Query parameters which will be used to filter the discovery service's search/all endpoint results, specified as a JSON object.", load_kwargs={'object_pairs_hook': collections.OrderedDict}),
+        ),
```
pwnage101: why on earth is this being altered?

pwnage101: I'll rebase on alex's PR to fix this. https://github.com/openedx/enterprise-catalog/pull/967

enterprise_catalog/apps/catalog/tests/factories.py
```diff
@@ -187,6 +189,31 @@ def json_metadata(self):
...
+        model = RestrictedCourseMetadata
+
+    content_key = factory.Faker('bothify', text='??????????+####')
+    content_uuid = factory.LazyFunction(uuid4)
+    content_type = COURSE
+    parent_content_key = None
+    _json_metadata = {}  # Callers are encouraged to set this.
+
+
+class RestrictedRunAllowedForRestrictedCourseFactory(factory.django.DjangoModelFactory):
```
pmakwana93, line 206: Q: so a `content_key` does not need to be passed in when instantiating a RestrictedRunAllowedForRestrictedCourseFactory obj?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -1,3 +1,158 @@
+"""
```
iloveagent57: We can probably get rid of this, or put it under a top-level directory like `scripts/`.

enterprise_catalog/apps/catalog/tests/factories.py
iloveagent57: No, because the model doesn't have a content_key field, it has a FK field that points at a `RestrictedCourseMetadata` instance (which _itself_ has a `content_key` - this will be the course key).  Additionally, to represent the run, this model doesn't refer directly to a course key, but rather FKs to a `ContentMetadata` record for the run (which has a `content_key` field holding the course _run_ key).  This model is really just the "glue" between a restricted course metadata object and a content metadata object representing a restricted _run_ available in the context of the restricted course.

iloveagent57 (commented):
This is looking good! It's reassuring that the changes in test files are all either net-new tests, or swapping `json_metadata` for `_json_metadata`.  I'll try to run through your spike test on my machine today.

enterprise_catalog/apps/catalog/tests/factories.py
pmakwana93: Makes sense, thank you for the context

enterprise_catalog/apps/catalog/migrations/0041_restrictedcoursemetadata_and_more.py
```diff
@@ -0,0 +1,109 @@
...
+            ],
+            options={
+                'abstract': False,
+            },
+        ),
+        migrations.AddField(
+            model_name='restrictedcoursemetadata',
+            name='unrestricted_parent',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, related_name='restricted_courses', to='catalog.contentmetadata'),
+        ),
```
pwnage101, line 77: Why would django decide to introduce AddField to the exact same migration where the model is being introduced via CreateModel?

iloveagent57 (approved):

enterprise_catalog/apps/catalog/migrations/0041_restrictedcoursemetadata_and_more.py
iloveagent57: I think because it's a FK?

PR #967, opened 2024-10-09
fix: fix weird migration

iloveagent57:
1. Deletes commented out model field and related commented out unique index.
2. Changes catalog migration 0039 to not be conditional on the engine, I believe we used this as a hueristic to *not* apply something in stage/prod, but since we use mysql locally to generate migration files, its messing with the makemigrations if we don't change the previous migration.
3. Introduces a (no-op in terms of SQL) migration to alter the CatalogQuery.content_filter_field. AFAICT, the only real change here is that the JSONEncoder on the field now sets ensure_ascii to False in `dump_kwargs`.

See https://github.com/openedx/enterprise-catalog/pull/793 for source of migration 0039, which I believe is the source of the problem *this* PR tries to address.

## Post-review


pwnage101 (approved):
By chance have you run sqlmigrate from an app-shell to confirm the AlterField on CatalogQuery.content_filter is a no-op when the db engine is mysql?  Nevertheless, even if it does recreate the table I think the relatively small size of the table in prod would make the migration performant enough.

PR #968, opened 2024-10-10
[ENT-9488] - Exclude archived courses

hamzawaleed01:
## Post-review


enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
```diff
@@ -95,6 +97,24 @@ def get(self, request, **kwargs):
...
+        """
+        Determines if a content item is active.
+        Args:
+            item (ContentMetadata): The content metadata item to check.
+        Returns:
+            bool: True if the item is active, False otherwise.
+                For courses, checks if any course run is active.
+                For other content types, always returns True.
+        """
+        if item.content_type == 'course':
```
sameenfatima78, line 110: have we checked if we return any other content types? course_run, program, pathway?

hamzawaleed01: Yes we return other types too. Can be seen in tests too. Here we're checking expired courses only, for other content types, it will always return `True`.

sameenfatima78 (approved):

PR #973, opened 2024-10-11
feat: synchronize restricted courses from discovery

iloveagent57:
https://2u-internal.atlassian.net/browse/ENT-9570
This is one of two PRs required for the ticket. In this one, we'll deal only with synchronizing top-level restricted courses.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -1,4 +1,5 @@
 import collections
+import copy
 import json
 from logging import getLogger
 from uuid import uuid4
@@ -20,7 +21,10 @@ from enterprise_catalog.apps.api.v1.utils import (
     get_most_recent_modified_time,
     update_query_parameters,
 )
-from enterprise_catalog.apps.api_client.discovery import CatalogQueryMetadata
+from enterprise_catalog.apps.api_client.discovery import (
+    CatalogQueryMetadata,
+    DiscoveryApiClient,
+)
 from enterprise_catalog.apps.api_client.enterprise_cache import (
     EnterpriseCustomerDetails,
 )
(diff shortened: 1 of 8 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -1307,30 +1361,65 @@ def update_contentmetadata_from_discovery(catalog_query, dry_run=False):
...
+    """
+    Fetch and assoicate any permitted restricted couress for the given catalog_query.
+    """
+    if not getattr(settings, 'SHOULD_FETCH_RESTRICTED_COURSE_RUNS', False):
+        return []
+
+    if not catalog_query.restricted_runs_allowed:
+        return []
+
+    restricted_course_keys = list(catalog_query.restricted_runs_allowed)
```
pwnage101: Nit: if the goal is to get the keys of a dict, I think dict.keys() is more explicit than list(dict).

enterprise_catalog/apps/catalog/models.py
```diff
@@ -818,7 +821,58 @@ def __str__(self):
...
+        parent_record = ContentMetadata.objects.get(content_key=course_key)
+        record, _ = cls.objects.update_or_create(
+            content_key=course_key,
+            content_uuid=course_metadata_dict['uuid'],
+            content_type=COURSE,
+            unrestricted_parent=parent_record,
+            catalog_query=catalog_query,
+            defaults={
+                '_json_metadata': course_metadata_dict,
+            },
```
pwnage101, line 853: shouldn't we just overwrite _json_metadata instead of deferring to the existing one?

pwnage101 (commented):
Looks good so far!

enterprise_catalog/apps/catalog/models.py
iloveagent57: This is what `defaults` does for `update_or_create` (and why"defaults" is a poorly-named keyward arg):
> The update_or_create method tries to fetch an object from database based on the given kwargs. If a match is found, it updates the fields passed in the defaults dictionary.
https://docs.djangoproject.com/en/4.2/ref/models/querysets/#update-or-create

pwnage101: :shriek: I fear I've been misusing update_or_create for years now...

pwnage101 (approved):
LGTM!

PR #974, opened 2024-10-11
chore: iterate on django admin fields for restricted runs

pwnage101:
ENT-9569

Screenshots
===

What a ContentMetadata COURSE looks like if it has restricted versions:

Screenshot 2024-10-11 at 12 10 49 PM

What a ContentMetadata RUN looks like if it is restricted and allowed by one or more catalog query:

Screenshot 2024-10-11 at 12 11 38 PM

### DIFF
--- a/enterprise_catalog/apps/catalog/admin.py
+++ b/enterprise_catalog/apps/catalog/admin.py
@@ -21,9 +21,17 @@ from enterprise_catalog.apps.catalog.models import (
 )
 
 
-def _html_list_from_objects(objs, viewname, str_callback=None):
-    str_callback = str_callback or str
+def _html_list_from_objects(objs, viewname, str_callback=str):
+    """
+    Get a pretty, clickable list of objects.
+
+    Args:
+      objs (iterable of Django ORM objects): List/queryset of objects to display.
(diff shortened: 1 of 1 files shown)


enterprise_catalog/apps/catalog/admin.py
```diff
@@ -96,15 +105,31 @@ def get_catalogs(self, obj):
...
+    @admin.display(description='Restricted Courses For This Restricted Run')
+    def get_restricted_courses_for_this_restricted_run(self, obj):
         restricted_runs_allowed_for_restricted_course = RestrictedRunAllowedForRestrictedCourse.objects.select_related(
             'course',
         ).filter(
             run=obj,
         )
         restricted_courses = (relationship.course for relationship in restricted_runs_allowed_for_restricted_course)
-        return _html_list_from_objects(restricted_courses, "admin:catalog_contentmetadata_change")
+        return _html_list_from_objects(restricted_courses, "admin:catalog_restrictedcoursemetadata_change")
```
pwnage101, line 121: FYI, this PR also fixes a bug from the previous PR where we display an object, but link to the wrong model.

iloveagent57 (approved):

PR #976, opened 2024-10-16
feat: include restricted runs in algolia objects

pwnage101:
[ENT-9505](https://2u-internal.atlassian.net/browse/ENT-9505)

Important features:
* **Restricted Runs Visible:** The metadata actually used for indexing courses always comes from the "canonical restricted course" object if one exists (i.e. the related RestrictedCourseMetadata that has catalog_query=null).
* **Unicorn Course Support:** A course normally is not indexed at all if it has no advertised course run.  Now we index "empty" courses as long as they contain at least one restricted run (i.e. the canonical restricted variant does have an advertised run).
  * Furthermore, we do NOT associate the course with the standard set of related catalog queries.  Instead we use a subset that represents only the catalog queries that explicitly allow runs in that course.

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -11,6 +11,7 @@ from operator import itemgetter
 from celery import shared_task, states
 from celery.exceptions import Ignore
 from celery_utils.logged_task import LoggedTask
+from django.conf import settings
 from django.db import IntegrityError
 from django.db.models import Prefetch, Q
 from django.db.utils import OperationalError
@@ -518,6 +519,11 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False): 
             f'{_reindex_algolia_prefix(dry_run)} invoking task with arguments force={force}, dry_run={dry_run}.'
         )
         courses_content_metadata = ContentMetadata.objects.filter(content_type=COURSE)
+        # Make sure the courses we consider for indexing actually contain restricted runs so that
+        # "unicorn" courses (i.e. courses that contain only restricted runs) do not get discarded by
+        # partition_course_keys_for_indexing() for not having an advertised run.
+        if getattr(settings, 'SHOULD_INDEX_COURSES_WITH_RESTRICTED_RUNS', False):
+            courses_content_metadata = courses_content_metadata.prefetch_restricted_overrides()
         indexable_course_keys, nonindexable_course_keys = partition_course_keys_for_indexing(
             courses_content_metadata,
         )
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -1746,74 +1755,136 @@ def mock_replace_all_objects(products_iterable):
         assert expected_call_args == self._sort_tags_in_algolia_object_list(actual_call_args)
 
     @mock.patch('enterprise_catalog.apps.api.tasks.get_initialized_algolia_client', return_value=mock.MagicMock())
-    # pylint: disable=too-many-statements
```
pwnage101, line 1749: github really fudged this diff view.  If you look at the file directly I only added two tests:
* test_index_algolia_restricted_runs_mixed_course
* test_index_algolia_restricted_runs_unicorn_course

enterprise_catalog/apps/api/tasks.py
```diff
@@ -888,6 +905,18 @@ def _get_algolia_products_for_batch(
...
+            # catalog queries that explicitly allow runs in those courses. We can tell that a course
+            # has only restricted runs simply by checking that it normally doesn't have an
+            # advertised run.
+            # pylint: disable=protected-access
+            is_unrestricted_course_advertised = bool(metadata._json_metadata.get('advertised_course_run_uuid'))
+            if not is_unrestricted_course_advertised:
+                associated_catalog_queries = filter(
+                    None,
+                    (rc.catalog_query for rc in metadata.restricted_courses.all())
+                )
```
iloveagent57: nit: why not filter in the queryset? i.e. 
```python
associated_catalog_queries = [rc.catalog_query for rc in metadata.restricted_courses.exclude(catalog_query=None)]
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -888,6 +905,18 @@ def _get_algolia_products_for_batch(
             # Course runs should contribute their UUIDs to the parent course.
             content_key = metadata.parent_content_key
         associated_catalog_queries = metadata.catalog_queries.all()
+        if metadata.content_type == COURSE and getattr(settings, 'SHOULD_INDEX_COURSES_WITH_RESTRICTED_RUNS', False):
+            # "unicorn" courses (i.e. courses with only restricted runs) should only be indexed for
+            # catalog queries that explicitly allow runs in those courses. We can tell that a course
+            # has only restricted runs simply by checking that it normally doesn't have an
+            # advertised run.
+            # pylint: disable=protected-access
```
iloveagent57, line 912: nit: you could add an `advertised_course_run_uuid()` property to `ContentMetadata` to avoid this protected-access warning.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -615,6 +613,29 @@ def get_xapi_activity_id(self, content_resource, content_key):
         return xapi_activity_id
 
 
+class ContentMetadataQuerySet(models.QuerySet):
```
iloveagent57, line 616: Nice!

iloveagent57 (approved):

enterprise_catalog/apps/api/tasks.py
pwnage101: Actually, I'm not sure that kind of getter design is sufficiently explicit when there's potentially two options: the values from _json_metadata vs. json_metadata, especially when I expect the value to differ depending on which is queried. My preference is to keep this as written and elaborate on the code comment.

PR #977, opened 2024-10-17
feat: syncrhonize restricted course runs

iloveagent57:
ENT-9570
Updates or creates restricted course _run_ metadata records when restricted courses are updated, and links the restricted run content to the restricted course. Note this makes additional call to discovery's `search/all/` endpoint for restricted run metadata, one per query that allows restricted runs.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -842,15 +842,13 @@ class RestrictedCourseMetadata(BaseContentMetadata):
 
         course_key = course_metadata_dict['key']
         parent_record = ContentMetadata.objects.get(content_key=course_key, content_type=COURSE)
+
         record, _ = cls.objects.update_or_create(
             content_key=course_key,
-            content_uuid=course_metadata_dict['uuid'],
             content_type=COURSE,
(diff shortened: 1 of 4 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -892,6 +901,32 @@ def filter_restricted_runs(cls, course_metadata_dict, catalog_query):
...
+        """
+        Given a ``course_metadata_dict``, returns a filtered list of ``course_runs``
+        containing only unrestricted runs and restricted runs that are allowed by
+        the provided ``catalog_query``.
+        """
+        allowed_restricted_runs = catalog_query.restricted_runs_allowed.get(course_metadata_dict['key'], [])
+        return [
+            run for run in course_metadata_dict['course_runs']
+            if run.get(COURSE_RUN_RESTRICTION_TYPE_KEY) is None or run['key'] in allowed_restricted_runs
+        ]
```
pwnage101: Nit: I think this would be more self-documenting if it composed restricted_runs_for_course(), then simply appended all unrestricted runs.

pwnage101 (approved):

PR #980, opened 2024-10-18
fix: clear and reset restricted run relationships

iloveagent57:
Small change so that we fully reset the restricted course <-> run relationships when we sync restricted courses.
## Post-review


brobro10000 (approved):

pwnage101 (approved):
good catch!

PR #983, opened 2024-10-18
Adding v2 of catalog-based get_content_metadata endpoint

pmakwana93:
Adding v2 of catalog-based get_content_metadata endpoint

JIRA: https://2u-internal.atlassian.net/browse/ENT-9406

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v2/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v2/tests/test_views.py
@@ -1,0 +1,485 @@
+from datetime import datetime
+from unittest import mock
+
+import ddt
+import pytz
+from rest_framework import status
+from rest_framework.reverse import reverse
+
+from enterprise_catalog.apps.api.v1.tests.mixins import APITestMixin
+from enterprise_catalog.apps.catalog.constants import (
+    COURSE_RUN_RESTRICTION_TYPE_KEY,
+    RESTRICTION_FOR_B2B,
+)
+from enterprise_catalog.apps.catalog.models import ContentMetadata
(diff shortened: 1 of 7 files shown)


enterprise_catalog/apps/api/v2/tests/test_views.py
```diff
@@ -0,0 +1,266 @@
...
+                                'key': 'course-v1:edX+course+run1',
+                                'status': 'published',
+                                'is_enrollable': True,
+                                'is_marketable': True,
+                            },
+                            {
+                                'key': 'course-v1:edX+course+run2',
+                                'status': 'published',
+                                'is_enrollable': True,
+                                'is_marketable': True,
```
iloveagent57: nit: just to make the data more prod-like, could we try setting `is_marketable: False`?

enterprise_catalog/apps/api/v2/tests/test_views.py
```diff
@@ -0,0 +1,266 @@
...
+                    },
+                },
+            },
+            'create_restricted_run_allowed_for_restricted_course': [
+                {'course': 1, 'run': 'course-v1:edX+course+run2'},
+            ],
+        },
+    )
+    @ddt.unpack
+    def test_get_content_metadata(
```
iloveagent57, line 230: Consider adding one or maybe two more test cases where no content keys filter is passed in the GET request, and the catalog query either allows or does not allow a restricted run.

iloveagent57 (commented):
This is looking good. One suggestion to add a test case or two, after that I think it'll be ready to go.

iloveagent57 (approved):

PR #985, opened 2024-10-21
feat: sync full metadata for restricted courses

iloveagent57:
ENT-9597

### Set up a restricted course run locally
It'll be associated with DemoX: https://2u-internal.atlassian.net/wiki/spaces/SOL/pages/1317306375/Setting+up+restricted+runs+locally
### Enable the setting
In your `private.py` settings file, set `SHOULD_FETCH_RESTRICTED_COURSE_RUNS = True`. Optionally set `SHOULD_INDEX_COURSES_WITH_RESTRICTED_RUNS = True` if you'd like to test against a dev algolia index, too.
### Sync full course metadata
1. `make worker-restart worker-logs`
2. In a different shell, do `make app-shell` and then `./manage.py update_content_metadata --force` (note that this mgmt command runs both the `update_catalog_metadata` task _and_ the `update_full_content_metadata_task`.)
3. Observe that a restricted course exists here: http://localhost:18160/admin/catalog/restrictedcoursemetadata/ and that it has a relationship to a restricted run.
4. Optionally update a test algolia index with `./manage.py reindex_algolia --force --no-async`
## Post-review


enterprise_catalog/apps/api/tasks.py
```diff
@@ -344,11 +394,6 @@ def _update_single_full_course_record(course_metadata_dict, metadata_record, cou
             metadata_record.content_key, json.dumps(metadata_record.json_metadata)
         ))
 
-    program_content_keys = create_course_associated_programs(
-        course_metadata_dict.get('programs', []),
-        metadata_record,
-    )
-    _update_full_content_metadata_program(program_content_keys, dry_run)
```
iloveagent57, line 351: Only change here was yanking this out of `_update_single_full_course_record()` and into the calling function, to better isolate responsibilities and avoid duplicate work.

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -768,6 +776,7 @@ def test_get_upcoming_course_runs(self, searchable_course, expected_course_runs)
                     'content_price': 0.0,
                     'is_active': False,
                     'is_late_enrollment_eligible': False,
+                    'restriction_type': None,
```
adamstankiewicz, line 796: [nit/curious]: are there any test cases wrt `test_get_course_runs` that should result in a non-null `restriction_type` in the returned runs?

enterprise_catalog/apps/catalog/migrations/0043_restricted_course_m2m_field.py
```diff
@@ -0,0 +1,24 @@
...
+            model_name='restrictedcoursemetadata',
+            name='restricted_run_allowed_for_restricted_course',
+            field=models.ManyToManyField(through='catalog.RestrictedRunAllowedForRestrictedCourse', to='catalog.contentmetadata'),
+        ),
+        migrations.AlterField(
+            model_name='restrictedrunallowedforrestrictedcourse',
+            name='course',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.SET_NULL, to='catalog.restrictedcoursemetadata'),
+        ),
+    ]
```
iloveagent57, line 24: This is a SQL no-op.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -851,55 +856,122 @@ def __str__(self):
...
+    def _store_record(cls, course_metadata_dict, catalog_query=None, is_full_update=False):
         """
         Given a course metadata dictionary, stores a corresponding
         ``RestrictedContentMetadata`` record. Raises if the content key
         is not of type 'course', or if a corresponding unrestricted parent
         record cannot be found.
         """
         content_type = course_metadata_dict.get('content_type')
-        if content_type != COURSE:
+        if content_type and (content_type != COURSE):
```
adamstankiewicz: [nit/clarification]

1. When will `content_type` be falsey where the added check on its truthiness is needed? For example, if `content_type` is not truthy (e.g., `None`), I believe this would still proceed with the code path below the raised `Exception` within this conditional block with an invalid `content_type`. Is this expected?
2. Are the parens needed around `(content_type != COURSE)`?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -851,55 +856,122 @@ def __str__(self):
...
+        # and restricted runs for this course before rebuilding them.
+        self.catalog_query.contentmetadata_set.filter(
+            parent_content_key=self.content_key,
+        ).delete()
+
+        for course_run_dict in self.restricted_run_dicts:
+            # Explicitly don't overwrite the content key, type, or parent.
+            defaults = _get_defaults_from_metadata(course_run_dict)
+            for key in ['content_key', 'content_type', 'parent_content_key']:
+                defaults.pop(key, None)
```
adamstankiewicz: [nit/sanity check] Is there any risk of addtl fields returned by `_get_defaults_from_metadata` in the future that also shouldn't be overwritten? For example, if/when that occurs, this list of fields to pop would need to be modified. I wonder if it's worth defining the explicit fields to overwrite vs. ignoring fields that shouldn't overwrite.

enterprise_catalog/apps/catalog/models.py
```diff
@@ -851,55 +856,122 @@ def __str__(self):
...
+        """
+        if is_full_update:
+            self._json_metadata.update(course_metadata_dict)
+        else:
+            self._json_metadata.update(
+                _get_defaults_from_metadata(course_metadata_dict)['_json_metadata'],
+            )
+        self.save()
+
+    def update_course_run_relationships(self):
```
adamstankiewicz, line 917: [nit/curious] I wonder if it might be helpful to log anything (i.e., via `logger.info`) within this method. For example, would it be helpful to log the existing M2M relationships vs. the new ones for observability/debugging?

enterprise_catalog/apps/catalog/models.py
```diff
@@ -851,55 +856,122 @@ def __str__(self):
...
+            if run['key'] in allowed_restricted_runs
+        ]
+
+    @property
+    def restricted_run_dicts(self):
+        return self.restricted_runs_for_course(self.json_metadata, self.catalog_query)
+
+    def update_metadata(self, course_metadata_dict, is_full_update=False):
+        """
+        Updates and saves the json_metadata for this restricted course.
```
adamstankiewicz, line 892: [nit] Might be helpful to describe the purpose of `is_full_update` in the docstring to remove any ambiguity over its intent.

adamstankiewicz (commented):
Looking good to me, left some nit-level comments to consider/clarify (have not tested locally).

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
iloveagent57: yup, added!

enterprise_catalog/apps/catalog/models.py
iloveagent57: hah, yeah, this seems like a vestige of test-debugging.  I'll switch it back and fix the test setup (if needed).

enterprise_catalog/apps/catalog/models.py
iloveagent57: This was more testing cruft, I've fixed up the test to reflect reality.

iloveagent57:
Made some recent updates to this branch:
* the `update_course_run_relationships()` function was previously _deleting_ and re-creating course run `ContentMetadata` records, instead of just resetting M2M relationships. This has been fixed to now use a M2M `.set()` call so only the relationships are re-created, and not the CM records.
* added a `update_or_create_run()` helper, specifically to ensure that we pop defaults prior to `update_or_create()` calls, which really is necessary, as described in the related code comment.
* adds more logging.

enterprise_catalog/apps/api/tasks.py
```diff
@@ -320,14 +367,17 @@ def _get_course_records_by_key(fetched_course_keys):
...
+    course_metadata_dict, metadata_record, course_review, dry_run, skip_json_metadata_update=False,
+):
     """
     Given a fetched dictionary of course content metadata and an option `course_review` record,
     updates an existing course `ContentMetadata` instance with the "full" dictionary of `json_metadata`
     for that course.
     """
-    # Merge the full metadata from discovery's /api/v1/courses into the local metadata object.
-    metadata_record._json_metadata.update(course_metadata_dict)  # pylint: disable=protected-access
+    if not skip_json_metadata_update:
```
adamstankiewicz, line 378: Nice

adamstankiewicz (approved):
LGTM!

PR #966, opened 2024-10-09
feat: take exec ed course data from course run instead of additional_metadata attempt 3

marlonkeating:
## Description
Updates references for Exec Ed courses originally contained in the `additional_metadata` field, to now be taken from course runs.  Fixing datetime parsing issue that was found in earlier attempt https://github.com/openedx/enterprise-catalog/pull/912, by incorporating fixes made in https://github.com/openedx/enterprise-catalog/pull/958

## Ticket Link
[ENT-8248: update custom logic for non-ocm courses in enterprise-catalog](https://2u-internal.atlassian.net/browse/ENT-8248)

## Outstanding Questions
- [ ] Do we still need to process Exec Ed courses differently from OCM courses?

## Post-review
* [ ] Squash commits into discrete sets of changes
* [ ] Ensure that once the changes have been [deployed to stage](https://gocd.tools.edx.org/go/pipeline/activity/stage-enterprise_catalog), prod is manually deployed

PR #988, opened 2024-10-24
feat: take exec ed course data from course run instead of additional_metadata attempt 4

marlonkeating:
## Description

Updates references for Exec Ed courses originally contained in the `additional_metadata` field, to now be taken from course runs.  This new version fixes datetime parsing issue that was found in earlier attempt https://github.com/openedx/enterprise-catalog/pull/966, and adds additional guardrails and logging around future unexpected parsing errors.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -229,7 +229,7 @@ def _should_index_course(course_metadata):
         return not is_course_run_active(advertised_course_run)
 
     def deadline_passed_checker():
-        return _has_enroll_by_deadline_passed(course_json_metadata, advertised_course_run)
+        return _has_enroll_by_deadline_passed(course_json_metadata)
 
     for should_not_index_function, log_message in (
         (no_advertised_course_run_checker, 'no advertised course run'),
@@ -246,23 +246,18 @@ def _should_index_course(course_metadata):
     return True
 
 
-def _has_enroll_by_deadline_passed(course_json_metadata, advertised_course_run):
+def _has_enroll_by_deadline_passed(course_json_metadata):
     """
     Helper to determine if the enrollment deadline has passed for the given course
-    and advertised course run.  For course metadata records with a `course_type` of "course" (e.g. OCM courses),
-    this is based on the verified upgrade deadline.
-    For 2u exec ed courses, this is based on the registration deadline.
+    based on normalized_metadata's enroll_by_date
     """
-    enroll_by_deadline_timestamp = 0
-    if course_json_metadata.get('course_type') == EXEC_ED_2U_COURSE_TYPE:
-        additional_metadata = course_json_metadata.get('additional_metadata') or {}
-        registration_deadline = additional_metadata.get('registration_deadline')
-        if registration_deadline:
-            enroll_by_deadline_timestamp = parse_datetime(registration_deadline).timestamp()
+    enroll_by_deadline = course_json_metadata.get('normalized_metadata')['enroll_by_date']
+    if isinstance(enroll_by_deadline, str):
+        enroll_by_deadline_timestamp = parse_datetime(enroll_by_deadline).timestamp()
+        return enroll_by_deadline_timestamp < localized_utcnow().timestamp()
     else:
-        enroll_by_deadline_timestamp = _get_verified_upgrade_deadline(advertised_course_run)
-
-    return enroll_by_deadline_timestamp < localized_utcnow().timestamp()
+        # Courses without enrollment deadline shouldn't be disqualified
+        return False
 
 
 def partition_course_keys_for_indexing(courses_content_metadata):
@@ -281,11 +276,16 @@ def partition_course_keys_for_indexing(courses_content_metadata):
     nonindexable_course_keys = set()
 
     for course_metadata in courses_content_metadata:
-        if _should_index_course(course_metadata):
-            indexable_course_keys.add(course_metadata.content_key)
-        else:
-            nonindexable_course_keys.add(course_metadata.content_key)
-
+        try:
+            if _should_index_course(course_metadata):
+                indexable_course_keys.add(course_metadata.content_key)
+            else:
+                nonindexable_course_keys.add(course_metadata.content_key)
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            logger.warning(
+                f"Failed determining indexable status for course_metadata "
+                f"'{course_metadata.content_key}' due to: {e}"
+            )
     return list(indexable_course_keys), list(nonindexable_course_keys)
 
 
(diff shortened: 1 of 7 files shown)


enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -246,23 +246,18 @@ def deadline_passed_checker():
...
-    enroll_by_deadline_timestamp = 0
-    if course_json_metadata.get('course_type') == EXEC_ED_2U_COURSE_TYPE:
-        additional_metadata = course_json_metadata.get('additional_metadata') or {}
-        registration_deadline = additional_metadata.get('registration_deadline')
-        if registration_deadline:
-            enroll_by_deadline_timestamp = parse_datetime(registration_deadline).timestamp()
+    based on normalized_metadata's enroll_by_date
+    """
+    enroll_by_deadline = course_json_metadata.get('normalized_metadata')['enroll_by_date']
+    if isinstance(enroll_by_deadline, str):
```
marlonkeating, line 255: [inform] Change from https://github.com/openedx/enterprise-catalog/pull/966 to fix the parsing issue we ran into during prod testing: `TypeError: fromisoformat: argument must be str`

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -281,11 +276,16 @@ def partition_course_keys_for_indexing(courses_content_metadata):
...
-            indexable_course_keys.add(course_metadata.content_key)
-        else:
-            nonindexable_course_keys.add(course_metadata.content_key)
-
+        try:
+            if _should_index_course(course_metadata):
+                indexable_course_keys.add(course_metadata.content_key)
+            else:
+                nonindexable_course_keys.add(course_metadata.content_key)
+        except Exception as e:  # pylint: disable=broad-exception-caught
```
marlonkeating, line 284: [inform] Change from https://github.com/openedx/enterprise-catalog/pull/966 to catch issue and prevent it from stopping the whole indexing process, while logging any future parsing issues (along with the culprit's content_key).

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -49,40 +62,56 @@ class AlgoliaUtilsTests(TestCase):
...
         },
         {
             'course_type': EXEC_ED_2U_COURSE_TYPE,
             'expected_result': False,
-            'additional_metadata': {'registration_deadline': '2021-03-21T23:59:59Z'},
+            'start': days_from_now(-30),
+            'end': days_from_now(-1),
+            'enrollment_end': days_from_now(-30)
         },
+        # Error cases for invalid enrollment_end values
```
marlonkeating: [inform] Change from https://github.com/openedx/enterprise-catalog/pull/966 to test two data cases that resulted in the error we saw in Prod: `TypeError: fromisoformat: argument must be str`

enterprise_catalog/apps/catalog/algolia_utils.py
adamstankiewicz: [nit/clarification/suggestion] What was the value of `enroll_by_deadline` when not a string? If it could be something other than `None` or a string, I think this type check to explicitly be a string makes sense. Though, if it's only ever a string vs. `None`, I might consider whether a check for truthiness might be preferable, e.g.:

```python
# Always return false if there is no enroll-by deadline present.
if not enroll_by_deadline:
    return False

# Otherwise, determine if the enroll-by deadline is before now.
enroll_by_deadline_timestamp = parse_datetime(enroll_by_deadline).timestamp()
return enroll_by_deadline_timestamp < localized_utcnow().timestamp()
```

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -246,23 +246,18 @@ def deadline_passed_checker():
...
+    enroll_by_deadline = course_json_metadata.get('normalized_metadata')['enroll_by_date']
+    if isinstance(enroll_by_deadline, str):
+        enroll_by_deadline_timestamp = parse_datetime(enroll_by_deadline).timestamp()
+        return enroll_by_deadline_timestamp < localized_utcnow().timestamp()
     else:
-        enroll_by_deadline_timestamp = _get_verified_upgrade_deadline(advertised_course_run)
-
-    return enroll_by_deadline_timestamp < localized_utcnow().timestamp()
+        # The check should fail if the enroll by date isn't present or can't be parsed
+        return True
```
adamstankiewicz: [sanity check] If a course doesn't have an enroll-by date, is it semantically correct to return `True` for `_has_enroll_by_deadline_passed`? That is, how could the enroll by deadline be passed when there is no deadline present?

I believe this might prevent courses without an enroll-by date from being indexed into Algolia due to the `deadline_passed_checker` method in `_should_index_course`; not sure if that's intended.

enterprise_catalog/apps/catalog/algolia_utils.py
marlonkeating: Unfortunately I don't know what the exact value was that caused the error, due to insufficient logging (which I've tried to rectify here).  In the unit tests I tried `int` or `None` values and got the same error message.

enterprise_catalog/apps/catalog/algolia_utils.py
marlonkeating: Do you know if courses lacking an enroll-by date is a valid scenario, and shouldn't be disqualified from indexing?

enterprise_catalog/apps/catalog/algolia_utils.py
adamstankiewicz: Based on the code, it seems it should only be `None` or string, but probably better to air on the side of caution given the uncertainty (i.e., keeping the explicit string check for now).

enterprise_catalog/apps/catalog/algolia_utils.py
adamstankiewicz: Conceptually, a course without an enroll-by date means there is no deadline to enroll, so as long as the other properties checks within `is_course_run_active` are fulfilled (e.g., `status="published"`, `is_enrollment=True`, `is_marketable=True`) the course should still be discoverable despite null enroll-by date since it's otherwise still enrollable.

We have handled null enroll-by date in at least a couple places in the past, at least within a couple frontend spots, e.g. 
* Related to which course runs appear during the assignment allocation flow in Admin Portal ([source](https://github.com/openedx/frontend-app-admin-portal/blob/4f9d197a9eb3ef2c7c9e4d36059e81da552fca27/src/components/learner-credit-management/data/utils.js#L642-L649)). That is, `isEnrollByDateWithinThreshold` always returns true if there is no enroll by date.
* Related to determining whether an audit enrollment is potentially upgradable in the Learner Portal ([source](https://github.com/openedx/frontend-app-learner-portal-enterprise/blob/01091da53297886d8b3ba1d78cc7b1edd558b21b/src/components/app/data/utils.js#L785-L787)). That is, within `isEnrollmentUpgradeable`, `isEnrollByLapsed` always returns false when no enroll-by date is present. `isEnrollByLapsed` is largely synonymous with `_has_enroll_by_deadline_passed`.

adamstankiewicz (approved):

brobro10000 (approved):

PR #994, opened 2024-11-06
feat: v2 catalog contains_content_items view

iloveagent57:
1. Set up restricted content metadata in a catalog as described here: https://github.com/openedx/enterprise-catalog/pull/985
2. Observe that `GET http://localhost:18160/api/v1/enterprise-catalogs/7467c9d2-433c-4f7e-ba2e-c5c7798527b2/contains_content_items/?course_run_ids=course-v1%3AedX%2BDemoX%2Brestricted_run` does *not* contain your restricted run.
3. Observe that the v2 endpoint `http://localhost:18160/api/v2/enterprise-catalogs/7467c9d2-433c-4f7e-ba2e-c5c7798527b2/contains_content_items/?course_run_ids=course-v1%3AedX%2BDemoX%2Brestricted_run` *does* contain your restricted content.

ENT-9408 

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views/enterprise_catalog_contains_content_items.py
+++ b/enterprise_catalog/apps/api/v1/views/enterprise_catalog_contains_content_items.py
@@ -41,6 +41,14 @@ class EnterpriseCatalogContainsContentItems(BaseViewSet, viewsets.ReadOnlyModelV
             return str(enterprise_catalog.enterprise_uuid)
         return None
 
+    def catalog_contains_content_items(self, content_keys):
+        """
+        Returns a boolean indicating whether all of the provided content_keys
+        are contained by the catalog record associated with the current request.
+        """
+        enterprise_catalog = self.get_object()
+        return enterprise_catalog.contains_content_keys(content_keys)
+
     # Becuase the edx-rbac perms are built around a part of the URL
     # path, here (the uuid of the catalog), we can utilize per-view caching,
     # rather than per-user caching.
@@ -56,6 +64,6 @@ class EnterpriseCatalogContainsContentItems(BaseViewSet, viewsets.ReadOnlyModelV
         """
         course_run_ids = unquote_course_keys(course_run_ids)
 
-        enterprise_catalog = self.get_object()
-        contains_content_items = enterprise_catalog.contains_content_keys(course_run_ids + program_uuids)
-        return Response({'contains_content_items': contains_content_items})
+        return Response({
+            'contains_content_items': self.catalog_contains_content_items(course_run_ids + program_uuids),
+        })
--- a/enterprise_catalog/apps/api/v2/tests/test_enterprise_catalog_views.py
+++ b/enterprise_catalog/apps/api/v2/tests/test_enterprise_catalog_views.py
@@ -1,0 +1,157 @@
+import uuid
+from datetime import datetime, timedelta
+from unittest import mock
+
+import ddt
+import pytest
+import pytz
+from rest_framework import status
+
+from enterprise_catalog.apps.api.base.tests.enterprise_catalog_views import (
+    BaseEnterpriseCatalogViewSetTests,
+)
+from enterprise_catalog.apps.catalog.constants import (
+    COURSE,
+    COURSE_RUN,
+    RESTRICTED_RUNS_ALLOWED_KEY,
+)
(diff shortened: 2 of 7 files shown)


enterprise_catalog/apps/api/v1/views/enterprise_catalog_contains_content_items.py
```diff
@@ -41,6 +41,14 @@ def get_permission_object(self):
             return str(enterprise_catalog.enterprise_uuid)
         return None
 
+    def catalog_contains_content_items(self, content_keys):
```
pwnage101, line 44: Nice refactor 👍

enterprise_catalog/apps/api/v2/tests/test_enterprise_catalog_views.py
```diff
@@ -0,0 +1,159 @@
...
+        )
+        return content_one, restricted_course, restricted_run
+
+    def test_contains_catalog_key_restricted_runs_allowed(self):
+        """
+        Tests that a catalog is considered to contain a restricted run.
+        """
+        content_one, _, restricted_run = self._create_restricted_course_and_run(self.enterprise_catalog)
+
+        self.add_metadata_to_catalog(self.enterprise_catalog, [content_one, restricted_run])
```
pwnage101: For this test to be fully rigorous (and accurate to what prod data looks like), I think we do not want to _directly_ add `restricted_runs` to the catalog here.
```suggestion
        self.add_metadata_to_catalog(self.enterprise_catalog, [content_one])
```

It should theoretically be sufficient (and what we really want to test) that restricted run is related via `RestrictedCourseMetadata` and `RestrictedRunAllowedForRestrictedCourse` which were set up above.

pwnage101: I just tested this change and it causes the result to flip to False. 🤔

iloveagent57: Same, but I think I'm missing a test setup step of defining the `restricted_runs_allowed` for the test catalog's query.

pwnage101: I think for this entire v2 test class, we're probably missing a step to modify the catalog query's content filter to include a top-level "restricted_runs_allowed" key.  Without that json key, some of the restricted runs logic gets short-circuited in the name of performance.

pwnage101: didn't see your comment before posting mine, agree!

iloveagent57: should be all fixed up now, along with the enterprise-customer restricted view and related tests

enterprise_catalog/apps/api/v2/tests/test_enterprise_catalog_views.py
```diff
@@ -0,0 +1,157 @@
...
+        restricted_run = ContentMetadataFactory.create(
+            content_key='course-v1:org+key1+restrictedrun',
+            parent_content_key=restricted_course.content_key,
+            content_type=COURSE_RUN,
+        )
+        restricted_course.restricted_run_allowed_for_restricted_course.set(
+            [restricted_run], clear=True,
+        )
+        catalog.catalog_query.content_filter[RESTRICTED_RUNS_ALLOWED_KEY] = {
+            content_one.content_key: [restricted_run.content_key],
```
pwnage101, line 125: Nit: the top-level key under the restricted_runs_allowed key should actually be an aggregation key, which means you should prefix this with "course:".  It's merely a formality for this test suite, but I think important for people who might ever need to touch this test or refer to it as a code reference.

pwnage101 (approved):

enterprise_catalog/apps/api/v2/tests/test_enterprise_catalog_views.py
iloveagent57: It can actually be a regular course key or an aggregation key, the code is flexible about it:
https://github.com/openedx/enterprise-catalog/blob/54255107f82fd3b067588f2215ea609a7b9d7c15/enterprise_catalog/apps/catalog/models.py#L117-L137

PR #1003, opened 2024-11-21
feat: added relevant sort on videos

irfanuddinahmad:
**Description**: This PR adds a replica index for sorting the search results by video duration. This has the effect of distributing the video hits randomly so that similar videos from the same source are not clustered together in the search results.

**JIRA**: [ENT-9780](https://2u-internal.atlassian.net/browse/ENT-9780)

## Post-review


jajjibhai008 (approved):
LGTM

PR #1009, opened 2024-12-07
fix: restricted run relationships should not be updated during update_full_content_metadata

pwnage101:


enterprise_catalog/apps/api/tasks.py
```diff
@@ -341,10 +341,6 @@ def _update_full_restricted_course_metadata(modified_metadata_record, course_rev
         # First, update the restricted course record's json metadata to use the full metadata
         # from above.
         restricted_course.update_metadata(full_restricted_metadata, is_full_update=True)
-        # Second, if it's not the canonical copy of the restricted course,
-        # we have to reset the restricted course *run* relationships.
-        if restricted_course.catalog_query:
-            restricted_course.update_course_run_relationships()
```
pwnage101, line 347: This is the only change to business logic.

iloveagent57 (approved):

PR #1017, opened 2024-12-16
[REFERENCE] fix: ensure course run key is passed in legacy enrollment urls

adamstankiewicz:
While further triaging https://2u-internal.atlassian.net/browse/ENT-9770, we raised the question of **_WHY_** learners are seemingly visiting the legacy `enrollment_url` with a top-level _course_ key vs. an explicit _course run_ key.

When this occurs, it seems the `InvalidKeyError` is raised, regardless of the `/enroll` suffix being included in the course identifier passed to `CourseKey.from_string(...)`.

This PR demonstrates a possible fix for what seems like a bug where generating an `enrollment_url` for a `ContentMetadata` record with `content_type="course"` results in the top-level course key being included instead. The proposed fix here would be to resolve the advertised course run key to use in the `enrollment_url` instead of the top-level course key, which should also help mitigate the `InvalidKeyError` error.

In theory, no `InvalidKeyError` errors should be raised from this legacy enrollment view.

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/models.py
+++ b/enterprise_catalog/apps/catalog/models.py
@@ -57,6 +57,7 @@ from enterprise_catalog.apps.catalog.utils import (
     get_parent_content_key,
     localized_utcnow,
 )
+from enterprise_catalog.apps.catalog.algolia_utils import get_advertised_course_run
 
 
 LOGGER = getLogger(__name__)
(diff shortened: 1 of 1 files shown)


adamstankiewicz:
Closing in favor of https://github.com/openedx/enterprise-catalog/pull/1018

PR #1018, opened 2024-12-19
fix: set enrollment url course id to advertised course run key

brobro10000:
Updates how we set the enrollment url if the learner portal is not enabled. This is to fix an error that is manifesting in the LMS where the enrollment url is parsed for a course_id, but a top level course is retrieved. This breaks the expectation of what the OpaqueKeys `CourseKey` is expecting which includes a course run. 

The changes included ensures a course run is always passed within the enrollment url.

This is a root cause solution to this [PR](https://github.com/openedx/edx-platform/pull/36028)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/catalog/algolia_utils.py
+++ b/enterprise_catalog/apps/catalog/algolia_utils.py
@@ -31,8 +31,8 @@ from enterprise_catalog.apps.catalog.constants import (
     VIDEO,
 )
 from enterprise_catalog.apps.catalog.content_metadata_utils import (
+    get_advertised_course_run,
     get_course_first_paid_enrollable_seat_price,
-    get_course_run_by_uuid,
     is_course_run_active,
 )
 from enterprise_catalog.apps.catalog.models import ContentMetadata
@@ -229,11 +229,7 @@ def _should_index_course(course_metadata):
         bool: Whether or not the course should be indexed by algolia.
     """
     course_json_metadata = course_metadata.json_metadata
-    advertised_course_run_uuid = course_json_metadata.get('advertised_course_run_uuid')
-    advertised_course_run = get_course_run_by_uuid(
-        course_json_metadata,
-        advertised_course_run_uuid,
-    )
+    advertised_course_run = get_advertised_course_run(course_json_metadata)
 
     # We define a series of no-arg functions, each of which has the property that,
     # if it returns true, means we should *not* index this record.
@@ -477,8 +473,7 @@ def get_course_language(course):
     Returns:
         string: human-readable language name parsed from a language code, or None if language name is not present.
     """
-    advertised_course_run = get_course_run_by_uuid(course, course.get('advertised_course_run_uuid'))
-    if not advertised_course_run:
+    if not (advertised_course_run := get_advertised_course_run(course)):
         return None
 
     content_language_name = advertised_course_run.get('content_language_search_facet_name')
@@ -496,8 +491,7 @@ def get_course_transcript_languages(course):
     Returns:
         list: a list of available human-readable video transcript languages parsed from a language code.
     """
-    advertised_course_run = get_course_run_by_uuid(course, course.get('advertised_course_run_uuid'))
-    if not advertised_course_run:
+    if not (advertised_course_run := get_advertised_course_run(course)):
         return None
 
     transcript_languages = advertised_course_run.get('transcript_languages_search_facet_names')
@@ -1168,23 +1162,6 @@ def _get_course_run(course, course_run):
     return course_run
 
 
-def get_advertised_course_run(course):
-    """
-    Get part of the advertised course_run as per advertised_course_run_uuid
-
-    Argument:
-        course (dict)
-
-    Returns:
-        dict: containing key, pacing_type, start, end, and upgrade deadline
-        for the course_run, or None
-    """
-    full_course_run = get_course_run_by_uuid(course, course.get('advertised_course_run_uuid'))
-    if full_course_run is None:
-        return None
-    return _get_course_run(course, full_course_run)
-
-
 def get_course_runs(course):
     """
     Extract and transform a list of course runs into what we'll index.
@@ -1514,6 +1491,8 @@ def _algolia_object_from_product(product, algolia_fields):
     """
     searchable_product = copy.deepcopy(product)
     if searchable_product.get('content_type') == COURSE:
+        advertised_course_run = get_advertised_course_run(searchable_product)
+        transformed_advertised_course_run = _get_course_run(searchable_product, advertised_course_run)
         searchable_product.update({
             'language': get_course_language(searchable_product),
             'availability': get_course_availability(searchable_product),
@@ -1522,7 +1501,7 @@ def _algolia_object_from_product(product, algolia_fields):
             'program_titles': get_course_program_titles(searchable_product),
             'subjects': get_course_subjects(searchable_product),
             'card_image_url': get_course_card_image_url(searchable_product),
-            'advertised_course_run': get_advertised_course_run(searchable_product),
+            'advertised_course_run': transformed_advertised_course_run,
             'course_runs': get_course_runs(searchable_product),
             'upcoming_course_runs': get_upcoming_course_runs(searchable_product),
             'skill_names': get_course_skill_names(searchable_product),
--- a/enterprise_catalog/apps/catalog/content_metadata_utils.py
+++ b/enterprise_catalog/apps/catalog/content_metadata_utils.py
@@ -89,7 +89,7 @@ def get_course_first_paid_enrollable_seat_price(course):
     # Use advertised course run.
     # If that fails use one of the other active course runs.
     # (The latter is what Discovery does)
-    advertised_course_run = get_course_run_by_uuid(course, course.get('advertised_course_run_uuid'))
+    advertised_course_run = get_advertised_course_run(course)
     if advertised_course_run and advertised_course_run.get('first_enrollable_paid_seat_price'):
         return advertised_course_run.get('first_enrollable_paid_seat_price')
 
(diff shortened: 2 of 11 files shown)


enterprise_catalog/apps/catalog/models.py
```diff
@@ -555,10 +556,16 @@ def get_content_enrollment_url(self, content_metadata):
         else:
             # Catalog param only needed for legacy (non-learner-portal) enrollment URLs
             params['catalog'] = self.uuid
+
+            course_key = content_key if parent_content_key else None
+            course_run_key = content_key if not parent_content_key else None
```
iloveagent57: These two feel kind of backwards to me - if `not parent_content_key`, then we're dealing with a course, not a run, right? Maybe we can write it like this:
```python
course_run_key = content_key # set a fallback if we can't find the advertised run, or if content_key is already a run
if not parent_content_key:
    if advertised_course_run := get_advertised_course_run(content_metadata.json_metadata):
        course_run_key = advertised_course_run['key']
url = '{}/enterprise/{}/course/{}/enroll/'.format(
    settings.LMS_BASE_URL,
    self.enterprise_uuid,
    content_key,
    course_run_key,
)
```

enterprise_catalog/apps/catalog/utils.py
```diff
@@ -176,3 +176,20 @@ def get_course_run_by_uuid(course, course_run_uuid):
     except IndexError:
         return None
     return course_run
+
+
+def get_advertised_course_run(course):
```
adamstankiewicz: With the addition of this method, there are now two definitions of `get_advertised_course_run`. Should the other usage be updated to utilize this added method and remove the previously implemented method from `algolia_utils.py`? I believe the methods are identical.

brobro10000: Ok, after going through the code a bit more, I noticed a difference in definition of the `algolia_utils` version of `get_advertised_course_run` and the `utils` version of `get_advertised_course_run`. The updated PR should handle both instances, though feedback on my methodology in how I handled it would be appreciated. The intent of my implementation is to have the least amount of impact to the existing usages of `get_advertised_course_run` while retaining its intended function definition in `algolia_utils.py`

When I initially worked on this PR,  I attempted to use the `get_advertised_course_run` from `algolia_utils` but ran into a cyclical dependency (hence the duplicative code). Upon cleanup of this PR I noticed a second duplicative function (`get_course_run_by_uuid`) that is also cleaned up. Since the defined version of `get_course_run_by_uuid` in `enterprise_catalog/apps/catalog/content_metadata_utils.py`  has 10 usages currently, I opted to make this instance of the function the new source of truth, and migrate `get_advertised_course_run` from `utils.py` to `content_metadata_utils.py`.

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -1179,10 +1181,9 @@ def get_advertised_course_run(course):
         dict: containing key, pacing_type, start, end, and upgrade deadline
         for the course_run, or None
     """
-    full_course_run = get_course_run_by_uuid(course, course.get('advertised_course_run_uuid'))
-    if full_course_run is None:
-        return None
-    return _get_course_run(course, full_course_run)
+    if advertised_course_run := imported_get_advertised_course_run(course):
+        return _get_course_run(course, advertised_course_run)
+    return advertised_course_run
```
adamstankiewicz: We should remove this `get_advertised_course_run` method within `algolia_utils.py` altogether. Anywhere it's currently used (e.g., `get_upcoming_course_runs`, `_algolia_object_from_product`, etc.) should be using the other `get_advertised_course_run` you've introduced in the more "globally accessible" `content_metadata_utils.py` file instead.

tl;dr; there should only be one instance of `def get_advertised_course_run`.

enterprise_catalog/apps/catalog/content_metadata_utils.py
```diff
@@ -89,7 +89,7 @@ def get_course_first_paid_enrollable_seat_price(course):
     # Use advertised course run.
     # If that fails use one of the other active course runs.
     # (The latter is what Discovery does)
-    advertised_course_run = get_course_run_by_uuid(course, course.get('advertised_course_run_uuid'))
+    advertised_course_run = get_advertised_course_run(course)
```
adamstankiewicz, line 92: [consideration] It looks like there's a few other usages of `get_course_run_by_uuid` to get the advertised course run. Did we want to replace usages of `get_course_run_by_uuid` in favor of `get_advertised_course_run` in these other spots, too for consistency throughout? For example, within `_should_index_course`:

```python
advertised_course_run = get_course_run_by_uuid(
  course_json_metadata,
  advertised_course_run_uuid,
)
```

brobro10000: Updated 👍🏽

enterprise_catalog/apps/catalog/tests/test_algolia_utils.py
```diff
@@ -508,7 +509,8 @@ def test_get_advertised_course_run(self, searchable_course, expected_course_run)
         Assert get_advertised_course_runs fetches just enough info about advertised course run
         """
         advertised_course_run = utils.get_advertised_course_run(searchable_course)
-        assert advertised_course_run == expected_course_run
+        transformed_advertised_course_run = _get_course_run(searchable_course, advertised_course_run)
+        assert transformed_advertised_course_run == expected_course_run
```
adamstankiewicz: Now that `get_advertised_course_run` is no longer defined within `algolia_utils.py`, in favor of `content_metadata_utils.py`, should we move/update this test case to `test_content_metadata_utils.py` instead?

As is, the `utils.get_advertised_course_run` is actually importing `get_advertised_course_run` from `content_metadata_utils.py` despite the `utils` here referencing `algolia_utils.py`.

enterprise_catalog/apps/catalog/tests/test_content_metadata_utils.py
```diff
@@ -51,3 +58,161 @@ def test_tansform_force_included_courses(self):
...
+            }
+        )
+    )
+    @ddt.unpack
+    def test_get_advertised_course_run(self, searchable_course, expected_course_run):
+        """
+        Assert get_advertised_course_runs fetches just enough info about advertised course run
+        """
+        advertised_course_run = get_advertised_course_run(searchable_course)
+        transformed_advertised_course_run = _get_course_run(searchable_course, advertised_course_run)
```
adamstankiewicz: Given the base `get_advertised_course_run` returns the full course run, I'm curious why this test would need to transform as if we're in an Algolia context? Should the test assertions be for the full course run?

brobro10000: I still think its important to retain the test related to the transform of `get_advertised_course_run` since it validates one of the key usages of this function which is to serialize the algolia results for `advertised_course_run`

That being said, I added an additional test to validate `get_advertised_course_run` 👍🏽

enterprise_catalog/apps/catalog/tests/test_content_metadata_utils.py
```diff
@@ -51,3 +57,84 @@ def test_tansform_force_included_courses(self):
...
+                'course_runs': [],
+                'advertised_course_run_uuid': None
+            },
+            None
+        ),
+    )
+    @ddt.unpack
+    def test_get_advertised_course(self, searchable_course, expected_course_run):
+        """
+        Assert get_advertised_course_runs fetches the expected_course_run
```
adamstankiewicz: nit:

```suggestion
        Assert get_advertised_course_run fetches the expected_course_run
```

enterprise_catalog/apps/catalog/tests/test_content_metadata_utils.py
```diff
@@ -51,3 +57,84 @@ def test_tansform_force_included_courses(self):
         courses = [content_metadata]
         tansform_force_included_courses(courses)
         assert courses[0]['course_runs'][0]['status'] == 'published'
+
+    @ddt.data(
+        (
+            {
```
adamstankiewicz: nit: Might be helpful to add helper comments above each test case to describe their intent for improved readability

enterprise_catalog/apps/catalog/tests/test_content_metadata_utils.py
```diff
@@ -51,3 +57,84 @@ def test_tansform_force_included_courses(self):
...
+                'min_effort': 10,
+                'max_effort': 14,
+                'weeks_to_complete': 13,
+                'status': 'published',
+                'is_enrollable': True,
+                'is_marketable': True,
+                'enrollment_start': '2013-10-01T14:00:00Z',
+            },
+        ),
+        (
```
adamstankiewicz, line 119: [consideration] Is it also worth having a case where there are `course_runs`, but `advertised_course_run_uuid` is `None` (as is, the only case where `advertised_course_run_uuid': None` is with an empty list of `course_runs`).

enterprise_catalog/apps/catalog/tests/test_models.py
```diff
@@ -623,6 +623,46 @@ def test_get_content_enrollment_url(
...
+        parent_content_key,
+    ):
+        enterprise_uuid = uuid4()
+        enterprise_slug = 'sluggy'
+
+        enterprise_catalog = factories.EnterpriseCatalogFactory(enterprise_uuid=enterprise_uuid)
+        content_metadata = factories.ContentMetadataFactory(
+            content_key=content_key,
+            parent_content_key=parent_content_key,
+            content_type=COURSE,
```
adamstankiewicz: Should `content_type=COURSE` change, too, depending on the inputs of `content_key` and `parent_content_key`? For example, if `parent_content_key` is truthy, `content_type` must be a `COURSE_RUN`?

adamstankiewicz (approved):
LGTM.

PR #1026, opened 2025-01-21
feat: add parent_course query parameter to basic content-metadata endpoint

pwnage101:
ENT-9840

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -2123,6 +2123,59 @@ class AcademiesViewSetTests(APITestMixin):
         self.assertEqual(response.data['count'], 0)
 
 
+def ddt_cross_product(data_x, data_y):
+    """
+    Given two lists of test data dicts, produce a flat list of test data dicts
+    comprising every test scenario in x crossed with with every test scenario
+    of in y.
+
+    Demo:
+        Invoking this:
+
+        ddt_cross_product(
+            [
+                {'date': 'past'},
+                {'date': 'future'},
+            ],
+            [
+                {'size': 'big'},
+                {'size': 'small'},
+            ],
+        )
+
+        Returns this:
+
+        [
+            {'date': 'past', 'size': 'big'},
+            {'date': 'past', 'size': 'small'},
+            {'date': 'future', 'size': 'big'},
+            {'date': 'future', 'size': 'small'}
+        ]
+
+    Usage:
+        @ddt.data(*ddt_cross_product(
+            [
+                {'date': 'past'},
+                {'date': 'future'},
+            ],
+            [
+                {'size': 'big'},
+                {'size': 'small'},
+            ],
+        ))
+        def test_things(self, date, size):
+            pass
+
+    Args:
+        data_x (list of dict): First ddt data args.
+        data_y (list of dict): Second ddt data args.
+
+    Returns:
+        list of dict: The result of x cross y. Elements usable as arguments for ``@ddt.data()``.
+    """
+    return [x | y for x in data_x for y in data_y]
+
+
 @ddt.ddt
 class ContentMetadataViewTests(APITestMixin):
     """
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/v1/views/content_metadata.py
```diff
@@ -47,16 +51,72 @@ class ContentMetadataView(viewsets.ReadOnlyModelViewSet):
...
-                queryset = queryset.filter(content_key__in=content_filters)
-            if content_uuids:
-                queryset = queryset.filter(content_uuid__in=content_filters)
+            queryset_direct = self.queryset.filter(
+                Q(content_uuid__in=content_uuids) | Q(content_key__in=content_keys)
+            )
+            queryset = queryset_direct
+
+        # If ``?parent_course=true`` was passed, exclude course runs.
+        if self.request.query_params.get('parent_course', False):
```
iloveagent57: maybe just "exclude_course_runs" for the parameter name?

enterprise_catalog/apps/api/v1/views/content_metadata.py
```diff
@@ -47,16 +51,72 @@ class ContentMetadataView(viewsets.ReadOnlyModelViewSet):
...
+        """
+        Override to 
+        """
+        # Support alternative pk types besisdes just the raw object IDs (which are completely opaque
+        # to API clients).
+        obj = None
+        if self.pk_is_content_uuid:
+            obj = get_object_or_404(self.queryset, content_uuid=pk)
+        if self.pk_is_content_key:
+            obj = get_object_or_404(self.queryset, content_key=pk)
```
iloveagent57, line 118: maybe we want to do the fallback to raw object IDs here?
```python
else:
    obj = get_object_or_404(self.queryset, id=pk)
```

enterprise_catalog/apps/api/v1/views/content_metadata.py
```diff
@@ -47,16 +51,72 @@ class ContentMetadataView(viewsets.ReadOnlyModelViewSet):
...
+        if self.request.query_params.get('parent_course', False):
+            query_filters = ~Q(content_type=COURSE_RUN)
+            # If ``?content_identifiers=`` was passed, follow any matched course run objects back up
+            # to their parent courses and include those in the response.
+            if content_filters:
+                parent_content_keys = (
+                    record[0] for record in
+                    queryset_direct.filter(content_type=COURSE_RUN).values_list('parent_content_key')
+                )
+                all_content_keys_to_find = content_keys+parent_content_keys
```
iloveagent57: Is it possible to reach this point without passing in `content_identifiers`? If so, might be good to default `content_keys` to an empty list outside of this block.

enterprise_catalog/apps/api/v1/views/content_metadata.py
pwnage101: Originally struck this down for performance reasons. But, now that I think about it, our consumers don't even use this endpoint with raw object IDs so it's probably not worth the performance benefits to make the code harder to read.

enterprise_catalog/apps/api/v1/views/content_metadata.py
pwnage101: `exclude_course_runs` suggests the only behavior change is that course runs are excluded.  But that is untrue for retrieve(), in which case requested course runs are converted to courses (rather than 404 which is what I'd expect based on the name `exclude_course_runs).

That said, I'm still open to alternative names.  Maybe "coerce_to_parent_course"?

enterprise_catalog/apps/api/v1/tests/test_views.py
```diff
@@ -2143,48 +2161,105 @@ def test_list_success(self):
...
+            {'request_content_type': COURSE_RUN, 'coerce_to_parent_course': False, 'expect_content_type': COURSE_RUN},
+            {'request_content_type': COURSE_RUN, 'coerce_to_parent_course': True, 'expect_content_type': COURSE},
+        ],
+        # Repeat ever test above given different types of identifier input types.
+        [
+            {'request_by_field': 'id'},
+            {'request_by_field': 'content_key'},
+            {'request_by_field': 'content_uuid'},
+        ],
+    ))
```
brobro10000, line 2284: Very cool implementation, if I am understanding correctly, this ends up being 12 test cases total?

brobro10000 (approved):
Great work. Approving to unblock. Very cool implementation using the test helper `ddt_cross_product`. I learned something new today!

enterprise_catalog/apps/api/v1/tests/test_views.py
pwnage101: Yes, 12!

PR #1029, opened 2025-01-27
feat: mark course as active in Agolia if is_marketable_external

hamzawaleed01:
Updated the utility functions to ensure that course runs with the `is_marketable_external` property set to `True` are indexed as `active` in Algolia. This change affects the logic in the `is_course_run_active` function, which now includes a check for the `is_marketable_external` property.

### DIFF
--- a/enterprise_catalog/apps/api/v1/utils.py
+++ b/enterprise_catalog/apps/api/v1/utils.py
@@ -10,6 +10,10 @@ from six.moves.urllib.parse import (
     urlunsplit,
 )
 
+from enterprise_catalog.apps.catalog.content_metadata_utils import (
+    is_course_run_active,
+)
+
 
 logger = logging.getLogger(__name__)
 
@@ -57,27 +61,6 @@ def get_enterprise_utm_context(enterprise_name):
         utm_context['utm_source'] = slugify(enterprise_name)
 
     return utm_context
-
-
-def is_course_run_active(course_run):
-    """
-    Checks whether a course run is active. That is, whether the course run is published,
-    enrollable, and marketable.
-
-    Arguments:
-        course_run (dict): The metadata about a course run.
-
-    Returns:
-        bool: True if course run is "active"
-    """
-    is_enrollable = course_run.get('is_enrollable', False)
-    if course_run.get("is_marketable_external") and is_enrollable:
-        return True
-    course_run_status = course_run.get('status') or ''
-    is_published = course_run_status.lower() == 'published'
-    is_marketable = course_run.get('is_marketable', False)
-
-    return is_published and is_enrollable and is_marketable
 
 
 def is_any_course_run_active(course_runs):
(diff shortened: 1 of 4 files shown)


sameeramin (approved):
LGTM 👍🏼

enterprise_catalog/apps/catalog/content_metadata_utils.py
```diff
@@ -71,9 +71,11 @@ def is_course_run_active(course_run):
     Returns:
         bool: True if course run is "active"
     """
+    is_enrollable = course_run.get('is_enrollable', False)
+    if course_run.get("is_marketable_external") and is_enrollable:
+        return True
```
iloveagent57: Do we also need the check for `is_published` here? For example, if a course run's status is "archived", we wouldn't want to consider that course run as active, even if it is marketable and enrollable (although I guess archived course runs shouldn't be enrollable).

hamzawaleed01: Seems like we're already taking care of `status` in Discovery service.
https://github.com/openedx/course-discovery/blob/master/course_discovery/apps/course_metadata/models.py#L2988

iloveagent57 (approved):

enterprise_catalog/apps/catalog/content_metadata_utils.py
pwnage101: I think it would be helpful to add a code comment to explain why checking is_published is redundant and not needed in this function.

enterprise_catalog/apps/catalog/content_metadata_utils.py
```diff
@@ -71,9 +71,11 @@ def is_course_run_active(course_run):
...
     course_run_status = course_run.get('status') or ''
     is_published = course_run_status.lower() == 'published'
-    is_enrollable = course_run.get('is_enrollable', False)
     is_marketable = course_run.get('is_marketable', False)
 
     return is_published and is_enrollable and is_marketable
```
pwnage101: Nit: instead of returning early, I think I'd rather see this pattern which should behave the same:
```suggestion
    is_marketable_internal = is_published and is_marketable
    return is_enrollable and (is_marketable_external or is_marketable_internal)
```

enterprise_catalog/apps/api/v1/utils.py
```diff
@@ -70,9 +70,11 @@ def is_course_run_active(course_run):
...
     """
+    is_enrollable = course_run.get('is_enrollable', False)
+    if course_run.get("is_marketable_external") and is_enrollable:
+        return True
     course_run_status = course_run.get('status') or ''
     is_published = course_run_status.lower() == 'published'
-    is_enrollable = course_run.get('is_enrollable', False)
     is_marketable = course_run.get('is_marketable', False)
 
     return is_published and is_enrollable and is_marketable
```
pwnage101: Nit: Is now a good time to remove one of these duplicate functions?

pwnage101 (approved):
non-blocking nits!

PR #1039, opened 2025-02-28
feat: add API endpoint to retrieve secured Algolia API key for an EnterpriseCustomer

adamstankiewicz:
**Ticket: [ENT-10079](https://2u-internal.atlassian.net/browse/ENT-10079)**

Adds a new API `GET /api/v1/enterprise-customer/<enterprise_uuid>/secured-algolia-api-key/` to return the following response structure:

```json
{
  "algolia": {
    "secured_api_key": "...",
    "valid_until": "2025-02-28T12:00:00Z"
  },
  "catalog_uuids_to_catalog_query_uuids": {
    "catalog_uuid_1": "catalog_query_uuid_1",
    "catalog_uuid_2": "catalog_query_uuid_2"
  }
}
```

When using this secured Algolia API key to filter Algolia search results for all content available to an enterprise customer, no additional client-side `filters` are needed to be passed as the generated secured Algolia API key itself baked in filtering search results by any `CatalogQuery` associated with the customers' `EnterpriseCatalog` records.

By default, this secured Algolia API key is valid for 1 hour, though it is configurable via the Django setting `SECURED_ALGOLIA_API_KEY_EXPIRATION`. The expiration is returned as an ISO timestamp.

Also included in this API response is a mapping of `EnterpriseCatalog` uuids to their associated `CatalogQuery` uuids. These mappings would be used by frontends when further filtering down all content available to the enterprise customer to a specific catalog, i.e.:
* Determine enterprise catalog UUID associated with a subsidy (e.g., subscription license).
* Find catalog query UUID associated with the above enterprise catalog UUID (discerned through these mappings).
* Add additional `filter` on the frontend for the found catalog query UUID.

image

## Post-review

### DIFF
--- a/.github/workflows/mysql-migrations-check.yml
+++ b/.github/workflows/mysql-migrations-check.yml
@@ -36,7 +36,7 @@ jobs:
         echo "dir=$(pip cache dir)" >> $GITHUB_OUTPUT
     - name: Cache pip dependencies
       id: cache-dependencies
-      uses: actions/cache@v2
+      uses: actions/cache@v4
       with:
         path: ${{ steps.pip-cache-dir.outputs.dir }}
         key: ${{ runner.os }}-pip-${{ hashFiles('requirements/pip_tools.txt') }}
--- a/enterprise_catalog/apps/api/v1/views/enterprise_customer.py
+++ b/enterprise_catalog/apps/api/v1/views/enterprise_customer.py
@@ -6,10 +6,10 @@ from django.core.exceptions import ImproperlyConfigured
 from django.utils.decorators import method_decorator
 from drf_spectacular.utils import OpenApiExample, extend_schema
 from edx_rbac.utils import get_decoded_jwt
+from rest_framework import status
 from rest_framework.decorators import action
 from rest_framework.exceptions import NotFound, PermissionDenied
 from rest_framework.response import Response
-from rest_framework.status import HTTP_400_BAD_REQUEST
 
 from enterprise_catalog.apps.api.v1.decorators import (
     require_at_least_one_query_parameter,
@@ -119,7 +119,7 @@ class EnterpriseCustomerViewSet(BaseViewSet):
             )
             return Response(
                 f'Error: invalid enterprice customer uuid: "{enterprise_uuid}" provided.',
-                status=HTTP_400_BAD_REQUEST
+                status=status.HTTP_400_BAD_REQUEST
             )
         customer_catalogs = EnterpriseCatalog.objects.filter(enterprise_uuid=enterprise_uuid)
 
@@ -268,7 +268,7 @@ class EnterpriseCustomerViewSet(BaseViewSet):
             }
             return Response(
                 SecuredAlgoliaAPIKeyErrorResponseSerializer(error_response).data,
-                status=HTTP_400_BAD_REQUEST,
+                status=status.HTTP_404_NOT_FOUND,
             )
 
         catalog_query_uuids = [
@@ -286,7 +286,7 @@ class EnterpriseCustomerViewSet(BaseViewSet):
             }
             return Response(
                 SecuredAlgoliaAPIKeyErrorResponseSerializer(error_response).data,
-                status=HTTP_400_BAD_REQUEST,
+                status=status.HTTP_422_UNPROCESSABLE_ENTITY,
             )
 
         # Create a mapping of catalog UUIDs to their corresponding catalog query UUIDs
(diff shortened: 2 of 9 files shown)


enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -209,3 +217,122 @@ def content_metadata(self, customer_uuid, content_identifier, **kwargs):  # pyli
...
+            logger.error(
+                f'No enterprise catalogs found for the given enterprise UUID: {enterprise_uuid}.'
+            )
+            error_response = {
+                'user_message': algolia_api_key_error_message,
+                'developer_message': 'No enterprise catalogs found for the specified enterprise customer.',
+            }
+            return Response(
+                SecuredAlgoliaAPIKeyErrorResponseSerializer(error_response).data,
+                status=HTTP_400_BAD_REQUEST,
```
iloveagent57: minor quibble: maybe 404 is more appropriate in this particular case?

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
```diff
@@ -209,3 +217,122 @@ def content_metadata(self, customer_uuid, content_identifier, **kwargs):  # pyli
...
+            logger.exception(
+                f'Secured Algolia API key generation failed for enterprise UUID {enterprise_uuid}'
+            )
+            error_response = {
+                'user_message': algolia_api_key_error_message,
+                'developer_message': exc,
+            }
+            return Response(
+                SecuredAlgoliaAPIKeyErrorResponseSerializer(error_response).data,
+                status=HTTP_400_BAD_REQUEST,
```
iloveagent57: should these maybe be 422s?

enterprise_catalog/apps/api_client/algolia.py
```diff
@@ -164,12 +194,75 @@ def remove_objects(self, object_ids):
...
+        if not self.algolia_search_api_key:
+            logger.error(
+                'Could not generate secured Algolia API key due to missing Algolia settings: %s',
+                'SEARCH_API_KEY',
+            )
+            raise ImproperlyConfigured(
+                'Cannot generate secured Algolia API key without the ALGOLIA.SEARCH_API_KEY in settings.'
+            )
+
+        expiration_time = getattr(settings, 'SECURED_algolia_api_key_EXPIRATION', 3600)  # Default to 1 hour
```
iloveagent57: what's up with the mixed casing in `SECURED_algolia_api_key_EXPIRATION`

enterprise_catalog/apps/api_client/algolia.py
```diff
@@ -164,12 +194,75 @@ def remove_objects(self, object_ids):
...
+        valid_until = int(time.time()) + expiration_time
```
iloveagent57: you can use `timezone.now()` from django to do this in a timezone-aware way: https://docs.djangoproject.com/en/5.1/topics/i18n/timezones/#naive-and-aware-datetime-objects
so like 
```python
valid_until = timezone.now() + timedelta(seconds=expiration_time)
iso_format = "%Y-%m-%dT%H:%M:%SZ"
valid_until_iso = valid_until.strftime(iso_format)
```

iloveagent57 (commented):
This is great, couple of minor suggestions!

enterprise_catalog/apps/api_client/algolia.py
adamstankiewicz: LOL, good catch. This line must have caught up in the lowercasing of the `ALGOLIA_API_KEY` property in my find/replace. It should be all uppercase, will fix!

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
adamstankiewicz: I agree, 404 does seem reasonable here; will update!

enterprise_catalog/apps/api/v1/views/enterprise_customer.py
adamstankiewicz: Agreed, 422s vs. 400s are probably more appropriate, given the request is technically valid, but unprocessable. Will update!

enterprise_catalog/apps/api_client/algolia.py
adamstankiewicz: Ah, as I was making this change, I realized I could likely also use `localized_utcnow() + timedelta(seconds=expiration_time)` as well. However, `localized_utcnow` uses a deprecated `datetime.utcnow()` method.

I've updated `localized_utcnow` in this PR to avoid the deprecated method:

```py
def localized_utcnow():
    """Helper function to return localized utcnow()."""
    return datetime.now(timezone.utc)
```

Edit: the revised `localized_utcnow` is per the suggestion from the `datetime.utcnow()` deprecation warning.

.github/workflows/mysql-migrations-check.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         echo "dir=$(pip cache dir)" >> $GITHUB_OUTPUT
     - name: Cache pip dependencies
       id: cache-dependencies
-      uses: actions/cache@v2
+      uses: actions/cache@v4
```
adamstankiewicz, line 39: [inform] CI ~is currently~ was failing its migrations checks due to `actions/cache@v2` no longer being supported, requiring an upgrade to either v3 or v4.

iloveagent57 (approved):

PR #1041, opened 2025-03-04
fix: allow periods in pk lookup_value_regex for ContentMetadataView

adamstankiewicz:
https://2u-internal.atlassian.net/browse/ENT-10109

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views/content_metadata.py
+++ b/enterprise_catalog/apps/api/v1/views/content_metadata.py
@@ -43,6 +43,7 @@ class ContentMetadataView(viewsets.ReadOnlyModelViewSet):
     """
     View for retrieving and listing base content metadata.
     """
+    lookup_value_regex = r'[^/]+'  # Allow any non-slash character in the pk (including periods).
     serializer_class = ContentMetadataSerializer
     authentication_classes = [JwtAuthentication, SessionAuthentication]
     permission_classes = [permissions.IsAuthenticated]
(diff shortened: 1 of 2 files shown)


enterprise_catalog/apps/api/v1/views/content_metadata.py
```diff
@@ -43,6 +43,7 @@ class ContentMetadataView(viewsets.ReadOnlyModelViewSet):
     """
     View for retrieving and listing base content metadata.
     """
+    lookup_value_regex = r'[^/]+'  # Allow any non-slash character in the pk (including periods).
```
brobro10000, line 46: [curious] Would this regular expression also pick up the query parameter? Could we add a test to validate if query parameters are being picked up?

Screenshot 2025-03-04 at 8 53 09 AM

Here is a regular expression that only explicitly grabs until the slash or a query parameter

`^[^/?]+`

Screenshot 2025-03-04 at 8 56 21 AM

adamstankiewicz: Yes, this regex handles query params as expected. It's not much different than the default DRF regex `[^/.]+`; only difference is the removal of the period (`.`).

image

Other existing tests within `ContentMetadataViewTests` already cover query params like `?coerce_to_parent_course=true`.

brobro10000 (approved):

iloveagent57 (approved):
For posterity, here's the docs on `lookup_value_regex`: https://www.django-rest-framework.org/api-guide/routers/#using-django-path-with-routers

PR #1053, opened 2025-04-10
docs: add schema documentation for GET /enterprise-catalogs/{catalog_id}

ahmad-hassan-dev:
Improves documentation for the GET /enterprise/v2/enterprise-catalogs/{uuid} endpoint. This change enhances the OpenAPI schema by:
-   Explicitly documenting the 'page' and 'page_size' query parameters via the @extend_schema decorator.

#### Jira:
[limit parameter is not working on the catalog discovery query](https://2u-internal.atlassian.net/browse/ENT-10015)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
+++ b/enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
@@ -3,6 +3,7 @@ from collections import OrderedDict
 
 from django.shortcuts import get_object_or_404
 from django.utils.functional import cached_property
+from drf_spectacular.utils import OpenApiParameter, extend_schema
 from rest_framework.decorators import action
 from rest_framework.generics import GenericAPIView
 from rest_framework.renderers import JSONRenderer
@@ -73,6 +74,25 @@ class EnterpriseCatalogGetContentMetadata(BaseViewSet, GenericAPIView):
         response.data['enterprise_customer'] = self.enterprise_catalog.enterprise_uuid
         return response
 
+    @extend_schema(
+        description="GET calls to the `enterprise-catalogs/{catalog_id}` endpoint return a list of all of the active courses in a specified course catalog."
+        "You can then make a GET call to the `/enterprise-catalogs/{catalog_id}/courses/{course_key}` endpoint to return details about a single course.",
+        parameters=[
+            OpenApiParameter(
+                name="page",
+                type=int,
+                location=OpenApiParameter.QUERY,
+                description="A page number within the paginated result.",
+            ),
+            OpenApiParameter(
+                name="page_size",
+                type=int,
+                location=OpenApiParameter.QUERY,
+                description="Number of results to return per page.",
+            ),
+        ],
+        responses={200: ContentMetadataSerializer(many=True)},
+    )
     @action(detail=True)
     def get(self, request, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


sameeramin (approved):
LGTM 👍🏼

enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
```diff
@@ -73,6 +74,28 @@ def get_response_with_enterprise_fields(self, response):
         response.data['enterprise_customer'] = self.enterprise_catalog.enterprise_uuid
         return response
 
+    @extend_schema(
```
sameenfatima78, line 75: Great work! Would it be possible to also mention the limitations of the page and page_size query parameters in the description, such as the maximum allowed values... and default values?

sameenfatima78: The API also accepts "content_keys" query param which seems to be missing from the API schema and also the docs.

hamzawaleed01: ++ we should also mention the default `page_size` if it's possible.

hamzawaleed01 (approved):

enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
```diff
@@ -73,6 +72,38 @@ def get_response_with_enterprise_fields(self, response):
...
+                type=int,
+                location=OpenApiParameter.QUERY,
+                description="A page number within the paginated result.",
+            ),
+            OpenApiParameter(
+                name="page_size",
+                type=int,
+                location=OpenApiParameter.QUERY,
+                description=f"Number of results to return per page. Defaults to {DefaultPagination.page_size}. "
+                f"Maximum value is {DefaultPagination.max_page_size}.",
```
sameenfatima78, line 102: is the `max_page_size` defined in settings?

enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
```diff
@@ -27,7 +26,7 @@ class EnterpriseCatalogGetContentMetadata(BaseViewSet, GenericAPIView):
     serializer_class = ContentMetadataSerializer
     renderer_classes = [JSONRenderer, XMLRenderer]
     lookup_field = 'uuid'
-    pagination_class = PageNumberWithSizePagination
+    pagination_class = DefaultPagination
```
sameenfatima78, line 29: Just took a look at the codebase and found out that there is another endpoint defined "`get_content_metadata`" defined in the view which will also be impacted with this change. Have we tested that endpoint?

enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
ahmad-hassan-dev: No, `max_page_size` isn’t defined in the settings — it’s explicitly set as a class attribute inside the [DefaultPagination](https://github.com/openedx/edx-drf-extensions/blob/master/edx_rest_framework_extensions/paginators.py#L9) class itself. So unless overridden somewhere else, the maximum value will be 100.

enterprise_catalog/apps/api/v1/views/enterprise_catalog_get_content_metadata.py
ahmad-hassan-dev: If we're referring to the same view (`EnterpriseCatalogGetContentMetadata`), the `get` method is the primary API entry point, while the `get_content_metadata` method serves as a helper to encapsulate the core data retrieval and formatting logic.

sameenfatima78 (approved):
Looks good!

PR #1054, opened 2025-04-13
feat: generate video title if title is missing

irfanuddinahmad:
## Post-review

### DIFF
--- a/enterprise_catalog/apps/video_catalog/management/commands/generate_video_transcript_summary.py
+++ b/enterprise_catalog/apps/video_catalog/management/commands/generate_video_transcript_summary.py
@@ -37,7 +37,7 @@ class Command(BaseCommand):
         videos = Video.objects.exclude(edx_video_id__in=processed_videos)
         for video in videos:
             try:
-                summary = generate_transcript_summary(video)
+                summary = generate_transcript_summary(video.json_metadata['transcript_urls'])
                 if summary:
                     VideoTranscriptSummary.objects.create(video=video, summary=summary)
             except Exception as ex:  # pylint: disable=broad-exception-caught
--- a/enterprise_catalog/apps/video_catalog/utils.py
+++ b/enterprise_catalog/apps/video_catalog/utils.py
@@ -12,8 +12,12 @@ from rest_framework.exceptions import ValidationError
 from enterprise_catalog.apps.ai_curation.openai_client import chat_completions
 from enterprise_catalog.apps.api_client.discovery import DiscoveryApiClient
 from enterprise_catalog.apps.api_client.studio import StudioApiClient
+from enterprise_catalog.apps.api_client.xpert_ai import chat_completion
 from enterprise_catalog.apps.catalog.constants import COURSE_RUN
 from enterprise_catalog.apps.catalog.models import ContentMetadata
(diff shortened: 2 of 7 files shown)


enterprise_catalog/apps/video_catalog/utils.py
```diff
@@ -51,6 +52,11 @@ def fetch_course_video_metadata(course_run_key, video_usage_key, video_title):
                             )
                         }
                     )
+                    if not video_title:
+                        video_title = generate_video_title(video)
+                    if video_title:
+                        video.title = video_title
+                        video.save()
```
saleem-latif: I think it would be better to make this part of the `update_or_create` part instead of making a second DB call to update the title. Something like,

```
video, __ = Video.objects.update_or_create(
    edx_video_id=video_data['edx_video_id'],
    defaults={
        'client_video_id': video_data['client_video_id'],
        'json_metadata': video_data,
        'video_usage_key': video_usage_key,
        'title': video_title or generate_video_title(video),
        'parent_content_metadata': ContentMetadata.objects.get(
            content_key=course_run_key, content_type=COURSE_RUN
        )
    }
)
```

enterprise_catalog/apps/video_catalog/utils.py
```diff
@@ -142,3 +148,30 @@ def generate_transcript_summary(video, language='en'):
     transcript_url = video.json_metadata['transcript_urls'].get(language)
     transcript = fetch_transcript(transcript_url, include_time_markings=False)
     return get_transcript_summary(transcript[:settings.MAX_TRANSCRIPT_LENGTH])
+
+
+def generate_video_title(video, language='en', max_length: int = 60) -> str:
+    """
+    Generate a title of the video
```
saleem-latif: nit: Generate a title for the video from its transcript.

enterprise_catalog/apps/video_catalog/utils.py
```diff
@@ -142,3 +148,30 @@ def generate_transcript_summary(video, language='en'):
...
+
+    Arguments:
+        video (Video): Video instance.
+        language (str): Title language.
+
+    Returns:
+        (str): The title of the video.
+    """
+    transcript_summary = generate_transcript_summary(video, language)
```
saleem-latif: nit: I think it would be better to skip this variable declaration and directly use `generate_transcript_summary(video, language)` below.

saleem-latif (changes requested):

enterprise_catalog/apps/video_catalog/management/commands/generate_video_transcript_summary.py
```diff
@@ -37,7 +37,7 @@ def handle(self, *args, **options):
         videos = Video.objects.exclude(edx_video_id__in=processed_videos)
         for video in videos:
             try:
-                summary = generate_transcript_summary(video)
+                summary = generate_transcript_summary(video.json_metadata['transcript_urls'])
```
irfanuddinahmad, line 40: This is also used in https://github.com/openedx/enterprise-catalog/blob/5ad25546df00342b80048d8c02faede8182a62a5/enterprise_catalog/apps/video_catalog/management/commands/generate_video_transcript_summary.py#L40

saleem-latif (approved):

PR #1061, opened 2025-04-28
Fix  add task deleting old tmp indices

jesperhodge:
## Internal JIRA (2U):

https://2u-internal.atlassian.net/browse/ENT-9872

## Description:

Algolia gets more and more catalog tmp indices over time. This is due to Algolia's behavior to keep tmp indices when running `replace_all_objects` and encountering errors, which happens sometimes. This method is used by us when reindexing.

These tmp indices take up a large space in Algolia, and therefore should be regularly deleted.

This adds a management task that can delete these tmp indices. It filters for them by name to ensure only catalog tmp indices are deleted, and also only deletes tmp indices that are between 10 and 60 days old, to ensure that no indices that are new and in use are deleted.

## To Do:

Add Jenkins command to allow us to manually run this task (including dry run)

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -626,7 +626,7 @@ def _created_between(datestring, min_days_ago, max_days_ago):
     return False
 
 
-def _retrieve_inactive_indices(client):
+def _retrieve_inactive_tmp_indices(client):
         indices = client.list_indices().get('items', [])
         tmp_indices = filter(lambda x: x.get('name', '').startswith(f'{settings.ALGOLIA["INDEX_NAME"]}_tmp_'), indices)
         inactive_tmp_indices = filter(lambda x: _created_between(x.get('createdAt', None), 10, 60), tmp_indices)
@@ -656,7 +656,7 @@ def _delete_indices(client, indices, dry_run=True):
 
 @shared_task(base=LoggedTaskWithRetry, bind=True, default_retry_delay=UNREADY_TASK_RETRY_COUNTDOWN_SECONDS)
 @expiring_task_semaphore()
-def remove_old_temporary_catalog_indices_task(self, force=False, dry_run=False):  # pylint: disable=unused-argument
+def remove_old_temporary_catalog_indices_task(self, force=False, dry_run=True):  # pylint: disable=unused-argument
     """
     Remove old temporary catalog indices from Algolia.
 
@@ -665,7 +665,7 @@ def remove_old_temporary_catalog_indices_task(self, force=False, dry_run=False):
     This task removes all indices that are older than 10 days and newer than 60 days.
 
     Args:
-        force (bool): If true, forces execution of task and ignores time since last run.
+        force (bool): Not used.
         dry_run (bool): If true, does everything except call Algolia APIs.
     """
     client = None
@@ -684,7 +684,7 @@ def remove_old_temporary_catalog_indices_task(self, force=False, dry_run=False):
         raise exep
 
     try:
-        inactive_tmp_indices = _retrieve_inactive_indices(client)
+        inactive_tmp_indices = _retrieve_inactive_tmp_indices(client)
     except Exception as exep:
         logger.exception(
             f'Retrieving old tmp indices from Algolia failed. Error: {exep}'
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -6,7 +6,7 @@ import re
 import sys
 import time
 from collections import defaultdict
-from datetime import timedelta, datetime
+from datetime import datetime, timedelta
 from operator import itemgetter
 
 from algoliasearch.exceptions import AlgoliaException
@@ -32,9 +32,9 @@ from enterprise_catalog.apps.catalog.algolia_utils import (
     get_initialized_algolia_client,
     get_pathway_course_keys,
     get_pathway_program_uuids,
+    new_search_client_or_error,
     partition_course_keys_for_indexing,
     partition_program_keys_for_indexing,
-    new_search_client_or_error,
 )
 from enterprise_catalog.apps.catalog.constants import (
     COURSE,
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -650,7 +650,7 @@ def _delete_indices(client, indices, dry_run=True):
 
     if dry_run:
         logger.info('dry_run=true, not deleting old tmp indices from Algolia')
-        return indices
+        return
 
     for index_name in indices:
         try:
@@ -664,7 +664,6 @@ def _delete_indices(client, indices, dry_run=True):
             raise exep
 
         logger.info('Finished deleting indices from Algolia')
-    return indices
 
 
 @shared_task(base=LoggedTaskWithRetry, bind=True, default_retry_delay=UNREADY_TASK_RETRY_COUNTDOWN_SECONDS)
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -627,10 +627,10 @@ def _created_between(datestring, min_days_ago, max_days_ago):
 
 
 def _retrieve_inactive_tmp_indices(client):
-        indices = client.list_indices().get('items', [])
-        tmp_indices = filter(lambda x: x.get('name', '').startswith(f'{settings.ALGOLIA["INDEX_NAME"]}_tmp_'), indices)
-        inactive_tmp_indices = filter(lambda x: _created_between(x.get('createdAt', None), 10, 60), tmp_indices)
-        return list(map(lambda x: x.get('name', ''), inactive_tmp_indices))
+    indices = client.list_indices().get('items', [])
+    tmp_indices = filter(lambda x: x.get('name', '').startswith(f'{settings.ALGOLIA["INDEX_NAME"]}_tmp_'), indices)
+    inactive_tmp_indices = filter(lambda x: _created_between(x.get('createdAt', None), 10, 60), tmp_indices)
+    return list(map(lambda x: x.get('name', ''), inactive_tmp_indices))
 
 
 def _delete_indices(client, indices, dry_run=True):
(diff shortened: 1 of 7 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -613,6 +616,86 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):
...
+
+@shared_task(base=LoggedTaskWithRetry, bind=True, default_retry_delay=UNREADY_TASK_RETRY_COUNTDOWN_SECONDS)
+@expiring_task_semaphore()
+def remove_old_temporary_catalog_indices_task(self, force=False, dry_run=True):  # pylint: disable=unused-argument
+    """
+    Remove old temporary catalog indices from Algolia.
+
+    Because Algolia's `replace_all_objects` method generates but does not delete temporary indices
+    named as `<index_name>_tmp_<timestamp>`, we need to remove them periodically.
+    This task removes all indices that are older than 10 days and newer than 60 days.
```
marlonkeating, line 679: Should we make the time periods configurable for this job, as flags that can be passed in (with `--older_than` default being 10, and `--newer_than` default being 60)?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -613,6 +616,86 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):
...
+    inactive_tmp_indices = filter(lambda x: _created_between(x.get('createdAt', None), 10, 60), tmp_indices)
+    return list(map(lambda x: x.get('name', ''), inactive_tmp_indices))
+
+
+def _delete_indices(client, indices, dry_run=True):
+    logger.info('Index names to delete: %s', indices)
+
+    if dry_run:
+        logger.info('dry_run=true, not deleting old tmp indices from Algolia')
+        return indices
```
marlonkeating: nit: Why are we returning indices here?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -613,6 +616,86 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):
...
+    created_timestamp = datetime.strptime(datestring, '%Y-%m-%dT%H:%M:%S.%fZ').timestamp()
+    difference_in_days = (time.time() - created_timestamp) / (60 * 60 * 24)
+    if difference_in_days > min_days_ago and difference_in_days < max_days_ago:
+        return True
+    return False
+
+
+def _retrieve_inactive_tmp_indices(client):
+    indices = client.list_indices().get('items', [])
+    tmp_indices = filter(lambda x: x.get('name', '').startswith(f'{settings.ALGOLIA["INDEX_NAME"]}_tmp_'), indices)
```
marlonkeating: nit: This line is a bit dense, can we break the check for whether a name is to a temp index into its own helper method?

enterprise_catalog/apps/catalog/algolia_utils.py
```diff
@@ -376,6 +377,19 @@ def get_initialized_algolia_client():
...
+def new_search_client_or_error():
+    """
+    Returns a new Algolia search client that is not initialized to any specific index.
+    """
+    client = SearchClient.create(
+        settings.ALGOLIA.get('APPLICATION_ID', None),
+        settings.ALGOLIA.get('API_KEY', None)
+    )
+    if not client:
+        raise TypeError(f'client should be an Algolia search client, but was type {type(client)}.')
```
marlonkeating: If client return value was falsy, what useful type information are we hoping to see in it?

enterprise_catalog/apps/api/tasks.py
```diff
@@ -613,6 +616,86 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):
...
+def remove_old_temporary_catalog_indices_task(self, force=False, dry_run=True):  # pylint: disable=unused-argument
+    """
+    Remove old temporary catalog indices from Algolia.
+
+    Because Algolia's `replace_all_objects` method generates but does not delete temporary indices
+    named as `<index_name>_tmp_<timestamp>`, we need to remove them periodically.
+    This task removes all indices that are older than 10 days and newer than 60 days.
+
+    Args:
+        force (bool): Not used.
```
marlonkeating: Why are we including this as an argument if it's not used?

enterprise_catalog/apps/catalog/management/commands/remove_old_temporary_catalog_indices.py
```diff
@@ -0,0 +1,58 @@
...
+            )
+            logger.info(
+                'index_enterprise_catalog_in_algolia_task from command index_enterprise_catalog_in_algolia'
+                'finished successfully.'
+            )
+        except Exception as exc:  # pylint: disable=broad-except
+            # celery weirdly hijacks and prefixes the path of the below Exception
+            # with `celery.backends.base` when it's raised.
+            # So this block still catches only a specific error, just in a roundabout way.
+            if type(exc).__name__ != 'TaskRecentlyRunError':
```
marlonkeating, line 73: I don't see any checks in the task code that would raise this exception, is it something the framework does automatically?

enterprise_catalog/apps/api/tasks.py
jesperhodge: The return statement is a guard clause, making it so that in case of dry runs we never delete indices.
Returning indices that have been deleted makes the function more reusable.

enterprise_catalog/apps/api/tasks.py
jesperhodge: It seems to be used by some of the decorators.
The task seems to have some debouncing or time between allowed runs.
I have taken this from `reindex_algolia`, which also doesn't use `force` directly.
I assume this argument is needed for the decorators, but I have no idea about the details, so I keep it for safety.

enterprise_catalog/apps/catalog/algolia_utils.py
jesperhodge: The idea is that the algolia library should create a search client. It's very hard to figure out what the exact type that's returned by `create` should be, but that's fine. If `client` is anything like None - anything that's not a truthy class instance - then there must have been an error creating the search client.
I want to not just say "Failed to create search client" because I like more information. Thus, a type error with the type should be helpful when debugging.

enterprise_catalog/apps/catalog/management/commands/remove_old_temporary_catalog_indices.py
jesperhodge: It's something done by the decorator, I think. Copied this from `reindex_algolia` task.

enterprise_catalog/apps/api/tasks.py
marlonkeating: The returning part is fine, but if we're not using the return value and have no concrete plans to use it in the future, I would just omit it for now.  It's an easy add later if we need it, and a potential source of confusion now if we leave it.

enterprise_catalog/apps/catalog/management/commands/remove_old_temporary_catalog_indices.py
```diff
@@ -0,0 +1,79 @@
...
+            action='store_true',
+            default=False,
+            help='List algolia indices to be removed, but do not actually remove them.',
+        )
+        parser.add_argument(
+            '--min-days',
+            dest='min_days',
+            default=10,
+            type=int,
+            help='List algolia indices to be removed, but do not actually remove them.',
```
marlonkeating: Update help text for `--min-days`, `--max-days`

enterprise_catalog/apps/catalog/management/commands/remove_old_temporary_catalog_indices.py
```diff
@@ -0,0 +1,79 @@
...
+            dest='max_days',
+            default=60,
+            type=int,
+            help='List algolia indices to be removed, but do not actually remove them.',
+        )
+
+    def handle(self, *_args, **options):
+        """
+        Runs a celery task to remove leftover catalog `_tmp_` indices
+        that have a creation date between 60 and 10 days ago.
```
marlonkeating, line 50: nit: Update comment to reflect new configurable date range

enterprise_catalog/apps/catalog/algolia_utils.py
marlonkeating: I think it's fine to just return the `{client}` text here.  Any falsy values (e.g. `None`, `0`, `''`,`[]`,`{}`) should be pretty clear from the printed context without needing the type.

Checking the source code for Algolia search api, it doesn't look like they return any exotic falsy error objects as far as I can see: https://github.com/algolia/algoliasearch-client-python/blob/3.0.0/algoliasearch/search_client.py#L48

marlonkeating (changes requested):

enterprise_catalog/apps/api/tests/test_tasks.py
```diff
@@ -1260,6 +1261,242 @@ def mock_get_algolia_products_for_batch(
...
+                mock_client.init_index.assert_has_calls(
+                    [mock.call(index_name) for index_name in expected_indices_to_delete],
+                    any_order=True
+                )
+
+                assert mock_client.init_index.called
+
+                # Test that each index `.delete` was called for those
+                for index_name in expected_indices_to_delete:
+                    mock_client.init_index.return_value.delete.assert_any_call()
```
marlonkeating: I don't see `index_name` being used in each check, it looks like it's just making the same check every time.

enterprise_catalog/apps/catalog/management/commands/remove_old_temporary_catalog_indices.py
```diff
@@ -0,0 +1,79 @@
...
+    remove_old_temporary_catalog_indices_task,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
+    help = (
+        'Reindex course data in Algolia, adding on enterprise-specific metadata'
```
marlonkeating: Add help text for this job

enterprise_catalog/apps/api/tests/test_tasks.py
jesperhodge: good catch!

marlonkeating (approved):

PR #1069, opened 2025-05-27
chore: delete tmp indices task should depend on updated_at field

jesperhodge:
## Description

- Testing deleting tmp indices on prod didn't work. An investigation showed that `created_at` for these tmp indices was set to the creation date of the original index they were copying, in 2020.
- I changed the selection of date range for which indices to delete to use the `updated_at` field. As far as I can see, that corresponds to the `Last Active` display on the Algolia dashboard.
- I introduced logging for each index including its metadata and added logging to track why it becomes marked for deletion or not. That way when testing on prod with dry run, we can exactly trace what is happening.
- I added one more safeguard, just in case, that makes sure the index's `entry` field is 0 (telling us that the index is empty).

## Testing Instructions

- If you don't have Algolia stage set up for enterprise-catalog, edit or create your `enterprise_catalog/settings/private.py`
- Add your Algolia stage credentials like this:

```
ALGOLIA = {
    'INDEX_NAME': 'enterprise_catalog_new',
    'REPLICA_INDEX_NAME': 'enterprise_catalog_new_duration_desc',
    'APPLICATION_ID': <secret>,
    'API_KEY': <secret>,
}
```

- `make dev.up`
- `make app-shell`
- `./manage.py remove_old_temporary_catalog_indices --dry-run`
- Search log output for `"Index names to delete"` to see dry run result

## Post-review

### DIFF
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -629,7 +629,7 @@ def _last_updated_between(index, min_days_ago, max_days_ago):
     max_days_ago_date = (datetime.today() - timedelta(days=max_days_ago)).date()
     if max_days_ago_date < datetime.fromisoformat(datestring).date() < min_days_ago_date:
         logger.info(
-            'Index %s meets condition: min_days %s and max_days %s, because updatedAt: %s', index.get('name', ''),
+            'Index %s meets condition: min_days %s and max_days %s, because updatedAt: %s', index_name,
             min_days_ago,
             max_days_ago,
             datestring
@@ -638,7 +638,7 @@ def _last_updated_between(index, min_days_ago, max_days_ago):
 
     logger.info(
         'Index %s does not meet condition: min_days %s and max_days %s, because updatedAt: %s',
-        index.get('name', ''),
+        index_name,
         min_days_ago,
         max_days_ago,
         datestring
@@ -673,23 +673,23 @@ def _is_empty_index(index):
     Returns whether the index has any entries.
     """
     res = index.get('entries', None)
+    index_name = index.get('name', '')
     if res == 0:
-        logger.info('Index %s meets condition: has 0 entries', index.get('name', ''))
+        logger.info('Index %s meets condition: has 0 entries', index_name)
         return True
 
-    logger.info('Index %s does not meet condition: has 0 entries, because entries: %s', index.get('name', ''), res)
+    logger.info('Index %s does not meet condition: has 0 entries, because entries: %s', index_name, res)
     return False
 
 
-def _is_inactive_tmp_index(index, min_days_ago, max_days_ago):
-    """
-    Returns whether the index is a temporary index that was last updated between min_days_ago and max_days_ago.
-    """
-    logger.info(index)
-    return _is_tmp_index(index) and _last_updated_between(index, min_days_ago, max_days_ago) and _is_empty_index(index)
-
-
 def _retrieve_inactive_tmp_indices(client, min_days_ago, max_days_ago):
+    def _is_inactive_tmp_index(index, min_days_ago, max_days_ago):
+        """
+        Returns whether the index is a temporary index that was last updated between min_days_ago and max_days_ago.
+        """
+        logger.info(index)
+        return _is_tmp_index(index) and _last_updated_between(index, min_days_ago, max_days_ago) and _is_empty_index(index)
+
     indices = _get_all_indices(client)
     logger.info('Retrieved %d indices from Algolia', len(indices))
     logger.info('Processing indices: \n')
--- a/enterprise_catalog/apps/api/tasks.py
+++ b/enterprise_catalog/apps/api/tasks.py
@@ -666,12 +666,25 @@ def _get_all_indices(client):
     return indices
 
 
+def _is_empty_index(index):
+    """
+    Returns whether the index has any entries.
+    """
+    res = index.get('entries', None)
+    if res == 0:
+        logger.info('Index %s meets condition: has 0 entries, because entries: %s', index.get('name', ''), res)
+        return True
+
+    logger.info('Index %s does not meet condition: has 0 entries, because entries: %s', index.get('name', ''), res)
+    return False
+
+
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/api/tasks.py
```diff
@@ -635,13 +649,49 @@ def _is_tmp_index(index):
...
+    return indices
+
+
+def _is_empty_index(index):
+    """
+    Returns whether the index has any entries.
+    """
+    res = index.get('entries', None)
+    if res == 0:
+        logger.info('Index %s meets condition: has 0 entries, because entries: %s', index.get('name', ''), res)
```
marlonkeating: If we know that `res` is going to be zero in this case, why print it?
```suggestion
    if res == 0:
        logger.info('Index %s meets condition: has 0 entries', index.get('name', ''))
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -615,18 +615,32 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):
...
+            datestring
+        )
         return True
+
+    logger.info(
+        'Index %s does not meet condition: min_days %s and max_days %s, because updatedAt: %s',
+        index.get('name', ''),
+        min_days_ago,
+        max_days_ago, datestring
+    )
```
marlonkeating: nit
```suggestion
        max_days_ago, 
        datestring
    )
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -615,18 +615,32 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):
...
 
-def _created_between(index, min_days_ago, max_days_ago):
+def _last_updated_between(index, min_days_ago, max_days_ago):
     """
     Returns whether the index was created between min_days_ago and max_days_ago.
     """
-    datestring = index.get('createdAt', None)
+    datestring = index.get('updatedAt', None)
     if not datestring:
+        logger.error('Index %s does not have an updatedAt field.', index.get('name', ''))
```
marlonkeating: nit: breaking index name into variable could help readability
```suggestion
    index_name = index.get('name', '')
    datestring = index.get('updatedAt', None)
    if not datestring:
        logger.error('Index %s does not have an updatedAt field.', index_name)
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -635,13 +649,49 @@ def _is_tmp_index(index):
...
+
+    logger.info('Index %s does not meet condition: has 0 entries, because entries: %s', index.get('name', ''), res)
+    return False
+
+
+def _is_inactive_tmp_index(index, min_days_ago, max_days_ago):
+    """
+    Returns whether the index is a temporary index that was last updated between min_days_ago and max_days_ago.
+    """
+    logger.info(index)
```
marlonkeating: Are we keeping this log line?  Should include context in the logging statement if so.

jesperhodge: This is actually intended.
The context is found in the function that calls this one:
`logger.info('Processing indices: \n')`
Since this is a filter function that's called in a loop, each time it's called it should print one more index. That's followed by statements describing what filter conditions the index met.

The output is something like this:
```
"Processing indices:"
"{...index 1}"
"Index <name> meets condition: ..."
"Index <name> does not meet condition: ..."
"{...index2}"
... etc
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -615,18 +615,34 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):
...
         return False
 
-    created_timestamp = datetime.strptime(datestring, '%Y-%m-%dT%H:%M:%S.%fZ').timestamp()
-    difference_in_days = (time.time() - created_timestamp) / (60 * 60 * 24)
-    if min_days_ago < difference_in_days < max_days_ago:
+    min_days_ago_date = (datetime.today() - timedelta(days=min_days_ago)).date()
+    max_days_ago_date = (datetime.today() - timedelta(days=max_days_ago)).date()
+    if max_days_ago_date < datetime.fromisoformat(datestring).date() < min_days_ago_date:
+        logger.info(
+            'Index %s meets condition: min_days %s and max_days %s, because updatedAt: %s', index.get('name', ''),
```
marlonkeating:
```suggestion
            'Index %s meets condition: min_days %s and max_days %s, because updatedAt: %s', index_name,
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -615,18 +615,34 @@ def index_enterprise_catalog_in_algolia_task(self, force=False, dry_run=False):
...
+            min_days_ago,
+            max_days_ago,
+            datestring
+        )
         return True
+
+    logger.info(
+        'Index %s does not meet condition: min_days %s and max_days %s, because updatedAt: %s',
+        index.get('name', ''),
```
marlonkeating:
```suggestion
      index_name,
```

enterprise_catalog/apps/api/tasks.py
```diff
@@ -635,13 +651,49 @@ def _is_tmp_index(index):
...
+    return indices
+
+
+def _is_empty_index(index):
+    """
+    Returns whether the index has any entries.
+    """
+    res = index.get('entries', None)
+    if res == 0:
+        logger.info('Index %s meets condition: has 0 entries', index.get('name', ''))
```
marlonkeating: nit: can also break out `index_name` here

enterprise_catalog/apps/api/tasks.py
marlonkeating: Sure, I can see the context in which the logging is happening.  However, this logging reduces the reusability of this helper function outside of its one current use case.  I would recommend moving the logging into `_retrieve_inactive_tmp_indices`, and if you want to keep the lambda as a one-liner, you can make an inner function that includes the logging for each index.

marlonkeating (approved):

PR #1072, opened 2025-06-04
fix remove tmp indices date check

jesperhodge:
## Description

Minor fix to the date check of the remove-tmp-indices task.
Since the index date was < task min_days_ago, that didn't include the day specified.
For example, if you set min_days_ago to 0, that won't include today.
That makes it difficult to test index deletion on stage, so I'm fixing it here.

## Testing instructions

Correct unit tests are sufficient, so no need for further manual testing.


marlonkeating (approved):
nit: Can amend the commit message to be prefixed with `fix: ` in order to be compliant with https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0051-bp-conventional-commits.html#type, but otherwise looks good!

PR #1089, opened 2025-08-26
fix: normalized course price should never fallback to None

pwnage101:
This fixes a bug where enterprise-access can-redeem returns 422 for an ExecEd course without an advertised course run because it could not tolerate a null price from enterprise-catalog. To solve that case, the fix in this commit is to set the normalized price to the ExecEd entitlement price instead of null. If the course is not an ExecEd course, then fallback to 0.0 dollars instead of null.

[ENT-10866](https://2u-internal.atlassian.net/browse/ENT-10866)

### DIFF
--- a/enterprise_catalog/apps/catalog/serializers.py
+++ b/enterprise_catalog/apps/catalog/serializers.py
@@ -152,14 +152,21 @@ class NormalizedContentMetadataSerializer(ReadOnlySerializer):
 
     @extend_schema_field(serializers.FloatField)
     def get_content_price(self, obj) -> float:  # pylint: disable=unused-argument
-        if not self.course_run_metadata:
-            return None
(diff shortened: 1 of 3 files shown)


enterprise_catalog/apps/catalog/serializers.py
```diff
@@ -152,14 +152,21 @@ def get_enroll_by_date(self, obj) -> str:  # pylint: disable=unused-argument
 
     @extend_schema_field(serializers.FloatField)
     def get_content_price(self, obj) -> float:  # pylint: disable=unused-argument
-        if not self.course_run_metadata:
-            return None
```
pwnage101, line 156: If we were actually using and enforcing python type hints, we would never have let this pass CI...

brobro10000 (approved):

PR #1121, opened 2025-12-10
chore: add deprecation notice to PR template

pwnage101:
ENT-11240


kiram15 (approved):

pwnage101:
Closing because we're not allowed to deprecated these yet until following through with the DEPR process first.

PR #1132, opened 2026-02-03
Exclude unpublished courses from catalog content metadata API,ENT-11150

vshaikismail-sonata:
Root cause:-
The active status filtering was conditionally applied only when no [content_keys_filter] was provided. When specific content keys were requested, the filtering was skipped, allowing inactive courses to slip through.

Changes:-
-Removed the conditional check: Eliminated the if not content_keys_filter: block
to ensure filtering runs unconditionally.
-Unindented the filtering logic: Moved the queryset = [item for item in queryset if self.is_active(item)]
and related debug logging outside the conditional, so it's executed for all requests.
-Updated the comment: Changed to # Always filter out inactive courses to ensure only active content is always returned via API.

Tests:-
-Added unit tests to verify inactive courses are filtered out, and all the test cases were successful.
-Verified locally using mocked catalog content metadata.
-Mocking was used due to lack of real LMS data and was unable to create test courses after lot of trials.
-confirmed courses without active runs are excluded.

Screenshot

### DIFF
--- a/enterprise_catalog/apps/api/v1/tests/test_views.py
+++ b/enterprise_catalog/apps/api/v1/tests/test_views.py
@@ -1530,6 +1530,43 @@ class EnterpriseCatalogGetContentMetadataTests(APITestMixin):
         False,
         True
     )
+    def test_excludes_unpublished_courses(self, learner_portal_enabled, mock_api_client):
+    #pylint: disable=unused-argument
+        """
+        Verify that unpublished courses are never returned
+        in the content metadata API response.
+        """
+
+        unpublished_course = ContentMetadataFactory(
+            content_type=COURSE,
+            json_metadata={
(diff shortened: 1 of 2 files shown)


gshivajibiradar (commented):
Reviewed the latest updates to the same files. The previously noted issues have been addressed: merge conflict markers have been removed, inactive-course filtering is now applied once unconditionally, and the unpublished-course test is no longer duplicated. The test coverage and view logic are now clean, consistent, and aligned with the intended behavior.

gshivajibiradar (approved):
Reviewed the latest changes—merge conflict markers are removed, inactive-course filtering is applied once, and the unpublished-course test is no longer duplicated; the code and tests now look clean and aligned with the intended behavior

zwidekalanga (approved):

